PR #1, opened 2022-06-25
fixes for libc++/macOS

veprbl:
This fixes the compilation on macOS. I can submit a similar PR for ip6 (and athena?).

### DIFF
--- a/src/BarrelBarDetectorWithSideFrame_geo.cpp
+++ b/src/BarrelBarDetectorWithSideFrame_geo.cpp
@@ -16,6 +16,7 @@
 #include "DDRec/DetectorData.h"
 #include "DDRec/Surface.h"
 #include "XML/Layering.h"
+#include <array>
 
 using namespace std;
 using namespace dd4hep;
--- a/src/BarrelTrackerWithFrame_geo.cpp
+++ b/src/BarrelTrackerWithFrame_geo.cpp
@@ -13,6 +13,7 @@
 #include "DDRec/Surface.h"
 #include "XML/Layering.h"
 #include "XML/Utilities.h"
+#include <array>
 
 #if defined(USE_ACTSDD4HEP)
 #include "ActsDD4hep/ActsExtension.hpp"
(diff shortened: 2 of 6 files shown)


wdconinc:
Thanks a lot for the pull request. We are still transitioning to using GitHub as the primary repository. Could you create an account at eicweb.phy.anl.gov (with any institutional email address, tagging @sly2j for fast approval), and create a merge request in the eicweb.phy.anl.gov/EIC/detectors/ecc repository? That will run through the full CI chain too. Same thing for ip6 (athena is optional; ecce is the main development repo now, and at some point we will rename that to a proper new collaboration name).

Two substantive comments:
- I don't see `std::array` used in those files (I looked on my phone and could very well have missed it). Is this a fix that addresses an issue in an included header? If so, can we also ensure this is reported upstream?
- Does clang not support `std::quick_exit`? That's a c++11 standard component. Are we maybe setting the standard incorrectly? I've not had issues with clang (13, 14, and now 15 dev) but that's on Linux.

Once we transition to GitHub, we will be able to run CI tests against Mac OS (at least until 10.15, since cvmfs still doesn't work in the GitHub macos-11 images... I guess we can probably work around that actually...).

wdconinc:
I'll keep this open until the merge request at eicweb is complete.

veprbl:
I was going to submit via the ANL instance, but my account is still blocked (apparently there is some email delivery issue). I was hoping we could get at least human-review started here.

* `std::array` is certainly used https://github.com/eic/ecce/blob/351f77ad0758ce488ddb0360d337b6cc36a7f2ef/src/BarrelBarDetectorWithSideFrame_geo.cpp#L47
* Both stdlibc++ and libc++ seem to assume that the `std::quick_exit` may not be available on every platform (`_GLIBCXX_HAVE_QUICK_EXIT`/`_LIBCPP_HAS_QUICK_EXIT`). Even though none of the documentation that I've seen seems to mention that. I have not dug too far into this, but it would seem like I don't have it with my toolchain, hence I'd like to treat this as a compatibility issue.

wdconinc:
Thanks for the additional info. Consider it approved here, and hopefully we get the account setup soon.

wdconinc:
Were you able to create this pull request on eicweb? Just want to make sure we keep this on the radar.

veprbl:
No, but I've just sent a ping (cc'ing you) to remind about this.

PR #4, opened 2022-07-09
treewide: use explicit getAttrOrDefault<std::string>

veprbl:
This is needed to support DD4hep built with `-DDD4HEP_USE_XERCESC=ON`.

### DIFF
--- a/src/BarrelCalorimeterInterlayers_geo.cpp
+++ b/src/BarrelCalorimeterInterlayers_geo.cpp
@@ -178,8 +178,8 @@ void buildFibers(Detector& desc, SensitiveDetector& sens, Volume& s_vol, xml_com
   double      f_radius                     = getAttrOrDefault(x_fiber, _U(radius), 0.1 * cm);
   double      f_spacing_x                  = getAttrOrDefault(x_fiber, _Unicode(spacing_x), 0.122 * cm);
   double      f_spacing_z                  = getAttrOrDefault(x_fiber, _Unicode(spacing_z), 0.134 * cm);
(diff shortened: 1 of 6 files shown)


wdconinc:
Still not able to accept PRs here... Can you file on eicweb? I'm hoping to switch the roles of the mirrors soon, maybe by Monday.

wdconinc:
And, yeah, eicweb has some disk space issues right now, so pipelines will probably fail.

wdconinc (approved):

wdconinc:
Do we need to add sqlite3 [here](https://github.com/AIDASoft/run-lcg-view/blob/master/run-macOS.sh#L31)?

wdconinc:
I've asked github support to detach the fork https://github.com/eic/run-cvmfs-osg-eic-shell from the upstream. Then we can have a new fork that I'm not going to repurpose. That will help us in figuring out if run-lcg-view needs any changes, and submitting a PR to upstream.

veprbl:
> Do we need to add sqlite3 [here](https://github.com/AIDASoft/run-lcg-view/blob/master/run-macOS.sh#L31)?

I suppose that won't help. I can see `__sqlite3_db_clone` symbol exported from `/usr/lib/libsqlite3.dylib`, but not from a build of sqlite 3.38.5. So, I suppose, the LCG environment must be using `DYLD_LIBRARY_PATH` and that breaks the normal library loading.

wdconinc:
In addition, now the ip6 macos jobs are now hanging until timeout... No idea why.

veprbl:
Link?

wdconinc:
https://github.com/eic/ip6/actions/runs/2644602915

PR #9, opened 2022-07-25
feat: limit workflows with concurrency

wdconinc:
In order to limit the number of active or queued workflows, this changes workflows to only run on pull_request and as part of a concurrency group which will cancel in-progress jobs.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -1,9 +1,12 @@
 name: linux-eic-shell
 
 on:
-- push
 - pull_request
 - workflow_dispatch
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
 
 jobs:
   xmllint-before-build:
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -1,8 +1,13 @@
 name: linux-eic-shell
 
 on:
-- pull_request
-- workflow_dispatch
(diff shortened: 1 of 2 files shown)


veprbl:
Could try https://github.com/marketplace/actions/skip-duplicate-actions

wdconinc:
Please feel free to implement. I'm just trying to get things in 'good
enough' state :-)

veprbl:
Ideas are cheap, right. Hence I will experiment a bit in https://github.com/veprbl/ecce/tree/pr/skip-duplicate-actions

.github/workflows/linux-eic-shell.yml
```diff
@@ -1,10 +1,13 @@
 name: linux-eic-shell
 
 on:
-- push
 - pull_request
 - workflow_dispatch
```
veprbl, line 5:
```suggestion
  push:
    branches:
      - main
    tags:
      - '*'
  pull_request:
  workflow_dispatch:
```

.github/workflows/linux-lcg.yml
```diff
@@ -1,6 +1,11 @@
 name: linux-lcg
 
-on: [push, pull_request]
+on:
+- pull_request
```
veprbl, line 4:
```suggestion
  push:
    branches:
      - main
    tags:
      - '*'
  pull_request:
```

veprbl (approved):
I now remember, for STAR we also essentially went for disabling push builds, so this is probably the best out of simplest approaches to avoid building for push trigger on a PR branch. A minor improvement is to allow CI at least on main branch and for tags.

PR #12, opened 2022-07-26
dRICH: Parse modified drich xml files (fork PR check)

c-dilks:
If the config yaml file is
```
features:
  pid:
    drich:
```
it will use the default `compact/drich.xml`, else if it is
```
features:
  pid:
    drich: my_custom_drich.xml
```
it will instead use `my_custom_drich.xml`


c-dilks:
This was a test PR from a fork. Close in favor of https://github.com/eic/ecce/pull/14

PR #17, opened 2022-07-26
ecce -> epic

veprbl:
#15 but with access to secrets


veprbl:
Note: this doesn’t touch workflows

PR #20, opened 2022-07-27
Reparametrize tracker and update to latest EPIC design

sly2j:
Reparametrized the tracker to be a bit more modular and maintainable. The tracker now uses a main configuration file (`compact/tracker.xml`), which sets up the primary parameters for the tracker, as well as the ACTS tracking assemblies. It then includes the relevant specific detector definitions from the `compact/tracker` subdirectory.

The tracker is currently implemented according to the June 2022 setup for the silicon barrel, with a disk configuration and service/support configuration similar to the ECCE proposal version. Some minor details were changed (eg. made the forward-most service cone projective) to avoid overlaps.

I completely reworked how the service/support structures are constructed - it now has its own compact file instead of attaching bits and pieces to the different detector plugins. This should make future updates much easier, and overall help maintainability of the implementation.

Furthermore, I disabled the TOF for now as it requires serious work to update it to the latest EPIC design. I left space allocated in the ACTS tracking assembly setup so the TOF could be integrated relatively easily in the future.

Geoviewer view of the tracker integrated in the inner central detector.


templates/epic.xml.jinja2
```diff
@@ -104,7 +104,7 @@
   <documentation level="10">
     ## Central tracking detectors
   </documentation>
-  <include ref="compact/tracking_config.xml"/>
+  <include ref="compact/tracker_ecce.xml"/>
```
wdconinc, line 107: we've got an `epic` tracker now ;-)

sly2j: so do we want `tracker_epic.xml `or `epic_tracker.xml`?

veprbl: Existing convention seems to say `tracker.xml`

wdconinc: I prefer my trackers big-endian.

sly2j: I wanted more granularity here to make it easier to switch/test different high-level configurations (as in with very different parametrization). The reason it was called `*_ecce` is because it is the configuration from the proposal. Maybe calling it `tracker_reference.xml` makes the most sense.

wdconinc (approved):
Lots of work in here, and no issues with it from my side (but of course I've not review this in as much detail as for a smaller PR).

wdconinc:
Re: https://eicweb.phy.anl.gov/EIC/benchmarks/detector_benchmarks/-/pipelines/35222. Benchmark issues are (to some extent) expected, but most of the failures there are actually fallout of the `$BEAMLINE_PATH` requirement in the main branch `epic.xml`. Those should be fixed by now (issues in `common_bench`).

PR #21, opened 2022-07-27
feat: tcsh support for setup script (for BNL and JLab systems)

wdconinc:
~~Note, FIXME: incorrect when LD_LIBRARY_PATH empty~~


wdconinc:
Closing this approach to supporting tcsh for now.

PR #24, opened 2022-07-28
fix(dRICH): improve optics

c-dilks:
- [x] maximize polar acceptance (with no B field)
- [x] try to ensure photons are at normal incidence to sensors
- [x] get close to the parallel-to-point focal region

# Updated Parameters
```
    # focus_tune_x          = -5
    # focus_tune_z          = 0
    # sensor_sphere_radius  = 100
    # sensor_sphere_centerz = -50
    # 
    # SECTOR 0 MIRROR:
    #   mirror x = 113.888311 cm
    #   mirror y = 0.000000 cm
    #   mirror z = 94.364601 cm
    #   mirror R = 218.535399 cm
```

# Changes

## Foci
Wide, collimated photon beams reflecting off the mirror, to show the parallel-to-point foci
#### Before changes:
before-foci
#### After changes:
after-foci

## Rings
8 GeV pions, thrown at 4 different polar angles within the dRICH acceptance. Note the highest-theta has no aerogel ring, because there is no aerogel coverage at high theta.
#### Before changes:
before
#### After changes (B-field off):
after
#### After changes (B-field on):
field_on

## Sensor Pixel Hit Map (after changes)
Thrown 8 GeV pions, equally distributed in theta and phi, to one dRICH sector. The lowest theta pions are thrown at phi=0 only; the higher thetas have 5 phi bins: one along phi=0, one along each sector boundary (adjacent sector hit maps are below), and one in between phi=0 and the sector boundary.

We have losses at high theta along the sector boundary (note the missing gas rings)

c1_n2
c1_n3
c1_n7

### DIFF
--- a/compact/drich.xml
+++ b/compact/drich.xml
@@ -151,8 +151,8 @@
   rmax="DRICH_rmax2 - DRICH_wall_thickness - 3.0*cm"
   phiw="59.5*degree"
   thickness="0.2*cm"
-  focus_tune_x="-10.0*cm"
-  focus_tune_z="15.0*cm"
+  focus_tune_x="-5.0*cm"
+  focus_tune_z="0.0*cm"
   />
 
 <!-- /detectors/detector/sensors -->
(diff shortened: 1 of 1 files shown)


wdconinc (approved):
This looks good (even great) to me.

PR #28, opened 2022-07-31
fix: trigger benchmarks with DETECTOR variables

wdconinc:
The JUGGLER_ variables are deprecated and the DETECTOR variables should be used.


wdconinc:
Implemented elsewhere and already in main.

PR #29, opened 2022-07-31
feat: install dd4hep_configure_scripts

wdconinc:
This allows for more convenient use of the geometry without (it is hoped) custom setup scripts:
- `source bin/thisepic.sh`
- `bin/run_test_epic.sh tcsh`
(also avoiding the need for tcsh setup scripts by directing users to start a subshell in the second example)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -146,6 +146,8 @@ endif()
 #-----------------------------------------------------------------------------------
 # Configure and install detector setup script
 #
+dd4hep_configure_scripts(${PROJECT_NAME} DEFAULT_SETUP)
+
 execute_process(
   COMMAND git rev-parse --abbrev-ref HEAD
   WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}


wdconinc:
Hmm, `dd4hep_configure_scripts` does some undesirable things, like installing into `examples`:
```console
-- Installing: /home/runner/work/epic/epic/install/examples/epic/compact
-- Installing: /home/runner/work/epic/epic/install/examples/epic/compact/unused
-- Installing: /home/runner/work/epic/epic/install/examples/epic/compact/unused/rwell_tracker_barrel.xml
-- Installing: /home/runner/work/epic/epic/install/examples/epic/compact/unused/ecal_barrel_hybrid.xml
-- Installing: /home/runner/work/epic/epic/install/examples/epic/compact/unused/ecal_backward_crystal_glass.xml
```
This is caused by https://github.com/AIDASoft/DD4hep/blob/master/cmake/DD4hepBuild.cmake#L416,
```
    if ( IS_DIRECTORY scripts )
      dd4hep_install_dir ( compact scripts DESTINATION examples/${_pkg} )
    endif()
    if ( IS_DIRECTORY compact )
      dd4hep_install_dir ( compact DESTINATION examples/${_pkg} )
    endif()
```
Because we have `scripts` and `compact` we end up installing everything into examples too. Not ideal...

veprbl:
I think we can drop the idea of using the DD4hep cmake macros (there are others we don't use...). They seem to only make things more opaque for reader of our cmake files, and the benefit is not clear.

wdconinc:
Yeah, I'm also thinking the `setup.sh` in `bin` may just be easier.

veprbl:
Let's table this for now, unless there are objections

PR #30, opened 2022-07-31
feat: install using GNUInstallDirs

wdconinc:
This will ensure that files are installed into their correct locations, even on systems that use lib64 as LIBDIR. This commit does not actually change the installed location of any files (but setup.sh should go into BINDIR).


veprbl (approved):

veprbl:
However should probably update https://github.com/eic/epic/blob/fb1c73f284934622038d307afc7a050f570cf50d/templates/setup.sh.in#L4

veprbl (approved):

Issue #23, opened 2022-07-27
Respect $BEAMLINE_PATH


veprbl:
**Is your feature request related to a problem? Please describe.**
The ecce compact file does not rely on BEAMLINE_PATH set in
https://github.com/eic/ip6/blob/5fe0178ea0fc6a490c776e02ff750693a60f576f/templates/setup.sh.in#L4
instead, some ad-hoc symlink to ip6 is used.

**Describe the solution you'd like**

```diff
diff --git a/templates/epic.xml.jinja2 b/templates/epic.xml.jinja2
index 5e30dbe..0e08a0c 100644
--- a/templates/epic.xml.jinja2
+++ b/templates/epic.xml.jinja2
@@ -44,8 +44,8 @@
       The ip6 (or other ip) defines should be included first.
       These files have only a define tags.
   </documentation>
-    <include ref="ip6/definitions.xml" />
-    <include ref="ip6/far_forward/fields_{{ pbeam | default("275", true) }}.xml" />
+    <include ref="${BEAMLINE_PATH}/ip6/definitions.xml" />
+    <include ref="${BEAMLINE_PATH}/ip6/far_forward/fields_{{ pbeam | default("275", true) }}.xml" />
     <include ref="compact/definitions.xml" />
   </define>
 
```

If the variable is not set, DD4hep will fail:
```
RunPlugin        ERROR ++ Exception while executing plugin DD4hep_CompactLoader:
                dd4hep: Severe error during environment lookup of ${BEAMLINE_PATH}/ip6/definitions.xml Evaluator::Object : unknown variable : ${BEAMLINE_PATH}
```

There doesn't seem to be a way to provide a default value for `${BEAMLINE_PATH}` in DD4hep evaluator, despite some similar feature already working:
https://github.com/AIDASoft/DD4hep/blob/02c6d5ac04c187f37631f9f15ca6c1c08452d992/examples/LHeD/compact/compact_Lhe_dip_sol_ell.xml#L34-L39

**Describe alternatives you've considered**
 - Merge ip6 into this repository

**Additional context**
Some upstream discussion in https://github.com/AIDASoft/DD4hep/issues/934

wdconinc:
I have a deep instinctive aversion to environment variables (somehow it's only physicists' software that seems to require them). But this may be one of those cases we can't avoid; because I dislike having to carefully arrange repositories even more.

DraTeots:
Put IP6 as a git submodule maybe?

DraTeots:
[...]

PR #31, opened 2022-07-31
feat*: remove the need for symlinks by using two variables

wdconinc:
We pretty much require setting environment variables already, so this just formalizes this. Based on #23 and discussion therein.

For consistency, ip6 should also install its xml files into `share/ip6/compact` but it doesn't exactly because we used a symlink and would conflict with the existing `compact` directory of the detector repository.

### DIFF
--- a/templates/epic.xml.jinja2
+++ b/templates/epic.xml.jinja2
@@ -44,15 +44,15 @@
       The ip6 (or other ip) defines should be included first.
       These files have only a define tags.
   </documentation>
-    <include ref="ip6/definitions.xml" />
-    <include ref="ip6/far_forward/fields_{{ pbeam | default("275", true) }}.xml" />
-    <include ref="compact/definitions.xml" />
+    <include ref="${BEAMLINE_PATH}/ip6/definitions.xml" />
+    <include ref="${BEAMLINE_PATH}/ip6/far_forward/fields_{{ pbeam | default("275", true) }}.xml" />
+    <include ref="${DETECTOR_PATH}/compact/definitions.xml" />
   </define>
 
   <includes>
-    <gdmlFile ref="compact/elements.xml"/>
-    <gdmlFile ref="compact/materials.xml"/>
-    <file     ref="compact/optical_materials.xml"/>
+    <gdmlFile ref="${DETECTOR_PATH}/compact/elements.xml"/>
+    <gdmlFile ref="${DETECTOR_PATH}/compact/materials.xml"/>
+    <file     ref="${DETECTOR_PATH}/compact/optical_materials.xml"/>
   </includes>
 
   <limits>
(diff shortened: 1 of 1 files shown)


templates/epic.xml.jinja2
```diff
@@ -44,15 +44,15 @@
...
+    <include ref="${DETECTOR_PATH}/compact/definitions.xml" />
   </define>
 
   <includes>
-    <gdmlFile ref="compact/elements.xml"/>
-    <gdmlFile ref="compact/materials.xml"/>
-    <file     ref="compact/optical_materials.xml"/>
+    <gdmlFile ref="${DETECTOR_PATH}/compact/elements.xml"/>
+    <gdmlFile ref="${DETECTOR_PATH}/compact/materials.xml"/>
+    <file     ref="${DETECTOR_PATH}/compact/optical_materials.xml"/>
```
veprbl, line 55: I have the same reservation about using DETECTOR_PATH as in https://github.com/eic/ip6/pull/9#issuecomment-1200526270. Would be interested to hear your line of thinking when you made this.

veprbl: I just had to add those `${DETECTOR_PATH}` when creating my own composition for the EPIC detector by first copying an existing xml from a package to my local directory. Having this PR as is would be useful for that case.

veprbl (approved):

wdconinc:
Reasons I have put this on hold is that I realized that we use a mixture of `BEAMLINE` and `BEAMLINE_CONFIG` variables, followed by `VERSION` etc in a way that is not consistent with how we treat `BEAMLINE`.

Now I have verified that at least 1) we are consistent in the definition and use of `DETECTOR_PATH`, but 2) we don't define `BEAMLINE_PATH` yet in the containers. So, it depends on rolling this into [eic_container](https://eicweb.phy.anl.gov/containers/eic_container) and onto cvmfs.

**tl;dr** I'll merge this tomorrow.

wdconinc:
The `setup.sh` in the geometry repos has it right; it's the `setup.sh` generated by [setup_detectors.py](https://eicweb.phy.anl.gov/containers/eic_container/-/blob/master/containers/jug/setup_detectors.py) that doesn't yet.

wdconinc:
Confirmed that `${BEAMLINE_PATH}` (and ${DETECTOR_PATH}`) is now set in the `eic-shell` containers and that `checkGeometry -c epic_sciglass.xml` now works without errors inside `eic-shell` after loading the correct `/opt/detector/epic-nightly/setup.sh`.

Potentially failure scenario to be mindful of:
1. Users loads `/opt/detector/setup.sh` and expects `checkGeometry -c epic_sciglass.xml` to work, but `/opt/detector/setup.sh` points to a non-EPIC geometry (as it does not)...

Nevertheless, ready for merging.

PR #32, opened 2022-07-31
test: post commit status with eicweb pipeline

wdconinc:
This replaces the posting of a commit comment with updating a commit status. This status is attached to the pull_request, with a link to the eicweb pipeline.

GitHub actions tokens expire, so we cannot pass them to the triggered pipelines. We will have to figure out a way to pass the success of those pipelines back using a token in those pipelines.

For now we will need to check the pending status by hand and override.

Credit for the idea: @veprbl

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -267,6 +267,7 @@ jobs:
         variables: |
           JUGGLER_DETECTOR_REPOSITORYURL=${{ github.server_url }}/${{ github.repository }}
           JUGGLER_DETECTOR_VERSION=${{ github.ref_name }}
+          GITHUB_REPOSITORY=${{ github.repository }}
           GITHUB_SHA=${{ github.event.pull_request.head.sha || github.sha }}
     - run: |
         gh api \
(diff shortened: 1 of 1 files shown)


wdconinc:
An interesting question is whether we can post a status back against a pull request from a fork? But at least for now this seems to do the job.

veprbl:
I think the status will need to be posted using an app token saved as a secret on gitlab. 
https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token
possibly not on the personal account but a bot/organization account
https://docs.github.com/en/get-started/learning-about-github/types-of-github-accounts#personal-user-accounts

wdconinc:
Unfortunately there are no organization or repository tokens that can be used for this. It needs to be a PAT, and unfortunately those cannot be restricted to a single repository either. Anyway, still this is better than the comments.

veprbl:
An application can be restricted to only have a permission to update a commit status. For more granular access one might need to create separate accounts.

wdconinc:
Yeah, I'm ok with putting my PAT restricted to repo:status on eicweb as a repository secret, so that's what I'm working on now.

wdconinc:
This should now also result in a successful status update back from eicweb (at least at the end of detector_benchmarks).

1. github triggers eicweb and updates status to "have been triggered"
2. eicweb starts a pipeline and updates status to "are running"
3. eicweb finishes pipeline and updates status to "have succeeded"

wdconinc:
This now works as intended for detector benchmarks, and should work with reconstruction and physics benchmarks as well (was implemented after the last trigger was passed to eicweb).

PR #33, opened 2022-07-31
fix: specify BEAMLINE for triggered jobs

wdconinc:
This fixes the issue where triggered benchmark jobs are using the (old) eicweb ip6 repository. See e.g. https://eicweb.phy.anl.gov/EIC/benchmarks/detector_benchmarks/-/jobs/826381.


veprbl:
Shouldn't be needed

PR #35, opened 2022-08-06
README.md: suggest -DEPIC_ECCE_LEGACY_COMPAT=OFF

veprbl:
That is the migration path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ ln -s ../ip6/ip6 epic/ip6
 
 To configure, build, and install the geometry (to the `install` directory), use the following commands:
 ```bash
-cmake -B build -S . -D CMAKE_INSTALL_PREFIX=install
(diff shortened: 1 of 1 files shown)


veprbl:
The idea here is to, at some point, make non-off state into a warning, or switch default to ON, and then drop support for ECCE by default.

Also, I've recently learned that the rename was not fully backwards-compatible – it broke old version EICrecon, as they were probably relying on ECCEConfig.cmake (which they shouldn't have).

wdconinc (commented):
I'd go with making the default off and suggesting in the docs that people can turn the compatibility flag to on.

wdconinc (approved):
Sorry, didn't realize this was just a docs change.

wdconinc:
> Sorry, didn't realize this was just a docs change.

Not like the title was unclear... doh

PR #36, opened 2022-08-06
setup.sh.in: add macOS support

veprbl:


wdconinc:
This would be one of the places where we benefit from `thisepic.sh` since it does this centrally (and hopefully correctly).

wdconinc (approved):
Here's what dd4hep itself does:
```bash
dd4hep_add_library_path () 
{ 
    p=${1};
    if [ ]; then
        [[ "$p" = "/usr/lib" || "$p" = "/usr/local/lib" ]] && unset p && return;
        dd4hep_add_path DYLD_LIBRARY_PATH "$p";
        dd4hep_add_path DD4HEP_LIBRARY_PATH "$p";
    else
        dd4hep_add_path LD_LIBRARY_PATH "$p";
    fi;
    unset p
}
```
with the `[ ]` just a reflection of this having been installed on my linux laptop so it always branches to the `else`.

Do we also need to set `DD4HEP_LIBRARY_PATH` on mac?

veprbl:
Last time I've checked, `DD4HEP_LIBRARY_PATH` didn't do anything for me.

Issue #38, opened 2022-08-06
No usr/bin/python on ubuntu 22.04


DraTeots:
- branch - main
- Fresh ubuntu 22.04

When building the detector: 

```
/usr/bin/env: /usr/bin/env: ‘python’‘python’: No such file or directory: No such file or directory
```

While there are solutions like: 

```
sudo apt install python-is-python3
```

I would advocate to fix it in epic repo by switching to python3. It is now everywhere

```
/usr/bin/env python3
```

wdconinc:
Can you please file a bug report with steps to reproduce as requested by the template? Then we can identify which file is actually causing this.

PR #39, opened 2022-08-06
bin: use python3 in shebangs

veprbl:
Resolves: #38
cc @DraTeots

### DIFF
--- a/bin/generate_prim_file
+++ b/bin/generate_prim_file
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # same as make_dawn_views but stops at generating the prim file.
 # W. Armstrong (ANL), original bash script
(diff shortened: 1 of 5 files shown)


wdconinc (dismissed):
Just wanted to verify that this does not just 'happen to work' on some subset of systems, but per [PEP-394](https://peps.python.org/pep-0394/#for-python-script-publishers) we can expect that any system with python 3.* installed will have a python3 executable. We should probably propagate this change to wherever we use python scripts with shebang lines.

wdconinc:
OT but for PRs from forks we could direct users to eicweb where they should submit a pipeline by hand (cumbersome) or write a bot to submit the pipeline when instructed by a comment from a member of epic-devs or epic-admins.

wdconinc:
`./bin/make_dawn_views` is even worse:
```script
#!/usr/local/bin/python
```

wdconinc:
Test command
```console
grep -rIzl '^#!.*python[^3]' .
```

veprbl:
> `./bin/make_dawn_views` is even worse:
> 
> ```
> #!/usr/local/bin/python
> ```

Nice catch!

PR #45, opened 2022-08-10
Updated SciGlass barrel ECal design

veprbl:
This is an updated design for the SciGlass barrel electromagnetic calorimeter. The main change compared to the
SciGlassCalorimeterTiltless is introduction of tower families and additional flaring to reduce the gaps between the towers. The most significant change is the removal of the phi tilt (hence the temporary name suffix "Tiltless"), which significantly simplifies the construction of the geometry.

Since the designs are so different I've decided to not incrementally evolve your code, @wdconinc, but write one from scratch.

This geometry is derived by reverse-engineering the CAD design mesh kindly provided by Joshua Crafts.

<details>
<summary>Screenshots</summary>

</details>


src/SciGlassCalorimeterPanda_geo.cpp
```diff
@@ -257,7 +257,15 @@ static Ref_t create_detector(Detector& desc, xml::Handle_t handle, SensitiveDete
...
+  return create_detector(desc, handle, sens);
+}
+
 #ifdef EPIC_ECCE_LEGACY_COMPAT
-DECLARE_DETELEMENT(ecce_SciGlassCalorimeter, create_detector)
+DECLARE_DETELEMENT(ecce_SciGlassCalorimeter, create_detector_deprecated)
 #endif
-DECLARE_DETELEMENT(epic_SciGlassCalorimeter, create_detector)
+DECLARE_DETELEMENT(epic_SciGlassCalorimeter, create_detector_deprecated)
+DECLARE_DETELEMENT(epic_SciGlassCalorimeterPanda, create_detector)
```
wdconinc, line 271: In the past we placed deprecated geometry plugins in `src/deprecated` to make this more manifest as well.

templates/epic.xml.jinja2
```diff
@@ -157,7 +157,7 @@
   {% endif -%}
 
   {% if features is not defined or features['ecal'] is none or 'barrel_sciglass' in features['ecal'] %}
-  <include ref="compact/ecal_barrel_sciglass.xml"/>
+  <include ref="compact/ecal_barrel_sciglass_panda.xml.xml"/>
```
wdconinc, line 160: One imagines this should be:
```suggestion
  <include ref="compact/ecal_barrel_sciglass_panda.xml"/>
```

wdconinc (changes requested):
We probably need more documentation in the src plugin. Right now I think it would be hard for someone new coming in to make a change here or there.

src/SciGlassCalorimeterPanda_geo.cpp
veprbl: I'm not deprecating the Panda design yet, only the unspecific detector type `epic_SciGlassCalorimeter`. The `epic_SciGlassCalorimeterPanda` is still the default.

templates/epic.xml.jinja2
veprbl: Right. Was a last-minute xml rename.

veprbl:
> We probably need more documentation in the src plugin. Right now I think it would be hard for someone new coming in to make a change here or there.

The design is not so simple. It took us a meeting with an engineer and many hours of geometry mesh analysis to get here. The code should be relatively straightforward to understand, but one needs to learn how G4Trap is parametrized. If there are places that look tricky, please point them out.

wdconinc:
> If there are places that look tricky, please point them out.

It's just that there's only one line of comments in the function body :-)

E.g.
```cpp
  double sector_phi = sectors_handle.phi0();
  for (; sector < sectors_handle.number();
       sector++, sector_phi += sectors_handle.deltaphi()) {
    int row = 0;
    double row_phi = -rows_handle.deltaphi() / 2 * rows_handle.number();
    for (; row < rows_handle.number();
         row++, row_phi += rows_handle.deltaphi()) {
```
here one my wonder which one is actually looping over the `theta` direction.

src/SciGlassCalorimeterPanda_geo.cpp
wdconinc: Ah, I see now.

veprbl:
> > If there are places that look tricky, please point them out.
> 
> It's just that there's only one line of comments in the function body :-)

Maybe, it's just a self-documenting code :D

> here one my wonder which one is actually looping over the `theta` direction.

No mistake here, they both loop in phi. Center of the sector makes large steps for each sector and rows fill each sector. But, yeah, point taken. I would not see that by myself, though.

veprbl:
I've tried the strategy of building a row volume out of towers, using those to build sector volumes and using latter to build the whole detector. It does produce a nicer code (less loop nesting), but, to my surprise, it kills the performance of the simulation.

veprbl:
I seems like a simple tube envelope overlaps with the forward tracker:
```
 = Overlap ov00000: world_volume/InnerTrackerSupport_assembly_9/ForwardEndcapSupportCone_9 overlapping world_volume/EcalBarrel_15 ovlp=11.0219
 = Overlap ov00001: world_volume/InnerTrackerSupport_assembly_9/ForwardEndcapSupportRing_8 overlapping world_volume/EcalBarrel_15 ovlp=5.5
 = Overlap ov00002: world_volume/OuterMPGDSubSubAssembly_8/OuterMPGDBarrel_0/layer1_0 overlapping world_volume/EcalBarrel_15 ovlp=2.5
```
<details>
<summary>Screenshots</summary>


</details>

veprbl:
Opened #60 to address the overlap with MPGD.

veprbl:
epic_imaging benchmarks are already failing on main. Filed https://eicweb.phy.anl.gov/EIC/benchmarks/physics_benchmarks/-/issues/52 to track that.

compact/ecal_barrel_sciglass_tiltless.xml
```diff
@@ -0,0 +1,100 @@
...
+      An EM calorimeter with SciGlass blocks
+    </comment>
+
+    <detector id="4" name="EcalBarrel" type="epic_SciGlassCalorimeterTiltless" readout="EcalBarrelHits" vis="det_vis" calorimeterType="EM_BARREL">
+      <comment>
+        Global detector dimensions from https://eic.jlab.org/Geometry/Detector/Detector-20220624102507.html
+      </comment>
+
+      <dimensions rmin="EcalBarrel_rmin" rmax="EcalBarrel_rmax"
+                  zmax="195 * cm" zmin="-285 * cm"
```
wdconinc, line 37: Can these go into definitions, or can we maybe replace them with `zmin="max(EcalBarrel_zmin, -285 * cm)"` and such? We would want this never to get outside the volume imposed by the definitions file.

compact/ecal_barrel_sciglass_tiltless.xml
```diff
@@ -0,0 +1,100 @@
...
+
+          <comment>
+            Families in each row are constructed in two directions `dir_sign` - negative and positive rapidity - starting from the center.
+            The towers have the shape of truncated pyramids that have an opening angle in theta `flare_angle_polar`.
+            The towers are stacked in a row with their edges parallel to the edges of their neighbors, `gap` distance apart.
+            Families of towers are groups of towers having identical shape, i.e. the same tower is repeated `number` amount of times.
+            The front and rear faces are similar isosceles trapezoids only slightly deviating from a rectangular shape by an amount parametrized with `flare_angle_at_face`.
+          </comment>
+
+          <family dir_sign="+1" x1="2 * cm" y1="2 * cm" z_length="45.5 * cm" number="10" flare_angle_polar="1.16 * degree"  flare_angle_at_face="0.3059 * degree" vis="family1" />
```
wdconinc, line 58: Are `flare_angle_polar` and `flare_angle_at_face` somehow determined by other quantities here?

compact/ecal_barrel_sciglass_tiltless.xml
```diff
@@ -0,0 +1,100 @@
...
+    <detector id="4" name="EcalBarrel" type="epic_SciGlassCalorimeterTiltless" readout="EcalBarrelHits" vis="det_vis" calorimeterType="EM_BARREL">
+      <comment>
+        Global detector dimensions from https://eic.jlab.org/Geometry/Detector/Detector-20220624102507.html
+      </comment>
+
+      <dimensions rmin="EcalBarrel_rmin" rmax="EcalBarrel_rmax"
+                  zmax="195 * cm" zmin="-285 * cm"
+                  etamin="EcalBarrel_etamin" etamax="EcalBarrel_etamax"/>
+
+      <wedge-box inner_r="80.0 * cm" thickness="2 * mm" vis="becal_wedge" material="Aluminum5083" />
```
wdconinc, line 40: While a `-` as element name is allowed, it is uncommon and some xml processing software may not like it (in particular homegrown stuff). Can we just use `wedge_box` instead (here and in the plugin)?

wdconinc: Same with `carbon-fiber-support` and `cut-out`.

src/SciGlassCalorimeterTiltless_geo.cpp
```diff
@@ -0,0 +1,376 @@
...
+            family_dim_handle.x1() +
+            (dir_sign > 0) * (2 * y1) * tan(flare_angle_at_face);
+        const double x3 = x1 * (y2 / y1);
+        const double x4 = x2 * (y2 / y1);
+        const double theta = 0.;
+        const double phi = 0.;
+        const double alpha1 = 0.;
+        const double alpha2 = 0.;
+
+        for (unsigned int tower = 0; tower < number; tower++, tower_id += dir_sign) {
```
wdconinc, line 223: Is it possible to create a single module before the loop, and then just place it in the loop? This would allow geant4 to optimize the geometry better and treat tracks going into any block in the same row as if they are going into the same block.

wdconinc (commented):
- Do we need the png when the svg is included?
- Do we need to keep the PANDA design or can we just shift it up into `unused/`? In particular, do we need a separate configuration, or is `epic_sciglass.xml` just always going to be tiltless?

src/SciGlassCalorimeterTiltless_geo.cpp
veprbl: I went through this exercise once before. And even though the code looked much better, I saw a significant performance hit in Geant4 simulation. Back then I reverted without investigation. This would be nice to have implemented sometime.

compact/ecal_barrel_sciglass_tiltless.xml
veprbl: Good question! Both angles should be possible to determine `flare_angle_polar` from numbers of towers per family and azimuthal distribution of towers. In this geometry, we are looking to reproduce an engineering design for which the values of `flare_angle_polar` were hand-optimized for maintaining gaps between towers in azimuthal direction. If calculated, the values may not match the current design, which we are trying to reproduce for the campaign. The current flaring angles are based on a rule that towers are truncated pyramids, which, going forward, is something we might want to abandon anyway. Hence, there is no point to implement that here.

The `flare_angle_at_face` should be easy to constrain, given known `flare_angle_polar` values. That is a TODO.

veprbl:
> * Do we need the png when the svg is included?

png removed.

> * Do we need to keep the PANDA design or can we just shift it up into `unused/`? In particular, do we need a separate configuration, or is `epic_sciglass.xml` just always going to be tiltless?

I've moved PANDA design aside. I'd prefer to delete it in a separate PR to preserve last version for posterity.

compact/ecal_barrel_sciglass_tiltless.xml
wdconinc: Sounds good.

src/SciGlassCalorimeterTiltless_geo.cpp
wdconinc: Ok, we'll put it on the TODO list :-)

wdconinc (dismissed):

wdconinc:
I'm going to set this to auto-merge. Just update the branch to run a last suite of checks, and it will merge.

compact/unused/ecal_barrel_sciglass_panda.xml
```diff
@@ -0,0 +1,82 @@
...
+    Implementation of the projective barrel SciGlass calorimeter
+  </comment>
+
+  <define>
+    <constant name="EcalBarrel_Support_thickness" value="5*cm"/>
+
+    <documentation>
+      The EcalBarrel eta limits cut into the barrel shape.
+    </documentation>
+    <constant name="EcalBarrel_etamin" value="-3.5"/>
```
wdconinc, line 12: I'm implementing the DIRC and wondering about this `etamin` value. This can't really be right, it would be 176 degrees in theta... We need to cut at a different `etamin` so there is room for the DIRC expansion boxes.

wdconinc: In a case of 'design by wishful thinking' it seems to require about `etamin = -1.8` to avoid overlap with where the DIRC people think the DIRC expansion boxes go... In the old PANDA model that doesn't fit the last 4 rows of modules anymore (but I'm not that surprised by that).

veprbl: I also saw that the `etamin=-3.5` is not the acceptance for the current calorimeter and causes no cutout for the envelope if it is set according to the menagerie. The `-1.8` number sounds reasonable to me, we can confirm with GD/I later.

wdconinc: Can you verify that that value does indeed not lead to overlaps of modules with the envelope?

wdconinc:
I'll look into failing pi0 benchmarks today.

wdconinc:
> I'll look into failing pi0 benchmarks today.

Ah, I see, this is a rename of the readout fields in this branch. We are assuming `module` and `slice` in the benchmarks (and only using `module` since there is only 1 slice). Proposal: keep the old id naming scheme for now, and update the naming scheme to something different in a future PR so we can merge this in. Thoughts?

compact/unused/ecal_barrel_sciglass_panda.xml
veprbl: Yes, I will follow up on this

wdconinc (approved):

wdconinc (approved):

osbornjd:
To confirm, is the SciGlass ECal ready to go now that this is merged for the October campaign? @veprbl

wdconinc:
Maybe this was discussed, but we still need the 1.95 cm aluminum in front of the ecal, "EMCal Inner Surface" at https://eic.jlab.org/Geometry/Detector/Detector-20220929172703.html. That will affect resolutions and is important.

osbornjd:
It was not discussed, but we need some more robust way of comparing the actual simulation geometry to the table. There were some instances today where comparisons between the simulation and table were not clear in terms of envelopes or overlaps or whatever.

veprbl:
> To confirm, is the SciGlass ECal ready to go now that this is merged for the October campaign? @veprbl

From the geometry standpoint, it's mostly ready but there are few things that we could improve. I'm not sure how it is on the reconstruction side.

> Maybe this was discussed, but we still need the 1.95 cm aluminum in front of the ecal, "EMCal Inner Surface" at https://eic.jlab.org/Geometry/Detector/Detector-20220929172703.html. That will affect resolutions and is important.

Apparently, that is not the correct interpretation and the table might be updated to reflect that. I'm in contact with our engineers and the project right now to get more accurate and clear information. From what I know now, it might need to go from 2 mm to 3 mm, but not to 2 cm.

PR #46, opened 2022-08-10
.gitignore: only ignore xml files in the repository root, not in compact/

veprbl:
This is a minor convenience fix.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 !*.png
 
-*.xml
+/*.xml
 *.prim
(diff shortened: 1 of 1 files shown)


veprbl:
The `*.xml` is not present in https://github.com/eic/ip6/blob/master/.gitignore, first introduced in https://github.com/eic/epic/commit/4a93ef34d993c0c7202da3418c4ad5eb4d786113

wdconinc (approved):

PR #47, opened 2022-08-10
replace compact/elements.xml with one from DD4hep

veprbl:
```diff
--- compact/elements.xml   2022-08-10 13:35:12.246945134 -0400
+++ $DD4hep_ROOT/DDDetectors/compact/elements.xml   1969-12-31 19:00:01.000000000 -0500
@@ -1,4 +1,3 @@
-<?xml version="1.0" encoding="UTF-8"?>
 <materials>
   <element Z="89" formula="Ac" name="Ac">
     <atom type="A" unit="g/mol" value="227.028"/>
```

I suppose we are not going to be editing this file anytime soon. For example, removing any element leads to weird errors (in unrelated materials) like
```
Fatal in <TGeoMixture::AddElement>: Cannot add element having Z=56 to mixture Barium
```


wdconinc (approved):

wdconinc:
What is the outstanding issue here? I think LCG has DD4hep_INSTALL defined. Does that exist in other DD4hep installS?

wdconinc (approved):

PR #48, opened 2022-08-10
ci: generate slides for Wednesday meetings

wdconinc:
This may help us keep track of new and stale issues, merged and WIP PRs at our weekly meetings... Or it may not.


wdconinc:
I canceled the workflows since we're maxing out the cores again on actual check.

PR #49, opened 2022-08-11
ci: add weekly meeting slide config

wdconinc:
This was forgotten...


wdconinc:
Since CI jobs like this can only be run on main, the omission was only noticed after merging. This config is identical to test cases.

PR #50, opened 2022-08-11
fix: add LGPL license as used by juggler

wdconinc:
- [ ] SPDX top line in all files
- [ ] CI verifier of top line
- [ ] CI verifier of author copyright list

### DIFF
--- a/.github/workflows/copyright.yml
+++ b/.github/workflows/copyright.yml
@@ -9,7 +9,7 @@ jobs:
     - name: Get changed files
       id: files
       run: |
-        diff_tree=$(git diff-tree --no-commit-id --name-only -r ${{ github.event.pull_request.base.sha }} ${{ github.event.pull_request.head.ref}}) 
+        diff_tree=$(git diff-tree --no-commit-id --name-only -r ${{ github.event.pull_request.base.ref }} ${{ github.event.pull_request.head.ref}}) 
         echo "diff_tree=${diff_tree}"
         echo "::set-output name::diff_tree=${diff_tree}"
     - name: Check SPDX license header
--- a/.github/workflows/copyright.yml
+++ b/.github/workflows/copyright.yml
@@ -1,5 +1,4 @@
 on:
-  push:
   pull_request:
 
 jobs:
@@ -10,7 +9,7 @@ jobs:
     - name: Get changed files
       id: files
       run: |
-        diff_tree=$(git diff-tree --no-commit-id --name-only -r ${{ github.sha }})
+        diff_tree=$(git diff-tree --no-commit-id --name-only -r ${{ github.event.pull_request.base.sha }} ${{ github.event.pull_request.head.ref}}) 
         echo "diff_tree=${diff_tree}"
         echo "::set-output name::diff_tree=${diff_tree}"
     - name: Check SPDX license header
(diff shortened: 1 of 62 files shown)


wdconinc:
Top of C++ file
```cpp
// Copyright (C) 2022 Top Authors On Single Line
```

Top of XML file
```xml
<!-- Copyright (C) 2022 Top Authors On Single Line -->
```

wdconinc:
Starting point for copyright author list:
```bash
git ls-files *.cpp *.h *.hpp | while read f ; do authors=$(git blame -w -M -C -C --line-porcelain "$f" | grep -I '^author ' | sort -f | uniq -ic | sort -n --reverse | awk 'NR==1 {ref=$1; $1=$2="";print $0} {if($1>0.1*ref){$1=$2="";print $0}}' | sed 's/^\s*//g' | sed ':x {N;s/\n/, /g; bx}') ; echo -e "// SPDX-License-Identifier: LGPL-3.0-or-later\n// Copyright (C) 2022 $authors\n" | cat - $f > $f.tmp ; mv $f.tmp $f ; done
```
but "CI verifier of author copyright list" will just have to be a check that the committer for a specific file is on the line with the current year, not attempt at ordering (as we lost most of the git history in the athena to ecce transition, sadly, wish I could take that back).

wdconinc:
Bit more escaping needed for xml files:
```console
git ls-files compact/*.xml | while read f ; do authors=$(git blame -w -M -C -C --line-porcelain "$f" | grep -I '^author ' | sort -f | uniq -ic | sort -n --reverse | awk 'NR==1 {ref=$1; $1=$2="";print $0} {if($1>0.1*ref){$1=$2="";print $0}}' | sed 's/^\s*//g' | sed ':x {N;s/\n/, /g; bx}') ; echo -e '<!-- SPDX-License-Identifier: LGPL-3.0-or-later -->'"\n"'<!-- Copyright (C) 2022 '"$authors"' -->'"\n" | cat - $f > $f.tmp ; mv $f.tmp $f ; done
```

plexoos:
Are you sure you can use authors names rather their employing organizations?

sly2j:
AFAIK we can from National Labs if (and only if) it’s a contribution to an open source project and it’s the project policy (both apply here). I’m sure it is the same from other institutions.

sly2j:
It's based off internal documents and training I fear, that's how the Labs work... We looked into this in detail a ~half year ago when this first came up.

sly2j (dismissed):

wdconinc:
Meh
```console
compact/central_tracker.xml:4: parser error : XML declaration allowed only at the start of the document
<?xml version="1.0" encoding="UTF-8"?>
```

wdconinc:
Absence of `<?xml version="1.0" encoding="UTF-8"?>` requires that the xml document is interpreted as `<?xml version="1.0" encoding="UTF-8"?>`, per the [XML 1 specification](https://www.w3.org/TR/xml/#sec-prolog-dtd) ("should") and [XML 1.1 specification](https://www.w3.org/TR/xml11/#sec-prolog-dtd) ("must") and included examples.

c-dilks (approved):

wdconinc:
I've left this linger too long. We should merge this and deal with a CI check later ....

PR #51, opened 2022-08-12
fix (dRICH): swap mirror and sensor construction order

c-dilks:
Swap the construction of mirrors and sensors in `epic/src/dRICH_geo.cpp`. The sensors must be added last, since a call to `CherenkovDetector::CreatePhotonDetectorInstance` will internally add the sensor surface itself to the list of optical boundaries a typical photon will encounter. So the list of boundaries must be in this order:
- vessel entrance
- aerogel entrance
- aerogel exit
- filter entrance
- filter exit
- mirror
- sensor

The way we had it before, mirror and sensor boundaries were switched, and `IRT::Transport` was therefore never seeing the mirror boundary

The variables `sensorCentroidX,Y` and `sensorCount` can no longer be used (without some mild refactoring), and currently aren't being used anyway. They have been commented out, but remain in the code in case we want them for usage in an alternate mirror focusing technique.

### DIFF
--- a/src/DRICH_geo.cpp
+++ b/src/DRICH_geo.cpp
@@ -376,9 +376,9 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
 
   // initialize sensor centroids (used for mirror parameterization below); this is
   // the average (x,y,z) of the placed sensors, w.r.t. originFront
-  double sensorCentroidX = 0;
-  double sensorCentroidZ = 0;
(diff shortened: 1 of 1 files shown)


c-dilks:
We plan to use the geometry service in `EICrecon`, similar to what (some of us) were trying to do in the Juggler `IRTAlgorithm`. Usage of `IRT_AUXFILE` here in `epic` is supposed to be temporary, until we start using `EICrecon` (see #6)

PR #61, opened 2022-08-17
Split HCal systems into different files

rymilton:
Three new XML files are created. hcal.xml now includes the XML file for each hcal system -- similar to ecal.xml. The XML files in the repository root need to be updated to remove the FIXME, but unsure how to do this.


wdconinc:
> The XML files in the repository root need to be updated to remove the FIXME, but unsure how to do this.

That's in `templates/epic.xml`.

wdconinc:
Are there actual changes in the HCAL with this PR? Or just restructuring only?

sly2j:
I’m all for small restructuring changes. Would also prefer all hcal to go in it’s own subdirectory, like I did for tracking (keep main hcal.xml with includes and possible high-level parametrization at the top level). Should make things more manageable going forward.

>

wdconinc:
Also all for restructuring! Was asking mainly to determine how closely to look for changes :-)

rymilton:
This is just restructuring since I saw the fixme in the XML files. Just moves the detector details from hcal.xml to separate XML files.

rymilton:
When I get the chance, I can update the templates file and move the hcal files to their own directory!

wdconinc:
> I have started that pipeline for you.

We need something like this: https://github.com/spack/spackbot/blob/main/spackbot/handlers/pipelines.py.

wdconinc:
For future reference, the way to submit a pipeline by hand is to take the failed submit job, copy the curl command, and submit with the trigger pipeline token from the detector benchmarks Settings > CI/CD > Pipeline Triggers section.

wdconinc (approved):

wdconinc:
All benchmarks succeeded.

PR #65, opened 2022-08-18
Implement forward HCal insert and simplified ECCE LFHCAL

rymilton:
Adding new classes for the forward HCal insert and the surrounding HCal (ECCE design).
HCal needs new class to have insert shape cut out from its middle.

Original implementation from stand-alone studies is [here](https://github.com/rymilton/eic_endcap_insert).

Re-created version of [PR#62](https://github.com/eic/epic/pull/62) to avoid using fork.


rymilton:
I'm also aware that there is an overlap between the insert and the second flange of the beampipe. This needs to be addressed (likely by changing the IP6 files).

rymilton:
I created a pull request to remove the flanges from the forward area of the beampipe ([IP6 PR#16](https://github.com/eic/ip6/pull/16)). This will remove the overlap between the hcal insert and the beampipe.

compact/hcal/hcal_forward.xml
```diff
@@ -1,54 +1,69 @@
...
+      <layer repeat="HcalEndcapPLayer_NSteelRepeat" vis="HcalEndcapLayerVis" thickness="HcalEndcapPSingleLayerThickness">
+        <slice material="Steel235" thickness="HcalEndcapPAbsorberThickness" vis="HcalAbsorberVis"/>
+        <slice material="Polystyrene" thickness="HcalEndcapPPolystyreneThickness" vis="HcalSensorVis" sensitive="yes" limits="cal_limits"/>
+      </layer>
+      <layer repeat="HcalEndcapPLayer_NTungstenRepeat" vis="HcalEndcapLayerVis" thickness="HcalEndcapPSingleLayerThickness">
+        <slice material="Tungsten" thickness="HcalEndcapPAbsorberThickness" vis="HcalAbsorberVis"/>
         <slice material="Polystyrene" thickness="HcalEndcapPPolystyreneThickness" vis="HcalSensorVis" sensitive="yes" limits="cal_limits"/>
       </layer>
+      <insert>
+        <position x="HcalEndcapPInsert_xposition" y="0" z = "0" />
```
wdconinc, line 61: Remove spaces around `=` in z

src/InsertCalorimeter_geo.cpp
```diff
@@ -0,0 +1,207 @@
...
+        +z_distance_traversed goes to the front of each layer,
+        +layer_thickness / 2. is the half-length of a layer
+          (i.e. where to put the cutout shape)
+      */
+
+      SubtractionSolid envelope_with_hole(
+          envelope, layer_hole, Position(hole_x_tracker, 0., (-length + layer_thickness) / 2. + z_distance_traversed));
+
+      // Removing the hole layer by layer from envelope
+      envelope = envelope_with_hole;
```
wdconinc, line 101: I think we should avoid these serial subtraction solids. They are notoriously inefficient since they can't be optimized and always incur a geometry lookup on all components. Can we use a polycone here, which at least makes this a bit more efficient?

src/InsertCalorimeter_geo.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+      z_distance_traversed += layer_thickness; // Moving hole along z
+      hole_x_tracker -= 0.0569;                // Moving hole along x
```
wdconinc, line 104: This, and below, seems like a magical constant that should be kept consistent or is derived from other quantities. Can it be entered through xml?

Also initial_hole_x is similar.

wdconinc (changes requested):

src/InsertCalorimeter_geo.cpp
rymilton: I tried looking into this yesterday evening but I'm pretty lost on how to implement this. Do you have any references/guides on how to use a polycone? We can also discuss this during the office hours tomorrow.

wdconinc: This may be useful: https://github.com/eic/ip6/blob/533d83a4e7c993b9a3b729ba20c89a7b01c41e27/src/IP6BeamPipe.cpp#L143 (but it's also overkill for this here). Essentially you specify rmin/rmax pairs for fixed z planes. So, in your case you'd fill a `std::vector<double> rmax, rmin, z` in the loop like you are already doing, then use those once to create a `Polycone(0, 2.0 * M_PI, rmin, rmax, z)` that uses those rmin/rmax pairs at z planes. If necessary, you can pick rmin always zero, only vary rmax, then finally do a single subtraction from the original `envelope`. It would change `repeat` serial subtractions into just one polycone subtraction from `envelope`.

rymilton: Hmm it seems Polycone doesn't take x and y positions for each segment. Each layer of the insert has a hole with a different radius and at a different x position. Is there a way to account for this easily? Maybe I could make a union of a bunch of circles at different positions and then subtract that? But I'm guessing that causes the same optimization issue as before

rymilton: For documentation: Implemented an assembly instead of an envelope. Polycone doesn't support shifts in x and y. Need twisted cone shape but doesn't exist and time consuming to implement. Using assembly is more efficient than serial subtraction solids.

wdconinc (approved):

PR #67, opened 2022-08-19
updated resolutions in mpgd_barrel.xml from 50um to 150um

mposik1983:
updated resolutions in mpgd_barrel.xml from 50um to 150um


sly2j:
A note copied from Mattermost for future reference (by me): 

> Right now the readout is implemented using faux pixels (we should talk about the correct strip-based implementation). The resolution of a square pixel is pitch/sqrt(12), so seeing as I knew the target resolution I did the inverse of that to get effective pixel size to yield the desired resolution.

sly2j (approved):

PR #68, opened 2022-08-19
POC tag added to mpgd_barrel.xml

mposik1983:
mposik1983 will be POC for mpgd trackers. Added this info to info tag in mpgd_barrel.xml


wdconinc:
Can you rebase and add `status="development"`? It's apparently a mandatory attribute which is why linux-lcg fails.

wdconinc (dismissed):

PR #69, opened 2022-08-20
Bugfix: Tracker bar triangular CarbonFiber frame thickness

ShujieL:
Saying that the triangular structure of silicon staves are made of carbon fiber layer of thickness d. Given the triangle's outer height h, width w,  the inner height of the triangle should be the h - d - d/cos(theta) instead of h - d - d/tan(theta),  where tan(theta)=h/(w/2).


wdconinc (changes requested):
I'm deferring to Sylvester, but can you explain in the pull request description a bit more about what this fix does? That's what will become the commit message for the merge commit, so we would like it to be a bit descriptive.

sly2j (approved):

wdconinc (approved):

PR #77, opened 2022-08-25
updated silicon tracker geometry to LBL design with symmetric disks

ShujieL:
Vertex barrel: removed support cylinder outside the 2nd vertex layer, and connect vertex layers to cone with disks.
Silicon disks: adjusted z positions, and make the backward design symmetric to forward region. 5 disks on each side.

### DIFF
--- a/compact/tracker.xml
+++ b/compact/tracker.xml
@@ -5,7 +5,8 @@
   <define>
     <comment> Primary directions for the tracker </comment>
     <comment> TODO: use actual design angles rather than calculated projective directions </comment>
-    <constant name="TrackerPrimaryAngle"             value="2*atan(exp(-1.109))" />
+    <comment> projective cone at 45 degree </comment>
+    <constant name="TrackerPrimaryAngle"             value="atan(1)" />
     <constant name="TrackerForwardAngle"             value="2*atan(exp(-1.6472))" />
     <constant name="TrackerBackwardAngle"            value="2*atan(exp(-1.5166))" />
 
@@ -18,7 +19,7 @@
     <constant name="SiBarrel1_rmin"                  value="27.0*cm"/>
     <constant name="SiBarrel2_rmin"                  value="42.0*cm"/>
     <constant name="SiBarrel_angle"                  value="TrackerPrimaryAngle"/>
-    <constant name="SiBarrel_dz"                     value="5.0*cm"/>
+    <constant name="SiBarrel_dz"                     value="2*cm"/>
     <comment> barrel length: 2R/tan(angle) - dz </comment>
 
     <comment> Main parameters for the MPGD layer geometries </comment>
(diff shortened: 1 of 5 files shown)


compact/tracker.xml
```diff
@@ -32,8 +32,8 @@
 
     <comment> Service/Support setup </comment>
     <constant name="VertexSupportCyl1_rmin"          value="VertexBarrel_rmin + 2.1*cm"/>
-    <constant name="VertexSupportCyl2_rmin"          value="VertexBarrel_rmax"/>
-    <constant name="VertexSupportCylCF_thickness"    value="300.0*um" />
+    <constant name="VertexSupportCyl2_rmin"          value="VertexBarrel_rmax-0.3*cm"/>
```
wdconinc, line 35: 0.3 cm is magical constant. Needs naming?

ShujieL: done

wdconinc (approved):

compact/tracker/support_service_assembly.xml
```diff
@@ -1,24 +1,35 @@
...
-    <comment> Inner vertex Support </comment>
-    <constant name="VertexSupportCyl1_length"             value="2*VertexSupportCyl1_rmin/tan(TrackerPrimaryAngle)" />
+    <comment> Inner vertex Support</comment>
+    <comment> get rid of the inner vertex support barrel </comment>
+<!--     <constant name="VertexSupportCyl1_length"             value="2*VertexSupportCyl1_rmin/tan(TrackerPrimaryAngle)" />
     <constant name="VertexSupportCyl1_zmax"               value="VertexSupportCyl1_length/2." />
     <constant name="VertexSupportCyl1_thickness"          value="VertexSupportCylCF_thickness" />
-
-    <comment> Outer Vertex Support </comment>
+ -->
```
sly2j, line 13: You can just delete everything related to the inner cylinder if it's not needed

compact/tracker/support_service_assembly.xml
```diff
@@ -91,14 +102,36 @@
...
-      <support type="Cylinder"
+<!--       <support type="Cylinder"
         name="VertexSupportCyl1"
         vis="TrackerSupportVis"
         rmin="VertexSupportCyl1_rmin"
         length="VertexSupportCyl1_length"
         thickness="VertexSupportCyl1_thickness">
           <position x="0*cm" y="0*cm" z="0*cm" />
           <component material="CarbonFiber" thickness="VertexSupportCylCF_thickness" name="Carbon Fiber Shell" vis="TrackerSupportVis"/>
+      </support> -->
```
sly2j, line 113: This too, you can just delete.

sly2j (changes requested):
I'll approve and merge once you remove the dead code. I want to prevent having lots of commented out blocks, we can always add them back later if needed.

wdconinc:
Yeah, I agree with @sly2j, we should just remove code instead of commenting it out. If it is commented out, it is not tested, not updated, and it might as well be not working or even wrong (and that's how it should be treated). So just remove the commented blocks entirely.

compact/tracker/support_service_assembly.xml
ShujieL: done

wdconinc (approved):

PR #72, opened 2022-08-24
Modified material in mpgd_barrel.xml

mposik1983:
Modified material in the inner and outer barrel layers, since exact material budget is not setup yet.
Overall material budget for inner MPGD barrel is now ~0.5% X0 and ~1% X0 for outer barrel layer.

PR #78, opened 2022-08-25
Added POC info to mpgd_barrel.xml

mposik1983:
mposik1983 will be the point of contact for micro pattern gas detectors. This information has been added to mpgd_barrel.xml


wdconinc:
Can you update the branch you're using? This PR is still showing differences in other parts of the file which are not related to this PR (because you use the main branch of your forked repository instead of a single branch). Since those were already merged in https://github.com/eic/epic/pull/72, just updating the branch will make them disappear here.

wdconinc (approved):

PR #83, opened 2022-08-26
ci: add build caching across pipelines

wdconinc:
Building geometry plugins is slow and on the critical path. Speeding up the building by caching is therefore desirable. This adds caching of build products.


wdconinc:

PR #84, opened 2022-08-27
feat(dRICH): remove IRT `auxfile` and add constants to facilitate geometry access from reconstruction

c-dilks:
- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [x] Other: removal of `libIRT` dependence

#### Changes:
- removes optional dependence on `IRT` (auxfile creation is [moved to `drich_dev`](https://github.com/eic/drich-dev/pull/32) until integration of `IRT` with `EICrecon` or `algorithms`)
- adds `Constant`s with the unique name prefix `DRICH_RECON_`, which will be used in the reconstruction, since some geometry parameters are otherwise difficult to access from the `Detector` handle (e.g. the mirror centers)
- some `Constant`s are redundant, but we want it to be clear that any `DRICH_RECON_` `Constant` is required for the reconstruction algorithm; thus any changes to these constants' values may need to be accounted for on the reconstructed side (hopefully only major changes)

- [x] Tests for the changes have been added (in `drich-dev`)
- [x] Documentation has been added / updated (in `drich-dev`)

Only in the IRT reconstruction, which is undergoing refactoring. The IRT geometry creation code is moved to https://github.com/eic/drich-dev/pull/32 and eventually the creation of IRT objects will be integrated into the reconstruction algorithm initialization itself, removing the need for an auxfile altogether (but we can still optionally make auxfiles, as needed).

See above

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,11 +60,6 @@ else()
 endif()
 find_package(fmt REQUIRED)
 
-if(IRT_AUXFILE)
-  add_compile_definitions(IRT_AUXFILE)
-  find_package(IRT REQUIRED)
-endif()
-
 #-----------------------------------------------------------------------------------
 set(a_lib_name ${PROJECT_NAME})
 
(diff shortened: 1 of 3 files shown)


wdconinc:
FYI commented on npdet MR.

c-dilks:
@wdconinc another adjustment in `npdet_info` for TOML compatibility: https://eicweb.phy.anl.gov/EIC/NPDet/-/merge_requests/271

wdconinc (approved):
Looks good to me.

PR #86, opened 2022-08-27
fix: non-zero return for plugin success as of dd4hep-1.22

wdconinc:
As of DD4hep-1.22, https://github.com/AIDASoft/DD4hep/pull/936, any plugin returning a zero return value is assumed to have failed. This is somewhat counter-intuitive for plugins that simply return a long status code, but it is inteded to allow recursive calls and is in line with what other plugins already do, e.g. `return field` or `return sdet`.


veprbl:
SEO:
`runtime_error: dd4hep: apply-plugin: Failed to execute plugin epic_FileLoader`

PR #87, opened 2022-08-27
test(dRICH): validate external IRT auxfile creator

c-dilks:
### Do not merge
Validate #84 and https://github.com/eic/drich-dev/pull/32


c-dilks:
`IRTLOG` output is confirmed to be identical from both this branch and the new `geo_reader` in `drich-dev`. Note the sensors are read out of order, w.r.t. the ordering during creation.

c-dilks:
The momentum dependence of the reconstructed Cherenkov angle and number of photoelectrons are consistent for the two methods, verifying the IRT geometry in the standalone aux file creator is identical with that produced in `epic`.

PR #88, opened 2022-08-29
ci: add 'since' input; scheduled run Monday 16:30Z

wdconinc:
This adds a `since` input to the meeting slides run to allow for deviations from the default '1 week ago' durations. Also changes the scheduled run time to Monday 16:30 UTC, 12:30 EDT.


wdconinc:
Confirmed to work with a manual dispatch.

PR #89, opened 2022-08-31
fix(dRICH): remove `RINDEX` tables from dRICH sensor optical surfaces

c-dilks:
The dRICH sensitive detector now uses `Geant4OpticalTrackingAction` (see https://github.com/AIDASoft/DD4hep/pull/967 and https://eicweb.phy.anl.gov/EIC/NPDet/-/merge_requests/270/). This PR removes `RINDEX` tables from the RICH sensor surfaces, as it's not clear where they came from and they weren't needed in the ATHENA proposal (which effectively also used `Geant4OpticalTrackerAction`).

@alexander-kiselev @chchatte92 @ajitkmaurya 
With these changes, we go from about 60 raw dRICH hits to almost 300, for a 40 GeV pion.

No

Not beyond the dRICH

### DIFF
--- a/src/DRICH_geo.cpp
+++ b/src/DRICH_geo.cpp
@@ -40,7 +40,7 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
   xml::Component        dims    = detElem.dimensions();
   OpticalSurfaceManager surfMgr = desc.surfaceManager();
   DetElement            det(detName, detID);
-  sens.setType("opticaltracker");
+  sens.setType("tracker");
 
   // attributes, from compact file =============================================
   // - vessel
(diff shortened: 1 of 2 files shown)


src/DRICH_geo.cpp
```diff
@@ -40,7 +40,7 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
   xml::Component        dims    = detElem.dimensions();
   OpticalSurfaceManager surfMgr = desc.surfaceManager();
   DetElement            det(detName, detID);
-  sens.setType("tracker");
+  sens.setType("opticaltracker");
```
wdconinc, line 43: The way we should roll this out is probably by keeping the original tracker type and setting the SD action in npsim.py. Otherwise we break ddsim functionality for all released versions.

c-dilks: Yes, I will try again later using `tracker`. I didn't notice the new `npsim` until after adding `opticaltracker` to DD4hep; and after doing this, it's clear that wasn't the best way.

c-dilks: reverted to `tracker`

wdconinc (approved):

PR #96, opened 2022-09-02
A few adjustments for material map

ShujieL:
modified a few tracker xml to update one disk radius and some layer_material binning for Acts material map.


sly2j (approved):

wdconinc:
https://github.com/eic/epic/runs/8150956415?check_suite_focus=true overlap...

sly2j (approved):

PR #102, opened 2022-09-03
ci: auto-update pull requests

wdconinc:
We can set pull requests to automatically squash merge, but we can NOT set it to automatically rebase when the base branch receives a push.

A new automatic workflow triggered by pushes to the main branch that updates outstanding pull requests which are set to auto-merge.

- [ ] Bug fix (issue #__)
- [X] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No changes to geometry code.

This may introduce automatic pull request updates, which will require users to know how to deal with git pulls after a remote branch has been rebased and force pushed. However, this is aimed at branches that are set to auto-merge and therefore intended to have completed development. The use case where this may be an issue is for trivial changes where auto-merge is selected before first checks complete, those checks then fail, a push to main triggers and auto-update, and the user want to fix the code on a locally checked out copy.

No changes to default behavior of geometry code.

### DIFF
--- a/.github/workflows/auto-update.yml
+++ b/.github/workflows/auto-update.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - auto-update-pull-requests
 
 jobs:
   auto-update-pull-requests:
--- a/.github/workflows/auto-update.yml
+++ b/.github/workflows/auto-update.yml
@@ -1,0 +1,15 @@
+name: Auto-update pull requests
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  auto-update-pull-requests:
+    name: Auto-update pull requests
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


wdconinc:
@veprbl See demonstration of functionality with #103. This only applies auto-update to those pull requests that are already set to auto-merge (i.e. waiting for review or for checks to complete). This removes a last unnecessarily manual operation from an otherwise already automated workflow.

.github/workflows/auto-update.yml
```diff
@@ -0,0 +1,16 @@
+name: Auto-update pull requests
+
+on:
+  push:
+    branches:
+      - main
+      - auto-update-pull-requests
```
veprbl, line 7:
```suggestion
```

wdconinc:
It also triggers the actions with the update.

PR #106, opened 2022-09-08
fix: dRICH sensor material should not be `AirOptical`

### LINKED ISSUE #18: dRICH: sensor material should not be `AirOptical`
This was temporary and needs to be updated
- moreover, I had to add reflectivity properties to the dRICH sensor surface from the ATHENA `master` branch, in order to see hits; not sure if this is correct (since we did not have this in the ATHENA proposal production)

#### Tasks:
- [ ] check the photon count, to make sure we get the expected numbers
- [ ] check the sensor geometry and segmentation
- [ ] surface paramaterization (esp. reflectivity)
- [ ] sensor bases (how to model the back of the sensors?) - make sure there is enough space for this

**Additional context**
#### Notes from ATHENA studies (from discussions with Alexander, Chandra, and Chris):

- sensor used for proposal: https://www.hamamatsu.com/us/en/product/optical-sensors/mppc/mppc_mppc-array/S13361-3050AE-08.html

- the non-negligible gas-to-resin (numerically
<5%) and resin-to-silicon (~15%) reflections must be effectively
accounted in the quoted PDEs. At all incident angles of interest
for us (up to 30 degree or so) there numbers get shared between
the polarization states, but average stays more or less the same.

- However in our case (g4dRIChOptics.hh) the sensor surface was defined as dielectric_metal with
some bogus imaginary refractive index parameters. This killed the
photons, and also I verified that the MC .root files contain only
the ones which were detected somewhere. So I temporarily changed the material to AirOptical, and added
a benign surface to the optical_metarials.xml database, and we are
back at ~10 npe. And "Chandra's number" for 350..650nm integral is
now ~72 or so.

- I think the correct way to account Cherenkov photon polarization is
to create a resin volume, and perhaps even a silicon volume inside it,
but renormalize the PDE, accounting for the normal incident losses.

- look into `G4SiPM` package

c-dilks:
Close #18 

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [x] Other: improve material property tables

No

No


kkauder:
@c-dilks @chchatte92 Is this issue solved? When this file moved, git didn't understand that and marked it as deleted, and the new one is treated as a different one, so you can't easily compare/merge. Also, the current one looks to my eye substantially different, so I can't easily identify which instance of AirOptical needs (or needed) to be changed.

rahmans1:
We are tagging a new release for the February campaign soon. Is there anyone working on this PR? Is it expected to be ready for February release?

chchatte92:
> @chchatte92 We are tagging a new release for the February campaign soon. Is there anyone working on this PR? Is it expected to be ready for February release?

Dear Sakib, I will get back to you soon (either later today or in the weekends)

PR #113, opened 2022-09-14
fix(dRICH): correct the sensor segmentation offset and grid size

### LINKED ISSUE #108: dRICH: check and improve the readout `cellID` bit fields
**Is your feature request related to a problem? Please describe.**
Check to make sure if this is what we really want:
https://github.com/eic/epic/blob/11c1e541a46dce024c5bef81fc340a9e1cb0f498/compact/drich.xml#L244-L251

We only use `sector` and `module` downstream in the current reconstruction code, but we should use `x` and `y` too. Any changes here in `epic` need to be reflected (and validated) in https://github.com/eic/drich-dev/blob/main/src/draw_segmentation.cpp

For `x` and `y`, ideally we want 0-7, unsigned, and (maybe?) the highest 32 bits.

**Describe alternatives you've considered**
Ignoring the segmentation in reconstruction, and doing our own (which is what we are currently doing...)

**Additional context**
Be careful of off-by-one errors. In ATHENA, we actually were running with a segmentation of 9x9, thinking it was 8x8 (but it didn't matter, since again we did our own segmentation in the reconstruction code). This off-by-one issue is the reason for the `-1` in the denominator of `grid_size_x,y`

c-dilks:
Close #108 

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

  - testing externally with https://github.com/eic/drich-dev/blob/geo-reader/src/draw_segmentation.cpp
  - see #108 for notes and more details

No

No

### DIFF
--- a/compact/drich.xml
+++ b/compact/drich.xml
@@ -222,19 +222,9 @@
 <documentation level="10">
 #### Readout
 - segmentation: square matrix of pixels
-  - `grid_size_x,y`: size of each sensor, but note we must divide sensor size
-    by 1 less than the number of pixels, to account for fenceposting
-  - `offset_x,y`: specified such that the `x` and `y` indicators are unsigned
-- indicators and `cellID` bits:
-
-  | indicator | offset | length |
-  |-----------|--------|--------|
-  | dRICh ID  | 0      | 8      |
-  | sector    | 8      | 3      |
-  | sensor    | 11     | 12     |
-  | x pixel   | 23     | 16     |
-  | y pixel   | 39     | 16     |
-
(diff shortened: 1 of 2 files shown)


wdconinc (approved):

c-dilks:
A DIS physics benchmark failed (bad Q2 fit). Doesn't seem to be this PR's fault.

wdconinc:
> A DIS physics benchmark failed (bad Q2 fit). Doesn't seem to be this PR's fault.

Yeah, it's not your problem. I'll override-merge this.

Issue #110, opened 2022-09-12
Update barrel hcal geometry to match sPHENIX oHCAL


johnlajoie:
The EPIC barrel HCAL needs to be updated to match the sPHENIX oHCAL geometry.  The "as-built" geometry is available in a set of gdml files that are used directly in G4 in the sPHENIX Fun4All.  The nice part about the way this is implemented is that the absorber structures are defined separately (in separate gdml files) from the sensitive elements (the tiles). I think this means it should be possible to handle the fact that gdml doesn't include the ability to flag sensitive elements and wrap the gdml in an xml that dd4hep will be happy with.

Based on the tutorials over the past two weeks I have started some work on this.

PR #117, opened 2022-09-16
starting point with HCAL sector steel

johnlajoie:
This PR is the start of introducing the correct oHCAL geometry for EPIC associated with re-use of the sPHENIX outer hadronic calorimeter. At the present time it just includes the oHCAL sector steel, including both the normal HCAL sectors and the three modified "chimney" sectors that have a cutout to accommodate the solenoid cryogenic stack. 

While the sPHENIX simulations use a very complicated, full geometry for the oHCAL based on a gdml export of the CAD files for the as-built sectors, the geometry here is drastically simplified to include only the longitudinal steel plates.  The endplates, sector combs, and endplates have been removed as these are small perturbations that are unlikely to affect the overall performance. 

I have, however, decided to continue with a gdml export for the simplified geometry. The reason for this is that at the present time the simplified geometry has a very simple mesh, and it is straightforward to implement in a generic way in the xml using a TesslelatedSolid.  The geometry is created very quickly and the root viewer seems to have no problem with the exported geometry.  This approach allows us to add additional relaism to the HCAL sectors by just updating the xml (gdml) for the solid with no additional code changes. For example, the combs or endplates with their curved surfaces would be added in trivially if it was though to be necessary/interesting in the future. 

The geometry is generated by importing the STEP file from the CAD design and using the GDML wokrbench in FreeCAD to mesh the solid and write out the gdml. The workbench has additional tools, like the ability to decimate and/or simplify the mesh if desired.  The resulting gdml is added into the xml description for the detector. 

The work that remains for the geometry is to: 

- Add in the scintillation tile shapes. There are 12 shapes, which an additional special set of shapes in the chimney sector. 
- Modify the stepping action to include the measured light yield maps. This is not strictly necessary, but I would like to do it now so it doesn't get forgotten in the future (or the information lost)

After this, there will need to be some work at the reconstruction level to collect the scintillating tiles in groups of five corresponding to the preamps on the calorimeter towers. 

Here's a viewer image showing the beam pipe, solenoid and the new oHCAL. The cryo cutout is visible at the top left. 

example

- [ ] Bug fix (issue #__)
- [ X] New feature (issue #110)
- [ ] Documentation update
- [ ] Other: __

At the present time this should be independent of the rest of the detector geometry as the new outer HCAL geometry is stored in a new xml file, and a new type "epic_oHcalBarrel" was created for this development.  This will aloo current simulations to continue until we are ready to switch geometry. 

Yes - it replaces the existing ePIC HCAL geometry witha more realistic version based on the sPHENIX HCAL.


wdconinc:
Sorry, accidentally clicked the 'ready for review' button.

johnlajoie:
All tiles shapes are now in place and the tile assemblies that go between the gaps are created and properly rotated next step will be to get the tile assembly placement between the gaps completed and checked for overlaps. 
Untitled

johnlajoie:
Latest version of the ePIC outer HCAL geometry.  New to this version is that instead of building the full sectors as tessellated solids, only the plates (half and full) are read in as solids and the sectors constructed from the plates. This eliminated the errors from TGeo when doing an overlapCheck: 

`Error in <TGeoTessellated::GetPointsOnSegments>: You should require at least 4160 points`

This error appeared to be related to TGeo not liking tessellated solids that are comprised of multiple disconnected solids (like the sector made up of separate plates). Reading in the plates and building the sectors as assemblies solved this. 

This version also has the tile locations adjusted so the tile assemblies form flat planes at the proper tilt angle. 

Unfortunately, I have not been able to get around the completely spurious overlap messages when using tessellated solids. Plates that are clearly not overlapping and are separated by many cm still show an overlap in overlapCheck. On Friday Dmitry Kalinkin worked with me on debugging these overlaps in a test case. There are no disconnected vertices in the solids that are created, and CloseShape() with error checks set reports no errors and no flipped normals. In a two-plate test example, the "overlaps" that appear to be flagged are physically separated in the viewer by several cm.  Exporting the TGeo geometry to gdml and reading that back into a CAD program works fine, and the geometry looks OK, except the CAD reports three volumes instead of two (not clear why or if this is spurious?)  

As another bit of information, checkOverlaps.py segfaults G4.

At this point it looks to me like the solids are created correctly (they display properly in the viewer) but there is a bug in TGeo in handling these solids in the detector geometry and exporting them to G4. Obviously this will need to be solved before I can go any further with this, and I will have to consult with more experts to understand how to figure this out. 

Since everything in the calorimeter is parallel tilted planes, it's fairly easy to check visually for overlaps and that is what I have done in tuning all the final positions in this version.  However, with the completely spurious overlapCheck messages it's not possible to be certain that there are not some real overlaps remaining. 

HCAL_2Oct2022

src/BarrelHCalCalorimeter_geo.cpp
```diff
@@ -0,0 +1,452 @@
...
+
+  std::string   env_vis = getAttrOrDefault<std::string>(x_det, _Unicode(env_vis), "HcalBarrelEnvelopeVis"); 
+  envelope.setAttributes(description, x_det.regionStr(), x_det.limitsStr(), env_vis);
+  return sdet;
+
+}
+
+#ifdef EPIC_ECCE_LEGACY_COMPAT
+DECLARE_DETELEMENT(ecce_oHcalBarrel, create_detector)
+#endif
```
veprbl, line 446:
```suggestion
```

src/BarrelHCalCalorimeter_geo.cpp
```diff
@@ -0,0 +1,452 @@
...
+DECLARE_DETELEMENT(epic_oHcalBarrel, create_detector)
+#ifdef EPIC_ECCE_LEGACY_COMPAT
+DECLARE_DETELEMENT(ecce_HcalBarrel, create_detector)
+#endif
```
veprbl, line 450:
```suggestion
```

veprbl (commented):
No need to add legacy declarations

johnlajoie:
Two overlap messages remaining in G4: 

`-------- WWWW ------- G4Exception-START -------- WWWW -------
*** G4Exception : GeomVol1002
      issued by : G4PVPlacement::CheckOverlaps()
Overlap with volume already placed !
          Overlap is detected for volume AV_114!Sector_17#17!OuterHCalTile04_0#0:2718 (G4TessellatedSolid) with AV_114!Sector_17#17!HCAL_Sector_Plate_6#6:26 (G4TessellatedSolid)
          overlap at local point (280.019,-197.81,-732.823) by 25.6577 cm  (max of 1 cases)
NOTE: Reached maximum fixed number -1- of overlaps reports for this volume !
-------- WWWW -------- G4Exception-END --------- WWWW -------


-------- WWWW ------- G4Exception-START -------- WWWW -------
*** G4Exception : GeomVol1002
      issued by : G4PVPlacement::CheckOverlaps()
Overlap with volume already placed !
          Overlap is detected for volume AV_128!Sector_31#31!OuterHCalTile03_4#4:3936 (G4TessellatedSolid) with AV_128!Sector_31#31!OuterHCalTile03_0#0:3932 (G4TessellatedSolid)
          overlap at local point (196.949,230.155,164.35) by 15.7363 cm  (max of 1 cases)
NOTE: Reached maximum fixed number -1- of overlaps reports for this volume !
-------- WWWW -------- G4Exception-END --------- WWWW -------

`
These overlaps are ridiculously large, which makes me suspicious. It's also odd that the overlaps only occur for Sector 17 and 31 but all sectors are a copy of the same structure. Will work to eliminate these later today.  

Would appreciate it if someone could take a look at the readout structure and see that it makes sense?  I am trying to make the readout unit as a single tile,.

src/BarrelHCalCalorimeter_geo.cpp
johnlajoie: completed

src/BarrelHCalCalorimeter_geo.cpp
johnlajoie: completed

veprbl:
This might be not a good time to bring this up, but would you consider loading GDML directly, like it was done in https://github.com/veprbl/epic/commit/d8ee9d2427a92a542afadb5bfbf9d981fe805f9a ?

veprbl:
Good point, locating the xml files is definitely a concern. As far as managing several files, I don't think that's a requirement. One can always load one gdml file and pick out volumes from the TGeo sub-nodes.

johnlajoie:
> Good point, locating the xml files is definitely a concern. As far as managing several files, I don't think that's a requirement. One can always load one gdml file and pick out volumes from the TGeo sub-nodes.

One thing I found early on was that I got a TGeo error message when working with a geometry (as shapes or a gdml file) that had multiple, disconnected solids in the shape - this is why I went to loading the plates separately and reconstructing the sector instead of just loading the sector.  I haven't gone back and re-checked this with all the new changes, however. 

So yes - I think we could put everything into one gdml file.  Even easier would be to create the sectors as single complex objects and then just place the tiles. 

I'd be happy to look at changing this if people think it would be easier to maintain, but I'd like to see if I can understand the remaining issue in the current geometry first.

veprbl:
> I'd be happy to look at changing this if people think it would be easier to maintain, but I'd like to see if I can understand the remaining issue in the current geometry first.

Actually, my thinking was that the next issue around the corner can always be a bug in that long code. Although TGeo GDML reader is, arguably, a long code as well.

johnlajoie:
All barrel HCAL overlaps have been resolved. 

Ready for review - this adds the barrel oHCAL geometry to the repository but doesn't include it in ePIC by default yet.

compact/bhcal.xml
```diff
@@ -0,0 +1,14 @@
...
+  <define>
+  </define>
+
+  <documentation level="10">
+    ### Barrel Hcal configuration
+  </documentation>
+
+  <include ref="hcal/hcal_barrel_gdml.xml"/>
+
+</lccdd>
```
wdconinc, line 14: We probably don't need this file. Most subsystems are labeled 'little endian': e.g. `ecal_barrel_sciglass.xml`. So here I'd probably just keep the name `hcal_barrel_gdml.xml` in line with that approach.

compact/hcal/hcal_barrel_gdml.xml
```diff
@@ -0,0 +1,7682 @@
...
+      <detector
+	  id="oHCalBarrel_ID"
+	  name="oHcalBarrel"
+	  type="epic_oHcalBarrel"
+	  readout="oHcalBarrelHits"
+	  vis="HcalBarrelVis"
+	  env_vis="HcalBarrelEnvelopeVis"
+	  rmin="1767"
+	  rmax="2690"
+	  length="3170"
```
wdconinc, line 35: Not sure if you are using this to construct an envelope volume, but it should use the HCal parametrization volumes from definitions.xml. That allows us to apply any future shifts centrally without having to edit all detector files.

compact/hcal/hcal_barrel_gdml.xml
```diff
@@ -0,0 +1,7682 @@
...
+        <position name="OuterHCalTile05_EJ200_v128" unit="mm" x="486.33" y="798.5449" z="7.0"/>
+        <position name="OuterHCalTile05_EJ200_v129" unit="mm" x="486.7111" y="798.5853" z="1.5"/>
+        <position name="OuterHCalTile05_EJ200_v130" unit="mm" x="486.7111" y="798.5853" z="7.0"/>
+        <position name="OuterHCalTile05_EJ200_v131" unit="mm" x="487.0695" y="798.7209" z="1.5"/>
+        <position name="OuterHCalTile05_EJ200_v132" unit="mm" x="487.0695" y="798.7209" z="7.0"/>
+        <position name="OuterHCalTile05_EJ200_v133" unit="mm" x="487.3818" y="798.9429" z="1.5"/>
+        <position name="OuterHCalTile05_EJ200_v134" unit="mm" x="487.3818" y="798.9429" z="7.0"/>
+        <position name="OuterHCalTile05_EJ200_v135" unit="mm" x="487.6277" y="799.2367" z="1.5"/>
+        <position name="OuterHCalTile05_EJ200_v136" unit="mm" x="487.6277" y="799.2367" z="7.0"/>
+        <position name="OuterHCalTile05_EJ200
```
wdconinc, line 7680: Remove emtpy plugins environment.

compact/definitions.xml
```diff
@@ -185,15 +185,17 @@ The unused IDs below are saved for future use.
...
-      - Unused IDs: 116-119
+      - Unused IDs: 117-119
     </documentation>
     <constant name="HCalSubAssembly_ID" value="110"/>
     <constant name="HCalBarrel_ID"      value="111"/>
     <constant name="HCalEndcapP_ID"     value="112"/>
     <constant name="HCalEndcapN_ID"     value="113"/>
     <constant name="PassiveSteelRingEndcapP_ID" value="114"/>
     <constant name="HCalEndcapPInsert_ID" value="115"/>
+    <constant name="oHCalBarrel_ID"      value="116"/>
```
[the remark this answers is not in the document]
johnlajoie: I was concerned about downstream effects of re-using the HCALBarrel_ID - I assume this keys into the reconstruction at some point and teh structure of the real HCAL is quite different from the ATHENA HCAL.  I thought it might be better to separate them to allow the possibility to have both for a limited time, in case this crashes and burns at the reconstruction stage and needs a lot more work to be useful. 

Happy to just change it to HCALBarrel_ID if it won't cause dramatic downstream disruption.

compact/hcal/hcal_barrel_gdml.xml
```diff
@@ -0,0 +1,7682 @@
+<lccdd>
+
+    <limits>
+    </limits>
+
+    <regions>
+    </regions>
+
+    <display>
+    </display>
```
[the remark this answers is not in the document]
johnlajoie: Done - will be part of next commit.

compact/hcal/hcal_barrel_gdml.xml
```diff
@@ -0,0 +1,7682 @@
...
+    <triangular vertex1="Mesh2Tess_14" vertex2="Mesh2Tess_3" vertex3="Mesh2Tess_13" type="ABSOLUTE"/>
+    <triangular vertex1="Mesh2Tess_14" vertex2="Mesh2Tess_13" vertex3="Mesh2Tess_11" type="ABSOLUTE"/>
+    <triangular vertex1="Mesh2Tess_2" vertex2="Mesh2Tess_3" vertex3="Mesh2Tess_14" type="ABSOLUTE"/>
+    <triangular vertex1="Mesh2Tess_2" vertex2="Mesh2Tess_7" vertex3="Mesh2Tess_3" type="ABSOLUTE"/>
+    <triangular vertex1="Mesh2Tess_2" vertex2="Mesh2Tess_9" vertex3="Mesh2Tess_1" type="ABSOLUTE"/>
+    <triangular vertex1="Mesh2Tess_2" vertex2="Mesh2Tess_14" vertex3="Mesh2Tess_9" type="ABSOLUTE"/>
+  </tessellated>
+  </solid>
+	
+  <solid name="OuterHCalTile01" material="EJ200" unit="mm" x="0.0" y="0.0" z="0.0">
```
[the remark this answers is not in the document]
johnlajoie: They can!  I'll need to get the STEP files from the engineers and then regenerate the gdml, but I have a way to do this in FreeCAD (which is how I did the plates).   I held off on this because I was working on the geometry and overlap issues.  

Will request the STEP files and get started on this, but it may have a timescale that goes into next week.

compact/hcal/hcal_barrel_gdml.xml
```diff
@@ -0,0 +1,7682 @@
...
+
+    <regions>
+    </regions>
+
+    <display>
+    </display>
+
+    <materials>
+    <!-- EJ200 (Plastic scintillator) -->
+    <material name="EJ200" state="solid">
```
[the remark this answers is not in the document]
johnlajoie: Done - will be part of next commit.

compact/hcal/hcal_barrel_gdml.xml
```diff
@@ -0,0 +1,7682 @@
...
+    </material>
+    </materials>
+
+    <!-- Define detector -->
+    <detectors>
+      <documentation>
+      ### Barrel Hadronic Calorimeter
+      </documentation>
+      <detector
+	  id="oHCalBarrel_ID"
```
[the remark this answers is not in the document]
johnlajoie: See above - I did this to distinguish from the existing HCAL ID.  That may or may not be relevant.

compact/bhcal.xml
johnlajoie: I created this file to have a configuration that allowed me to debug the new barrel hcal development without interaction with all the other development going on. Are you suggesting it disappear, or it be renamed hcal_barrel_gdml.xml?

compact/hcal/hcal_barrel_gdml.xml
johnlajoie: Not sure I understand this comment? What do you want removed?

compact/bhcal.xml
johnlajoie: Nevermind - I understood this question in the context of another question below

compact/hcal/hcal_barrel_gdml.xml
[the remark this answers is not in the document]
johnlajoie: Have the STL source, I can work on this over the weekend.

templates/epic.xml.jinja2
```diff
@@ -176,6 +176,13 @@
   <include ref="${DETECTOR_PATH}/compact/hcal.xml"/>
 {% endif -%}
 
+{% if features is not defined or 'bhcal' in features %}
+  <documentation level="10">
+    ## Central hadronic calorimetry
+  </documentation>
+  <include ref="${DETECTOR_PATH}/compact/bhcal.xml"/>
+{% endif -%}
+
```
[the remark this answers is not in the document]
johnlajoie: Completed and committed.

compact/hcal/hcal_barrel_gdml.xml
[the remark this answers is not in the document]
johnlajoie: Changed and committed.

compact/definitions.xml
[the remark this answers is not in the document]
johnlajoie: Change made and committed

compact/hcal/hcal_barrel_gdml.xml
```diff
@@ -0,0 +1,7682 @@
...
+    <constant name="csectorRotate" unit="deg" value="6.086"/>
+    <constant name="sectorRotate" unit="deg" value="6.100"/>
+    
+    <matrix name="plates_x" unit="mm" coldim="1" values="161.935 118.175 -741.606 -771.303"/>
+    <matrix name="plates_y" unit="mm" coldim="1" values="2295.048 2267.198 2160.230 2119.176"/>
+    <matrix name="plates_z" unit="mm" coldim="1" values="235.587 253.54 0.0 0.0"/>
+    
+    <matrix name="tweak_tiles" unit="rad" values="0.0 0.0 0.0 0.0 0.0 0.0001 0.0 0.0 0.0 0.0 0.0001 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0002"/> 
+    <matrix name="tweak_chimney_tiles" unit="rad" values="-0.0002 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0"/> 
+    <matrix name="tweak_sectors" unit="rad" values="0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 -0.00008 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0"/>
```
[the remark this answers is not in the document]
johnlajoie: Done.

compact/hcal/hcal_barrel_gdml.xml
johnlajoie: Ah - found it, done.

wdconinc: My browser was struggling with the long file, so I think it misplaced the comment...

compact/hcal/hcal_barrel_gdml.xml
johnlajoie: Updated to use HCAL parametrization volumes

compact/hcal/hcal_barrel_gdml.xml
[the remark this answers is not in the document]
wdconinc: The readouts section is intended, e.g., to implement very large numbers of fine pixels in an otherwise homogeneous material. The geometry would be slow as molasses in Winnipeg if we were to implement those pixels in geometry. So, the volume is segmented 'outside' of the geometry. DD4hep makes the single volume sensitive, then figures out from the local hit position which pixel was hit, and assigns that as a 64-bit cell ID. Each 64-bit cell ID is unique, since it also stores the detector ID (`system`) and other components such as `sector`, `tower`, etc.

If you don't need the finer segmentation than what is provided at the volume level, you can use the `NoSegmentation` type, like the backward ecal (`compact/ecal_backward_PbWO4.xml`):
```xml
    <readout name="EcalEndcapNHits">
      <segmentation type="NoSegmentation" key="sector"/>
      <id>system:8,sector:4,module:20</id>
    </readout>
```
In this case, there is no need for the `x` and `y` fields since there is no XY segmentation at the tower level.

There are signal merging reconstruction algorithms in case the signals from multiple towers are combined in hardware before digitization (multiple fibers to the same PMT is the usual example).

compact/hcal/hcal_barrel_gdml.xml
johnlajoie: Unfortunately this leads to overlaps because the rmin defined this way is too small for the HCAL sectors:

`    
<constant name="HcalBarrel_rmin"      value="ForwardServiceGap_rmax"/>
`

resolves to 195cm.  I believe that this is too small because in this region the plates in the HCAL are smaller radially, so the ForwardServiceGap_rmax is too large and near near the olenoid the plates protrude from the volume.  

I updated this to: 

`<constant name="HcalBarrel_rmin"      value="Solenoid_rmax"/>`

in definitions.xml.

compact/hcal/hcal_barrel_gdml.xml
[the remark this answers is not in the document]
johnlajoie: Changing the readout name to HcalBarrelHits gives a runtime error: 

`runtime_error: dd4hep: Attempt to add an already existing object:HcalBarrelHits.`

in the "linux-lcg / build-lcg-ubuntu-2004 (LCG_101/x86_64-ubuntu2004-gcc9-opt) (pull_request) " check.  Presumably a conflict with the same readout name in the old Hcal code?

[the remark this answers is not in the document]
johnlajoie: Had both versions of the HCAL in place due to messed up jinja template.  FIxed and committed.

compact/hcal/hcal_barrel_gdml.xml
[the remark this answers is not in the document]
johnlajoie: After consulting with Wouter, updated readout to:

`  <readouts>
    <readout name="HcalBarrelHits">
      <segmentation type="NoSegmentation"/>
      <id>system:8,sector:6,tower:6</id>
    </readout>
`

johnlajoie:
It looks like there is an overlap between the dRICH and the HCAL envelope volume.  This is because on the notch in the HCAL plates on the end which means the inner radius should be larger in the middle and smaller on the ends, so probably the union of two tubes is what is required.

wdconinc:
Or a polycone which is slightly better than a union.

src/BarrelHCalCalorimeter_geo.cpp
```diff
@@ -0,0 +1,481 @@
...
+
+  double rmin1 = x_det.rmin1(); 
+  double rmin2 = x_det.rmin1(); 
+  double rmax = x_det.rmax(); 
+  double length1 = x_det.z1(); 
+  double length2 = x_det.z2();
+  
+  std::vector<double> rmins; 
+  rmins[0] = rmin1; 
+  rmins[1] = rmin2;
```
wdconinc, line 53: Suggestion to use an initializer list (since c++11):
```suggestion
  std::vector<double> rmins = {rmin1, rmin2};
```
Same with `rmaxs` and `zs` below.

johnlajoie: Done. Previous version passed overlap checks for epic.xml.  Looking forward to a lean bill of health from the next set of checks....

wdconinc:
> Looking forward to a lean bill of health from the next set of checks....

Well, there's a whole other set of benchmarks that only start running when the overlaps are coming back clean. I anticipate those will fail due to the different structure of this hcal compared to the previous one...

wdconinc:
See e.g. https://eicweb.phy.anl.gov/EIC/benchmarks/reconstruction_benchmarks/-/blob/master/benchmarks/clustering/options/full_cal_reco.py#L328 where we are using assumptions of `layers` and `slices` in how clusters are reconstructed. That will have been simply copied into EICrecon (at best) so it will need some updating. In order to do this in a somewhat backwards compatible way that allows other work to continue unaffected, we should probably introduce a new configuration, similar to imaging and sciglass, that contains the new hcal, and then branch in the benchmarks when that configuration is used. Once the benchmarks are behaving fine, the benchmark defaults can be swapped around to assume the new hcal is the default. I can help with this.

johnlajoie:
I expected that the reconstruction would be an issue one we got the new geometry set. I'll start digging into this tomorrow, and I've also go a new postdoc who is working furiously to come up to speed on EICrecon.

johnlajoie:
> See e.g. https://eicweb.phy.anl.gov/EIC/benchmarks/reconstruction_benchmarks/-/blob/master/benchmarks/clustering/options/full_cal_reco.py#L328 where we are using assumptions of `layers` and `slices` in how clusters are reconstructed. That will have been simply copied into EICrecon (at best) so it will need some updating. In order to do this in a somewhat backwards compatible way that allows other work to continue unaffected, we should probably introduce a new configuration, similar to imaging and sciglass, that contains the new hcal, and then branch in the benchmarks when that configuration is used. Once the benchmarks are behaving fine, the benchmark defaults can be swapped around to assume the new hcal is the default. I can help with this.

Before I jump in I wanted to make sure I had a full picture in mind, since this is a change from what we previously decided.  I used to have two barrel hcal configurations.  What is in the pull request right now would just replace the old with the new. 

So I think what you want is: 

- A new configuration, call in "newhcal.yml" and "newhcal.xml" that references the new barrel hcal. 
- Modify the jinja template to use the old hcal config (by default)

I think that would not break the current reconstruction and allow use to put together the new reconstruction in parallel.

wdconinc:
Right, I think that's the approach that will allow us to merge this in without having to fix the reconstruction first.

johnlajoie:
The corresponding changes for EICrecon have been added as a pull request (EICrecon 260): 

https://github.com/eic/EICrecon/pull/260

The EICrecon changes basically change the default Barrel HCAL reconstruction behavior, but the pull request for the detector geometry does not make the new barrel HCAL the default.  This will be updated with a new commit, and then this should be ready to merge.  

Ideally both this PR and the EICrecon PR should be merged at once to avoid breaking behavior.

johnlajoie:
New version just committed with completely regenerated meshes for the tiles, now following exactly what was done for the sector plates.  This reduces the mesh size slightly for the tiles, but when generating the solids the COM of the solid is interior to the solid.  Because i could regenerate the mesh referenced to the coordinate system I chose, the calculations to place the tiles in the sectors are a bit simpler. 

The chimney sector and ordinary sectors, by themselves, pass the G4 overlap check down to 1um.  The TGeo check for the sectors by themselves claims many overlaps at the 0.13-0.15mm level. Not sure which to believe. While i had been concentrating on the G4 ovberlaps up to now, it may be better to try to eliminate the TGeo overlaps first? 

Essentially, this is the same situation as before and all the effort to refine the tile geometry has not yielded an improvement.

johnlajoie:
Making relatively large changes to the positioning of the tiles in the gaps make no change to the overlaps reported by TGeo.  Nothing I do seems to have a significant effort on what the TGeo overlap check reports, for better or worse.

wdconinc (approved):

wdconinc:
As agreed in today's meeting, we'll merge this and deal with benchmark breakage next.

PR #122, opened 2022-09-21
feat: introduce MPGD DIRC bars and update hpDIRC model and parameters

wdconinc:
This introduces the hpDIRC with the correct geometry and the integrated MPGD DIRC bars (courtesy @mposik1983).

- [ ] Bug fix (issue #__)
- [X] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

Yes, new hpDIRC and MPGD DIRC detectors by default.


wdconinc:
Easy to pass the overlap checks if the DIRC is not actually included in the geometry that is checked...

wdconinc:
Outstanding issues for `definitions.xml`:
- [ ] introduce `DIRC_thickness`
- [ ] introduce `OuterBarrelMPGD_rmin` as `DIRC_rmin + DIRC_thickness`
- [ ] use `OuterBarrelMPGD_rmin` as rmin in MPGD @mposik1983

wdconinc:
This now has no more overlaps with either the MPGD or the backward tracking support cylinder.

Question for @rdom: The DIRC as specified sits centered between r = 70 cm and 73 cm. With a 50 mm high lens on the end of the bar, optical axis aligned with the center of the bar at 71.5 cm, this interferes with tracking support cones which are extending to just under 70 cm. This means that the lens has to shift outwards, and the center of the bar is not on the optical axis of the lens. There is a parameter (lens_shift) that I have set to 10 mm to avoid overlaps, but needs input from the DIRC group.

image

image

mposik1983 (dismissed):

wdconinc:
https://eicweb.phy.anl.gov/EIC/benchmarks/detector_benchmarks/-/pipelines/39661 fails in the materialScan benchmark for the sciglass due to a stack overflow in the recursive determination of DistFromInside in the polycone class at geom/geom/src/TGeoPcon.cxx.

wdconinc:
Removed material scan from requirements in detector_benchmarks.

mposik1983 (dismissed):

mposik1983 (approved):

PR #130, opened 2022-09-23
moved outer mpgd barrel to EPIC location

mposik1983:
Outer mpgd layer has been moved to radial location of 74.5 cm, as listed in table: https://eic.jlab.org/Geometry/Detector/Detector-20220905113308.html

This should avoid overlaps with the different calorimeter technologies being implemented. 

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.
Only location of mpgd outer barrel

### DIFF
--- a/compact/tracker.xml
+++ b/compact/tracker.xml
@@ -28,7 +28,7 @@
     <comment> barrel length: 2R/tan(angle) - dz </comment>
 
     <comment> Position in center of space behind DIRC </comment>
-    <constant name="OuterMPGDBarrel_rmin"            value="EcalBarrel_rmin - BarrelExtraSpace_thickness / 2."/>
+    <constant name="OuterMPGDBarrel_rmin"            value="74.5*cm"/>
     <comment> barrel asymmetryc between forward and backward lengths </comment>
 
     <comment> Service/Support setup </comment>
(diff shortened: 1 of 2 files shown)


compact/tracker.xml
```diff
@@ -28,7 +28,7 @@
     <comment> barrel length: 2R/tan(angle) - dz </comment>
 
     <comment> Position in center of space behind DIRC </comment>
-    <constant name="OuterMPGDBarrel_rmin"            value="EcalBarrel_rmin - BarrelExtraSpace_thickness / 2."/>
+    <constant name="OuterMPGDBarrel_rmin"            value="74.5*cm"/>
```
wdconinc, line 31: We should still use the main parametrization, not hardcoded numbers. I'd say we should rename `BarrelExtraSpace_thickness` to `OuterMPGDBarrel_thickness` of 2 cm in `definitions.xml`. Essentially, what's in the detector geometry table should be defined in `definitions.xml`. @sly2j for comment (though he's busy today).

wdconinc (changes requested):

compact/tracker.xml
wdconinc: Here: https://github.com/eic/epic/blob/barrel-mpgd-relocation/compact/definitions.xml#L468

mposik1983: I agree with keeping this parameterized with OuterMPGDBarrel_thickness. Do we want to keep this layer parameterized via EcalBarrel_rmin or should this eventually be changed to be parametrized relative to the DIRC position. Although I guess if both DIRC and this outer MPGD are parameterized via the Ecal than there is no issue keeping EcalBarrel_rmin.

veprbl: I think we should not subtract anything from `EcalBarrel_rmin` (defined as [1]). Instead, let's define `PIDBarrel_rmin` = `CentralTrackingRegion_rmax` and go from there.

[1] https://github.com/eic/epic/blob/ec158412e9f3958ce7c17ee9f147c9479b84633f/compact/definitions.xml#L530

sly2j: I agree, we're parametrizing from inside out, so mixing in outside-in parametrization will make things a bit harder to work with down the road.

PR #131, opened 2022-09-23
doc(dRICH): generation of dRICH and pfRICH material property tables 

### LINKED ISSUE #105: dRICH: add material property parameterizations
**Is your feature request related to a problem? Please describe.**
Material property tables for the dRICH are generally obtained from outside this repository in [g4dRIChOptics.hh](https://github.com/cisbani/dRICh/blob/main/share/source/g4dRIChOptics.hh), and copy-pasted into the XML files here in `epic`. It would be nice to have the parameterizations in `g4dRIChOptics.hh` stored here in `epic` somehow.

One idea is writing a script that (1) contains or is able to call these parameterizations, (2) generate the tables, and (3) can automatically update the XML files. Step (3) may not be the most useful, if we don't update this table often.

**Describe alternatives you've considered**
If anything, we should preserve where these numbers came from. More detailed comments in the XML file would be a good start.

**Additional context**
This header file can dump the tables; if I recall correctly one just needs to run a simulation in Fun4all following https://github.com/cisbani/dRICh documentation

c-dilks:
Document how to generate material property tables for the dRICH and pfRICH.

Close #105 

Alternatively, we could consider moving this generation code here to `epic`. Another possibility is removing these XML tables and use the material property table generation code directly as a plugin here, but I think for that we would need to be able to convert a `G4Material` to a `TGeoMaterial` (the reverse is done when converting `TGeo`->`Geant4`). Anyway, better documentation is a start in the right direction.

This also makes the change `dRICh` -> `dRICH` everywhere; despite attempts from some to call it "dRICh", it seems "dRICH" is much more popular...

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [x] Documentation update
- [ ] Other: __

No

No

### DIFF
--- a/compact/drich.xml
+++ b/compact/drich.xml
@@ -6,8 +6,8 @@
 <constant name="DRICH_Length"             value="ForwardRICHRegion_length"/>  <!-- overall vessel length -->
 <constant name="DRICH_zmin"               value="ForwardRICHRegion_zmin"/>
 <constant name="DRICH_zmax"               value="DRICH_zmin + DRICH_Length"/>
-<constant name="DRICH_rmin0"              value="DRICH_zmin * ForwardRICHRegion_tan1"/>  <!-- bore radius at dRICh entrance -->
-<constant name="DRICH_rmin1"              value="DRICH_zmax * ForwardRICHRegion_tan2"/>  <!-- bore radius at dRICh exit -->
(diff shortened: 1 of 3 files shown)


wdconinc (approved):

c-dilks:
Merging, since this PR only changed comments and is therefore not responsible for the failed benchmarks.

PR #133, opened 2022-09-25
Update the forward ECal to leave the space for ECal insert

jizhongling:
Share the same source code "src/EndcapCalorimeterWithInsert_geo.cpp" with HCal.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No. The users need to use the updated forward ECal XML file.

Yes, the default forward ECal XML file has been changed.


wdconinc:
@rymilton Based on my understanding, removing the previous forward ECal and replacing it with an ECal of type `epic_EndcapCalorimeterWithInsert` will just leave you with only the insert, right? Should we rename `epic_EndcapCalorimeterWithInsert` to make clear it just provides the insert only?

rymilton:
`epic_EndcapCalorimeterWithInsert` would give the ECal with a square cutout. `epic_InsertCalorimeter` gives you only the insert.

rymilton:
But you have a point here. The insert does not come for free with `epic_EndcapCalorimeterWithInsert`, so maybe it should be renamed to something like `epic_EndcapCalorimeterWithInsertCutout`

wdconinc:
And, the `epic_EndcapCalorimeterWithInsertCutout` is a layered HCal, not a homogen(e)ous tower or even scifi ECal, right? Just making sure we realize what we're doing here.

rymilton:
`epic_EndcapCalorimeterWithInsertCutout` is just a cylinder with layers and a rectangle hole at x = -10 cm. So what Zhongling has here is that shape but with layers of a cover, air, PCB, lightguides, and a homogeneous scintillator. But the number of layers is set to 1, so really it's just 1 layer of a cover, air, PCB, lightguides, and a homogeneous scintillator.

rymilton:
The scintillator slice is marked as sensitive and since it's homogeneous, it needs to be dealt with like how the previous homogeneous ecal was being handled

wdconinc:
I guess my underlying question is whether it makes sense in the long run to reuse this plugin written for a layered HCal for the endcap ECal, or whether is makes more sense to add cutout support to the HomogeneousCalorimeter plugin...

rymilton:
In the long run, it probably makes more sense to have a cutout option for `epic_ScFiCalorimeter`. Since ScFiCalorimeter_geo.cpp supports both fibers and homogeneous and `epic_EndcapCalorimeterWithInsertCutout ` does not, it's more future proof to just allow a cutout for the ecal insert.

rymilton:
A few things to note: 

1.  I wrote both the ECal and HCal code originally to be the same to minimize the number of files in our standalone simulations. I also wrote these with no intent of converting to fibers anytime soon.
2. `epic_EndcapCalorimeterWithInsertCutout` will likely become useless in the future anyways when Fredi/Nico implement their desired version of the LFHCAL. That class is just meant to be a temporary solution until they're done.
3. If not enough time, `epic_EndcapCalorimeterWithInsertCutout` can be used for the ECal, but it is less future proof

Issue #60, opened 2022-08-16
Outer MPGD layer overlaps with other detectors


veprbl:
### Environment: (where does this bug occur, have you tried other environments)
- main
- HEAD

### Steps to reproduce: (give a step by step account of how to trigger the bug)
Look at dump-constants job artifact (constants.toml), or
`npdet_info dump ${DETECTOR_PATH}/ecce.xml` should work

### Expected Result: (what do you expect when you execute the steps above)
According to https://eic.jlab.org/Geometry/Detector/Detector-20220624102507.html:
```
DIRC_rmax                      =       74.500
DIRC_rmin                      =       71.500 
OuterMPGDBarrel_rmin           =       75.500 (?)
OuterMPGDBarrelLayer_rmax      =       74.500 
OuterMPGDBarrelLayer_rmin      =       76.500 
EcalBarrel_rmin                =       79.500 
```

### Actual Result: (what do you get when you execute the steps above)
```
DIRC_rmax                      =       78.500 
DIRC_rmin                      =       71.500 
OuterMPGDBarrel_rmin           =       81.000
OuterMPGDBarrelLayer_rmax      =       82.000 
OuterMPGDBarrelLayer_rmin      =       80.000 
EcalBarrel_rmin                =       83.500 
```

veprbl:
Given https://github.com/eic/epic/blob/ef6fb8db6d08614e355a19e19940aba44396acd4/compact/definitions.xml#L529
it seems like `BarrelExtraSpace_thickness` was increased recently in 1385810f907ebe74a03be13e10debd500e125191, cc @sly2j

wdconinc:
The values in the detector menagerie should be considered with care. How is the radius of staves defined? Inscribed or circumscribed circle?

sly2j:
Feel free to set the correct values in ‘definitions.xml’ starting with the DIRC radius outward. Shouldn’t interfere with anything. I just made the tracker fit but nothing is validated against the menagerie right now.

veprbl:
Reducing the `BarrelExtraSpace_thickness` fudge constant causes an overlap with TOF:
```
 = Overlap ov00001: world_volume/InnerMPGDSubAssembly_7/InnerMPGDBarrel_0/layer1_0 overlapping world_volume/BarrelTOF_13/layer1_0 ovlp=1.1
```

sly2j:
There should not be a TOF right now, it’s not implemented right. If you rebase off main and regenerate XML the TOF should be disabled for now.

[...]

PR #139, opened 2022-09-27
remove Outer MPGD from EcalBarrel envelope

veprbl:

Fixes: #60
Closes: #130

- [x] Bug fix (issue #__)

- ~~Tests for the changes have been added~~
- ~~Documentation has been added / updated~~
- [x] Checked constants.toml artifact. New values are: `EcalBarrel_rmax                =      133.500` `EcalBarrel_rmin                =       79.500 `

No

No

### DIFF
--- a/compact/tracker.xml
+++ b/compact/tracker.xml
@@ -28,7 +28,7 @@
     <comment> barrel length: 2R/tan(angle) - dz </comment>
 
     <comment> Position in center of space behind DIRC </comment>
-    <constant name="OuterMPGDBarrel_rmin"            value="EcalBarrel_rmin - BarrelExtraSpace_thickness / 2."/>
+    <constant name="OuterMPGDBarrel_rmin"            value="CentralTrackingRegion_rmax"/>
     <comment> barrel asymmetryc between forward and backward lengths </comment>
 
     <comment> Service/Support setup </comment>
--- a/compact/definitions.xml
+++ b/compact/definitions.xml
@@ -531,6 +531,7 @@ Service gaps in FW direction (before endcapP ECAL) and BW direction (before endc
 
     <constant name="EcalBarrelRegion_thickness"     value="54.0*cm"/>
     <constant name="EcalBarrel_rmin"                value="CentralTrackingRegion_rmax + BarrelPIDRegion_thickness + BarrelExtraSpace_thickness"/>
+    <constant name="EcalBarrel_inner_margin"        value="1*cm"/>
     <constant name="EcalBarrel_rmax"                value="EcalBarrel_rmin + EcalBarrelRegion_thickness"/>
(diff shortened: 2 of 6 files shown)


mposik1983 (approved):

compact/tracker.xml
```diff
@@ -28,7 +28,7 @@
     <comment> barrel length: 2R/tan(angle) - dz </comment>
 
     <comment> Position in center of space behind DIRC </comment>
-    <constant name="OuterMPGDBarrel_rmin"            value="EcalBarrel_rmin - BarrelExtraSpace_thickness / 2."/>
+    <constant name="OuterMPGDBarrel_rmin"            value="CentralTrackingRegion_rmax"/>
```
mariakzurek, line 31: Do I understand correctly that this is a temporary fix to put the outer MPGD layer in the position of DIRC, so when the DIRC is properly implemented and put back the `OuterMPGDBarrel_rmin` will be shifted outwards after the DIRC?

wdconinc: Yeah, my initial question exactly, but I hadn't had a chance to really look into it. As the designated dirc implementer, I'd like to know.

veprbl: @mariakzurek You are absolutely right, this is temporary.

The main focus of this PR is to allow the current outer barrel MPGD geometry to exist at any radii (the current design fails the extrusion check, it could be that the current radius was fudged to avoid this). The fix needed is to set `rc` correctly, as it is already done for the inner MPGD and the inner tracker pieces. I also move the outer MPGD out of barrel region, so that it does not block the work on barrel calorimetry.

veprbl: This was done under assumption that DIRC is gone for this campaign. And I'm, of course, open to suggestions regarding nominal radius for MPGD.

mposik1983: I think for this simulation campaign, we should look to have at least a passive cylindrical layer that approximates the average DIRC material since this will affect the tracking and calorimeter performances.

wdconinc: We will have a working DIRC for this campaign: i.e. a DIRC that is dimensionally correct and has the correct material in the shape of DIRC bars.

mariakzurek: I think that for the purpose of this PR (removing MPGD from the Calorimeter, and fixing the calorimeter r position and parametrization of DIRC and MPGD positions), we can either leave the `OuterMPGDBarrel_rmin` at `CentralTrackingRegion_rmax` temporarily, as in the current geometry the DIRC is not implemented, and then move it after DIRC as @mposik1983 suggested. Or fix it to `CentralTrackingRegion_rmax+BarrelPIDRegion_thickness`, but the `BarrelPIDRegion_thickness` would need to be changed (from 8 cm to 3 cm), so that `OuterMPGDBarrel_rmin = 74.5 cm`.

PR #143, opened 2022-09-30
Modifications for ZDC replacement

shimasnd:

definition.xml and materials.xml are updated following the replacement of ZDC in eic/ip6.
(See pull request in eic/ip6 "replacement of the ZDC structure" https://github.com/eic/ip6/pull/26)

- [x] New feature (issue #__)

none

Will not work properly until the PR in eic/ip6 is merged.

### DIFF
--- a/compact/definitions.xml
+++ b/compact/definitions.xml
@@ -252,9 +252,12 @@ The unused IDs below are saved for future use.
     <constant name="ForwardOffMTracker_station_2_ID" value="160"/>
     <constant name="ForwardOffMTracker_station_3_ID" value="161"/>
     <constant name="ForwardOffMTracker_station_4_ID" value="162"/>
-    <constant name="ZDCEcal_ID"              value="163"/>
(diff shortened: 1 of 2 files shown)


ajentsch (approved):

ajentsch:
It seems there are some specific errors to address - I will be more careful before clicking the approve button next time.

wdconinc (approved):

wdconinc:
This looks like it will succeed now.

veprbl:
Can we force-merge this?

wdconinc:
> Can we force-merge this?

Technically we can, but the question is if we should. If we force merge this, the pipelines for other PRs are still going to fail because of the ZDC segmentation. I'm looking into fixing the underlying issue, then we can just merge this regularly by succeeding the pipelines normally.

wdconinc:
I'll force merge and then fix the benchmarks. There is too much backwards incompatible breakage downstream, and too many pipelines that require succeeding workflows before merges are possible.

Issue #94, opened 2022-09-01
Implementation of TTL layers


nschmidtALICE:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

wdconinc:
Can we flesh out this issue report a bit more? (it you wish)

What do we need here?

nschmidtALICE:
Hey, i will get to this when i'm a bit more familiar with what is already in DD4hep. But the main focus would be to get the barrel and forward layers close to the engineers designs regarding material budget.

PR #144, opened 2022-09-30
TOF endcap layer implementation

nschmidtALICE:
This PR introduces a new class for the endcap TOF layers. It contains the geometry as used in Fun4All and allows for further modifications once engineer inputs are available.

- [ ] Bug fix (issue #__)
- [x ] New feature (issue #94 )
- [ ] Documentation update
- [ ] Other: __

Since this i my first commit here, i cannot guarantee that everything will be perfect. However, geometry visualization and npsim event generation have been tested.


compact/display.xml
```diff
@@ -36,6 +36,11 @@
     <vis name="TOFHybridVis"      ref="AnlGold"             visible="true"  showDaughters="true"/>
     <vis name="TOFCoolingTube"    ref="AnlRed"              visible="true"  showDaughters="true"/>
     <vis name="TOFCoolant"        ref="TrackerServiceVis"   visible="true"  showDaughters="true"/>
+    <vis name="TOFWater"          ref="AnlBlue"   visible="true"  showDaughters="true"/>
+    <vis name="TOFAluminum"        ref="AnlLightGray"   visible="true"  showDaughters="true"/>
+    <vis name="TOFActiveMat"        ref="AnlGreen"   visible="true"  showDaughters="true"/>
+    <vis name="TOFLayers"        ref="AnlBlue"   visible="true"  showDaughters="true"/>
+    <vis name="TOFSensorAndReadoutLadder"        ref="AnlGold"   visible="true"  showDaughters="true"/>
```
wdconinc, line 43: Maintain alignment for readability.

compact/tof_endcap.xml
```diff
@@ -62,18 +72,23 @@
     <constant name="BackwardTOFLayer_rmax"      value="BackwardTOFMod_rmax + 1*um" />
     <constant name="BackwardTOFLayer1_zmin"     value="BackwardTOFMod1_zmin - TOFEndcapLayer_thickness / 2" />
     <constant name="BackwardTOFLayer2_zmin"     value="BackwardTOFMod2_zmin - TOFEndcapLayer_thickness / 2" />
-  </define>
+
+    <constant name="BackwardTOF_rMin"           value="6*cm" />
+    <constant name="BackwardTOF_rMax"           value="64*cm" />
+    <constant name="BackwardTOF_xOffset"        value="0.0*cm" />
+    <constant name="BackwardTOF_det_height"     value="2.0*cm" />  </define>
```
wdconinc, line 79:
```suggestion
    <constant name="BackwardTOF_det_height"     value="2.0*cm" />
  </define>
```

compact/tof_endcap.xml
```diff
@@ -95,12 +110,13 @@
...
         id="BackwardTOF_ID"
         name="BackwardTOF"
-        type="epic_TrapEndcapTracker"
+        type="epic_TOFEndcap"
         readout="TOFEndcapHits"
         vis="TOFVis"
         reflect="true">
         <module name="BackwardModule" vis="TOFModuleVis">
           <trd x1="BackwardTOFMod_x1/2.0" x2="BackwardTOFMod_x2/2.0" z="BackwardTOFMod_y/2"/>
+          <diskdimension rMin="BackwardTOF_rMin" rMax="BackwardTOF_rMax" xOffset="BackwardTOF_xOffset" det_height="BackwardTOF_det_height" zPos="BackwardTOF_zmin" cooling_plate_height="TOFEndcap_cooling_plate_height" wallthickness_coolingtube="TOFEndcap_wallthickness_coolingtube" diameter_coolingtube="TOFEndcap_diameter_coolingtube" material="Aluminum"/>
```
wdconinc, line 119: Do we need to call this `diskdimension` instead of `dimension`? Most dd4hep geometry plugins use simply `dimension` so I'd prefer to stay with that approach.

src/EndcapTOF_geo.cpp
```diff
@@ -0,0 +1,721 @@
...
+
+#if defined(USE_ACTSDD4HEP)
+#include "ActsDD4hep/ActsExtension.hpp"
+#include "ActsDD4hep/ConvertMaterial.hpp"
+#else
+#include "Acts/Plugins/DD4hep/ActsExtension.hpp"
+#include "Acts/Plugins/DD4hep/ConvertDD4hepMaterial.hpp"
+#endif
+
+using namespace std;
```
wdconinc, line 27: Avoid `using namespace std` and include explicit `using std::vector` or such.

src/EndcapTOF_geo.cpp
```diff
@@ -0,0 +1,721 @@
...
+        0.25 * mm, 0.79 * mm, 0.08 * mm, 0.25 * mm, 0.03 * mm, 0.3 * mm, 0.08 * mm, 0.51 * mm};
+    double widthLayer[nLayers] = {
+        baseplate_width - 0.3 * mm,
+        baseplate_width,
+        sensor_width + 1 * mm,
+        sensor_width + 1 * mm,
+        sensor_width - 0.2 * mm,
+        sensor_width,
+        sensor_width,
+        baseplate_width - 4 * mm};
```
wdconinc, line 155: There is still a lot of parametrization going on inside the cpp plugin. This would better be moved to the xml file with a `layer` support (as other detectors are doing). That also guarantees consistency since each layer would be required to have a thickenss, material, and name.

wdconinc (changes requested):
Some changes requested, but some of the other comments may be suitably addressed by filing an issues and deferring. I just don't want them to get lost either, and for a future inclusion of e.g. cooling pipes, it would be good to update the plugin for code quality.

Only parameters that are in xml will be able to be modified through conditions or geometry databases to the 'as-built' values that differ from detector sector to sector.

src/EndcapTOF_geo.cpp
nschmidtALICE: Moved these definitions to the xml file.

src/EndcapTOF_geo.cpp
nschmidtALICE: Done.

compact/tof_endcap.xml
nschmidtALICE: Done.

wdconinc (approved):

src/EndcapTOF_geo.cpp
```diff
@@ -236,9 +236,10 @@ static Ref_t create_detector(Detector& description, xml_h e, SensitiveDetector s
         materialLayer_SH[i_slice]  = description.material(x_slice.materialStr());
         strLayerName_SH[i_slice]   = getAttrOrDefault<std::string>(x_slice, _U(name), "slice" + std::to_string(i_slice));
         thicknessLayer_SH[i_slice] = x_slice.thickness();
-        widthLayer_SH[i_slice] = getAttrOrDefault(sensparams, _Unicode(width), 0.);
-        offsetLayer_SH[i_slice] = getAttrOrDefault(sensparams, _Unicode(offset), 0.);
-        lengthLayer_SH[i_slice] = getAttrOrDefault(sensparams, _Unicode(length), 0.);
+        widthLayer_SH[i_slice] = getAttrOrDefault(x_slice, _Unicode(width), 0.);
+        offsetLayer_SH[i_slice] = getAttrOrDefault(x_slice, _Unicode(offset), 0.);
+        lengthLayer_SH[i_slice] = getAttrOrDefault(x_slice, _Unicode(length), 0.);
+        // std::cout << "SH: " << strLayerName_SH[i_slice] << "\tthick: " << thicknessLayer_SH[i_slice] << "\twidth: " << widthLayer_SH[i_slice] << "\toffset: " << offsetLayer_SH[i_slice] << "\tlength: " << lengthLayer_SH[i_slice] << std::endl;
```
wdconinc, line 242:
```suggestion
```

wdconinc (approved):

compact/definitions.xml
```diff
@@ -445,8 +445,11 @@ Examples:
     <comment> Forward TOF region currenlty empty for future upgrades </comment>
     <constant name="ForwardTOFRegion_zmin"        value="ForwardPIDRegion_zmin" />
     <constant name="ForwardTOFRegion_length"      value="22.0*cm" />
-    <constant name="ForwardTOFRegion_rmax"        value="CentralTrackingRegion_rmax" />
     <constant name="ForwardTOFRegion_tan"         value="CentralTrackingRegionP_tan" />
+    <constant name="ForwardTOFRegion_minR"            value="7*cm" />
+    <constant name="ForwardTOFRegion_maxR"            value="87*cm" />
```
wdconinc, line 450: https://eic.jlab.org/Geometry/Detector/Detector-20220929172703.html has 8 cm to 67 cm.

nschmidtALICE: Done in new commit

wdconinc (approved):

PR #148, opened 2022-10-03
Updates of mRICH positioning to avoid overlaps

wdconinc:
This PR updates the module parameters for the mRICH and packs as many as possible in the envelope volume.

- [ ] Bug fix (issue #__)
- [X] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

Co-authored-by: Wouter Deconinck <wdconinc@gmail.com>


wdconinc:
As many smaller modules as reasonable...

Before (`main` branch, with larger modules):
image

Before (`mrich` branch, with smaller modules):
image

After (`mrich` branch, with closest packing):
image

compact/mrich.xml
```diff
@@ -4,17 +4,18 @@
     <constant name="MRICH_rmin"   value="10*cm"/>
     <constant name="MRICH_rmax"   value="BackwardPIDRegion_rmax"/>
     <constant name="MRICH_length" value="BackwardPIDRegion_length"/>
-    <constant name="MRICH_zmin"   value="BackwardPIDRegion_zmin"/>
+    <constant name="MRICH_zmax"   value="BackwardPIDRegion_zmax"/>
+    <constant name="MRICH_zmin"   value="MRICH_zmax-MRICH_length"/>
```
wdconinc, line 8: @c-dilks This is how I'm pushing the MRICH up against zmax. Wasn't as cumbersome as I had feared (and definitely avoided having to mess with jinja).

compact/mrich.xml
```diff
@@ -24,7 +25,8 @@
     <constant name="MRICHRearExtraSpace_thickness"    value="0.3*mm"/>
     <constant name="MRICHLensPhotoDet_length"         value="136.4*mm"/>
     <constant name="MRICHMirror_thickness"            value="2.0*mm"/>
-    <constant name="MRICHMirror_length"               value="MRICHLensPhotoDet_length - MRICHLensMirrorGap_thickness"/>
+    <!-- <constant name="MRICHMirror_length"               value="MRICHLensPhotoDet_length - MRICHLensMirrorGap_thickness"/>-->
```
wdconinc, line 28: Remove commented line.
```suggestion
```

compact/mrich.xml
c-dilks: Makes sense, I will update #158 to be consistent. Btw, I get `172.470` for `BackwardPIDRegion_zmax` here, but the menagerie says `172.7`. It's not a big enough difference to matter, at least for the pfRICH, so I'll leave it up to you if you want to update it here.

c-dilks: Just need to change `CentralTrackingRegionN_zmax`

osbornjd (commented):
Looks good to me from an integration standpoint, everything seems to be within the menagerie as far as I can tell. I defer to @c-dilks for full approval as the RICH expert.

compact/mrich.xml
wdconinc: I'm deferring to the `BackwardPIDRegion_zmax` as provided by the parametrization, so I'm not going to hard code it here.

c-dilks:
Looks good to me too. Once this PR is merged I'm going to change `BackwardPIDRegion_rmax` from 64->63 cm in #158 and rebase, so both the pfRICH and mRICH are consistent with the menagerie. The only side effect is on the endcap TOF radius, so that's fine.

c-dilks (dismissed):

c-dilks:
> I'm going to change BackwardPIDRegion_rmax from 64->63 cm
Hmm... after looking at the mRICH in `jsroot` I probably shouldn't....

c-dilks:
In that case, why don't we just change `BackwardPIDRegion_rmax` from 64 to 63 in `definitions.xml`? It's _only_ used by `mrich.xml`, `pfrich.xml`, and `tof_endcap.xml`, and no other parameters in `definitions.xml` depend on it.

wdconinc:
Yes, that's the goal but I wanted to keep this focused on the mRICH first.

wdconinc:
image

wdconinc:

c-dilks (approved):

PR #153, opened 2022-10-03
Forward hcal 130cm

rymilton:
Extends length of forward hcal to 130 cm based on [09/29/2022 documentation](https://eic.jlab.org/Geometry/Detector/Detector-20220929172703.html). Extended the hcal insert as close to 130 cm, without going over. 

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [x] Other: Updated numbers based on 09/29/2022 numbers.

None
Yes, changes the default length of the forward hcal + hcal insert.


rymilton:
Not sure why all my previous commits are maintained here, but this pull request is only one simple commit. It just updates the length of the forward hcal system to be 130 cm (latest numbers from the online documentation).

wdconinc (dismissed):

rymilton:
@wdconinc Is there a way to run the required checks so this PR can satisfy the auto-merge requirements?

wdconinc:
> Is there a way to run the required checks so this PR can satisfy the auto-merge requirements?

Yes :-) #163 was intended to address this and the fact that the checks were restarted here is a demonstrated that it is working :-)

compact/definitions.xml
```diff
@@ -567,11 +567,11 @@ Service gaps in FW direction (before endcapP ECAL) and BW direction (before endc
 
     <constant name="HcalBarrelStavesN"          value="12"/>
     <constant name="HcalEndcapP_zmin"     value="EcalEndcapP_zmin + EcalEndcapP_length"/>
-    <constant name="HcalEndcapP_length"   value="120.0*cm"/>
+    <constant name="HcalEndcapP_length"   value="130.0*cm"/>
```
wdconinc, line 570: Why 130 cm when the https://eic.jlab.org/Geometry/Detector/Detector-20220929172703.html says 140 cm? Are we sticking to that detector parameter table or not?

rymilton: I brought this up during the calorimetry meeting. The detector is 130 cm + 10 cm of electronics. I have not included electronics here and was told this was fine

rymilton: I can change the length in `definitions.xml` to equal the _actual_ length of 140 cm, and then make the hcal only actually 130 cm, perhaps. This would solve other definitions relying on `HcalEndcapP_length` in `definitions.xml`

rymilton: But essentially since this isn't actually my detector and I'm just helping out ORNL, I have no idea what the electronics system is like nor what I should implement for it

wdconinc: Yeah, I think that would be better. Thanks for the clarification!

rymilton: Alright, I'll adjust that it in a bit!

wdconinc (approved):

PR #126, opened 2022-09-22
feat: introduce Arches and BryceCanyon configurations

wdconinc:
### Briefly, what does this PR introduce?
For the October 2022 campaign we are targetting two major geometry configuration. This introduces these configurations with their components.

The graphic below is a starting point, not a target of the features to be included:
image

### What kind of change does this PR introduce?
- [ ] Bug fix (issue #__)
- [X] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [ ] Changes have been communicated to collaborators

### Does this PR introduce breaking changes? What changes might users need to make to their code?
Yes. `epic_imaging.yml` and `epic_sciglass.yml` have been removed.

### Does this PR change default behavior?
Yes. `epic.yml` will by default pick `epic_arches.yml`

PR #158, opened 2022-10-04
feat(PID): rescale pfRICH and sync fixes from the dRICH

c-dilks:
- synchronize developments and bug fixes from the dRICH, applying them to the pfRICH
- re-scale the pfRICH to the requested envelope

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

Not really, but choosing the mRICH vs. pfRICH config will be better addressed by #126


c-dilks:
At this point, this PR passes all CI checks, except for overlap checks (see #159) (not waiting for benchmarks to run). I will now remove the configuration YAML file changes, which will revert the default backward RICH to be the mRICH, and make this PR cleaner.

compact/definitions.xml
```diff
@@ -424,7 +424,7 @@ Examples:
 
     <constant name="CentralTrackingRegion_rmax"   value="715.0*mm" />
     <constant name="CentralTrackingRegionP_zmax"  value="1730.0*mm" />
-    <constant name="CentralTrackingRegionN_zmax"  value="1474.7*mm" />
+    <constant name="CentralTrackingRegionN_zmax"  value="1186.0*mm" />
```
c-dilks, line 427: Help needed here, to make the z-positioning consistent with the mRICH (outside->in, rather than inside->out)

compact/drich.xml
```diff
@@ -110,7 +110,7 @@
 <radiator
   rmin="DRICH_rmin0 + DRICH_wall_thickness + 0.2*cm"
   rmax="DRICH_rmax0 - DRICH_wall_thickness - 0.2*cm"
-  frontplane="DRICH_window_thickness + 0.5*DRICH_aerogel_thickness"
+  frontplane="DRICH_window_thickness"
```
c-dilks, line 113: style fix

src/DRICH_geo.cpp
```diff
@@ -268,7 +268,7 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
   // aerogel placement and surface properties
   // TODO [low-priority]: define skin properties for aerogel and filter
   // FIXME: radiatorPitch might not be working correctly (not yet used)
-  auto radiatorPos      = Position(0., 0., radiatorFrontplane) + originFront;
+  auto radiatorPos      = Position(0., 0., radiatorFrontplane + 0.5 * aerogelThickness) + originFront;
```
c-dilks, line 271: style fix, geometry is the same

c-dilks:
Checked with IRT reconstruction. Looks good enough!
_theta_scan_pfx agl

compact/pid_config.xml
```diff
@@ -1,5 +1,7 @@
 <lccdd>
 
+<!-- FIXME: this file not used anywhere... delete it? -->
+
```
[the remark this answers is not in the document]
wdconinc: I think the file can be deleted.

compact/pfrich.xml
```diff
@@ -151,26 +151,24 @@
...
   </comment>
-  <component name="aluminum" thickness="5*mm" vis="RICHServiceVis" material="Aluminum"/>
-  <component name="air" thickness="40*mm" vis="InvisibleNoDaughters" material="Air"/>
-  <component name="aluminum" thickness="5*mm" vis="RICHServiceVis" material="Aluminum"/>
-  <component name="air" thickness="40*mm" vis="InvisibleNoDaughters" material="Air"/>
-  <component name="aluminum" thickness="5*mm" vis="RICHServiceVis" material="Aluminum"/>
+  <component name="aluminum" thickness="0.5*cm" vis="RICHServiceVis" material="Aluminum"/>
+  <component name="air" thickness="(PFRICH_services_length - 3*0.5*cm) / 2.0" vis="InvisibleNoDaughters" material="Air"/>
+  <component name="aluminum" thickness="0.5*cm" vis="RICHServiceVis" material="Aluminum"/>
+  <component name="air" thickness="(PFRICH_services_length - 3*0.5*cm) / 2.0" vis="InvisibleNoDaughters" material="Air"/>
```
wdconinc, line 170: Is this correct? I assume 3*0.5*cm is to account for the three aluminum layers. But isn't 0.5*cm the half-length, so shouldn't this be `(PFRICH_services_length - 3*1.0*cm) / 2.0`?

compact/definitions.xml
```diff
@@ -456,13 +456,14 @@ Examples:
     <constant name="ForwardRICHRegion_tan2"        value="Eta3_6_tan * 0.89" />
 
     <constant name="BackwardPIDRegion_zmin"       value="CentralTrackingRegionN_zmax" />
-    <constant name="BackwardPIDRegion_length"     value="25.0*cm" />
+    <constant name="BackwardPIDRegion_length"     value="54.1*cm" />
```
wdconinc, line 459: I'm going to have to figure out how to make this work with the mRICH. Could be either here (if you have time) or in a separate PR. I couldn't immediately see the `position` element for the detector, but if we position it 'from the outside in' then we probably want position z to be `BackwardPIDRegion_zmax - pfRICH_length / 2.0` or so.

compact/pfrich.xml
c-dilks: 0.5 cm is the full length; thicknesses are divided by 2 in `PFRICH_geo.cpp`

c-dilks: Checking the numbers from the `placeVolume` calls, looks okay:
```
TEST                   service   Aluminum, thickness=0.500000, zpos=-7.250000
TEST                   service        Air, thickness=6.750000, zpos=-3.625000
TEST                   service   Aluminum, thickness=0.500000, zpos=0.000000
TEST                   service        Air, thickness=6.750000, zpos=3.625000
TEST                   service   Aluminum, thickness=0.500000, zpos=7.250000
TEST                   thickness_sum=7.500000, total_thickness=15.000000, global_zpos=-18.600000
```

compact/definitions.xml
c-dilks: I'm not sure how to do this either, need a better understanding of the differences between the two full configurations. It looks like the positions are:
```xml
<constant name="MRICH_zmin"   value="BackwardPIDRegion_zmin"/>
<constant name="PFRICH_zmin"  value="-BackwardRICHRegion_zmin"/>
```

c-dilks: Maybe it's time for another `jinja2` template :thinking:

wdconinc: Noooooo!

compact/pid_config.xml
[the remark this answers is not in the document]
c-dilks: deleted

compact/definitions.xml
c-dilks: resolved after https://github.com/eic/epic/pull/148

wdconinc:
@c-dilks @osbornjd Review is the only thing needed here, right?

c-dilks:
We might be in trouble with the differing `BackwardPIDRegion_length` for the mRICH and pfRICH. Let's merge #148 first and then figure out what to do about it here.

compact/definitions.xml
```diff
@@ -461,18 +460,23 @@ Examples:
     <constant name="ForwardRICHRegion_tan1"        value="CentralTrackingRegionP_tan * 0.88" />
     <constant name="ForwardRICHRegion_tan2"        value="Eta3_6_tan * 0.89" />
 
-    <constant name="BackwardPIDRegion_length"     value="25.0*cm" />
-    <constant name="BackwardPIDRegion_zmin"       value="CentralTrackingRegionN_zmax" />
-    <constant name="BackwardPIDRegion_zmax"       value="BackwardPIDRegion_zmin+BackwardPIDRegion_length" />
-    <constant name="BackwardPIDRegion_rmax"       value="64.0*cm" /> <!-- FIXME hardcoded, was CentralTrackingRegion_rmax -->
+    <constant name="MRICH_length"                 value="25.0*cm" />
+    <constant name="PFRICH_length"                value="54.1*cm" />
+    <constant name="BackwardPIDRegion_length"     value="max(MRICH_length,PFRICH_length)" /> <!-- FIXME: should this depend on brycecanyon vs. arches -->
```
c-dilks, line 465: This is something that should depend on `arches` vs. `brycecanyon`, I'm not sure what to do about it, so set it to `max` of the two RICH lengths for now.

c-dilks:
@wdconinc I tried to parameterize from `BackwardPIDRegion_zmax` inward to the central region. Some things to note:
- moved some `CentralTrackingRegion` constants to after the backward PID section
- the differing RICH lengths could be handled better, since they now impact the central region length

wdconinc:
I thought we decided on a fixed backward pid length, so there is no need for different tracking zmax. We don't need to define the mRICH and pfRICH length in definition, just the backward pid region length. If the mRICH stays inside that region, it is fine.

osbornjd:
> I thought we decided on a fixed backward pid length, so there is no need for different tracking zmax. We don't need to define the mRICH and pfRICH length in definition, just the backward pid region length. If the mRICH stays inside that region, it is fine.

This was also my impression from the discussion at the various GDI meetings.

c-dilks:
Ah that makes more sense. I must have missed that decision and assumed we still had to re-parameterize things from the outside->in.

compact/definitions.xml
c-dilks: No longer relevant

c-dilks:
So `BackwardPIDRegion_length` should be that of the pfRICH? If so, to get `BackwardPIDRegion_z{min,max}` correct, either `CentralTrackingRegionN_zmax` needs to be reduced, or we need to switch which of these 3 is the hard-coded number, or we add an offset to either `BackwardPIDRegion_z{min,max}`. Currently I have reduced `CentralTrackingRegionN_zmax`.

I'm confused and likely missing a simple solution, so feel free to push commits to this branch if you have a better idea.

Also, with the changed `BackwardPIDRegion_length`, the mRICH modules move in the +z direction.

wdconinc:
My understanding/suggestion is this:
- `definitions.xml` should provide the single `CentralTrackingRegionN_zmax` since there is only a single tracking configuration that ends at -118.6 cm; this should be `BackwardPIDRegion_zmin`.
- `definitions.xml` should provide the single `BackwardPIDRegion_zmax` (and `_length`) that ends at -172.7 cm.
- `pfrich.xml` can define `PFRICH_zmin` and `PFRICH_zmax` however it wants based on `BackwardPIDRegion_zmin` and `BackwardPIDRegion_zmax`, e.g. `PFRICH_zmin = BackwardPIDRegion_zmin` and `PFRICH_zmax = PFRICH_zmin + PFRICH_length` based on the actual `PFRICH_length`.
- `mrich.xml` can define `MRICH_zmin` and `MRICH_zmax` however it wants based on `BackwardPIDRegion_zmin` and `BackwardPIDRegion_zmax`, e.g. `MRICH_zmax = BackwardPIDRegion_zmax` and `MRICH_zmin = MRICH_zmax - MRICH_length` based on the actual `MRICH_length`.

I think that is what you have here now, isn't it?

Of course that introduces overlaps with the tracker support, because the detector parameter table does not leave any space for them.

wdconinc:
Maybe we should change (`tracker.xml`)
```xml
    <constant name="BackwardEndcapSupportCyl_zmax"   value="300*cm" />
```
to end at `BackwardPIDRegion_zmin`.

c-dilks:
Things are looking better now, but not yet ready for review.
@wdconinc, `MRICH_length` controls the placement of the modules. On `main` it is 25 cm (1st figure); the latest version of this feature branch has it at the menagerie value of 42.7 cm (2nd figure), shifting the modules forward. Do you prefer to have it set to 25 cm, or to have it at 42.7 cm and readjust the positioning?

2022-10-17-145449_240x597_scrot
2022-10-17-145608_300x594_scrot

c-dilks:
> Maybe we should change (`tracker.xml`)
> 
> ```
>     <constant name="BackwardEndcapSupportCyl_zmax"   value="300*cm" />
> ```
> 
> to end at `BackwardPIDRegion_zmin`.

I got some bounding box error from doing this. I just pushed a commit to test disabling `BackwardEndcapSupportCyl` altogether. Here is a before and after `dawn` view:

view1_top

(AFTER view will be posted after workflow run completes)

c-dilks:
Now testing overlap checks on both `arches` and `brycecanyon`.

In a prior workflow run, there are overlaps in `overlap-check-geant4-fast`, but not `overlap-check-tgeo`, between mRICH modules:
```
-------- WWWW ------- G4Exception-START -------- WWWW -------
*** G4Exception : GeomVol1002
      issued by : G4PVPlacement::CheckOverlaps()
Overlap with volume already placed !
          Overlap is detected for volume MRICH_module1_5:5 (G4Box) with MRICH_module1_1:1 (G4Box)
          overlap at local point (47.6352,-51.8861,-96.1951) by 113.936 um  (max of 2 cases)
NOTE: Reached maximum fixed number -1- of overlaps reports for this volume !
-------- WWWW -------- G4Exception-END --------- WWWW -------
```
etc.

wdconinc:
> between mRICH modules

mRICH is (regrettably) sensitive to being moved around due to the projective nature. If it is now closer to the IP, then the modules will be pointing more sharply inwards, and they will overlap.

c-dilks:
Both `arches` and `brycecanyon` have overlaps between `DIRC/DIRCModule` and `EcalBarrel`.

c-dilks:
Note that I've cherry picked #203 and #222 for validation purposes. I will clean up this PR once things are in better shape and tested. Also keeping an eye on #223 since this will do a better job than just disabling `BackwardEndcapSupportCyl` as tested here, and has similar changes in `definitions.xml`.

c-dilks:
@wdconinc should I revert `.github/workflows/linux-eic-shell.yml` after the checks pass? It enables `tgeo` and `geant4-fast` overlap checks for `arches`+`brycecanyon` (pulled from your PR)

wdconinc:
@c-dilks We can keep the arches and brycecanyon in there (but we should add the barrel imaging ecal to brycecanyon, since they both run with sciglass).

wdconinc (approved):
Looks good to go now, finally :-)

c-dilks:
Seconded. I double checked the constants for both configs, and their diffs; everything looks good!

c-dilks:
DIRC / ECalBarrel overlaps :man_facepalming:

wdconinc:
That needs to be solved in the ECal length pr.

wdconinc:

wdconinc:
eicweb benchmarks for brycecanyon resubmitted with fix on eicweb.

PR #163, opened 2022-10-05
ci: use a dedicated PAT for auto-update resubmit

wdconinc:
Using a standard GitHub workflow token for the pull request auto-updates does not allow resubmitting the workflows after the updates. This is annoying since it prevents auto-merges from actually merging. This replaces the use of a standard token with a personal access token that has these permissions.

- [X] Bug fix (issue: workflows do no restart when pull requests are auto-updated)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/workflows/auto-update.yml
+++ b/.github/workflows/auto-update.yml
@@ -12,4 +12,4 @@ jobs:
     steps:
       - uses: tibdex/auto-update@v2
         with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


wdconinc:
@veprbl As far as I can tell, this is the simplest approach (if slightly insecure for my account) of getting workflows to be resubmitted. As usual, it's hard to test these CI changes that are only active on `main`...

wdconinc:
The relevant secret was added to the settings of the repository so it will be used by pipelines run on branches from inside the repository.

veprbl (approved):

PR #173, opened 2022-10-06
fix(dRICH): slightly increase envelope radii to be consistent with menagerie

c-dilks:


wdconinc (approved):

c-dilks:
Hang on.. not sure I should have increased the aerogel radius...

wdconinc (approved):

c-dilks:
Rebased

PR #183, opened 2022-10-07
feat: run geometry checks

wdconinc:
DD4hep has the checkGeometry tool which can be used to produce histograms of time per track spent per volume. This workflow produces these histograms and log output for both fast and full geometry.

- [ ] Bug fix (issue #__)
- [X] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -233,33 +233,6 @@ jobs:
           noverlaps="$(grep -c ovlp doc/overlap_check_tgeo.out || true)"
           if [[ "${noverlaps}" -gt "0" ]] ; then echo "${noverlaps} overlaps found!" && false ; fi
 
-  check-geometry-fast:
-    runs-on: ubuntu-latest
-    needs: build-test
-    if: github.ref == 'refs/heads/main'
-    steps:
-    - uses: actions/checkout@v3
-    - uses: actions/download-artifact@v3
-      with:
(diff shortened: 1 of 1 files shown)


.github/workflows/linux-eic-shell.yml
```diff
@@ -233,6 +233,60 @@ jobs:
...
+        name: build-fast-eic-shell
+        path: install/
+    - uses: cvmfs-contrib/github-action-cvmfs@v3
+    - uses: eic/run-cvmfs-osg-eic-shell@main
+      with:
+        platform-release: "jug_xl:nightly"
+        setup: install/setup.sh
+        run: |
+          mkdir -p doc
+          checkGeometry -f true -c ${DETECTOR_PATH}/${DETECTOR_CONFIG}.xml | tee geometry_check_fast.out
```
veprbl, line 253: If this is "fast", then it should be `-f false` here?

wdconinc: Yeah, probably. Originally the goal was to have this work on the fast geometry, no fibers and such, but it doesn't run faster than the full geometry. So yeah, we'd probably switch to false.

veprbl: Or remove, if it's currently identical.

wdconinc: Removed.

veprbl (approved):

Issue #154, opened 2022-10-04
Add TOF into tracker.xml


yezhenyu2003:
create a TrackerSubAssembly in tracker.xml for TOF

PR #191, opened 2022-10-10
Include new Forward TOF in tracker assembly

yezhenyu2003:
test


yezhenyu2003:
Fails on tracker geometry check when including new forward TOF model in the tracker assembly
(#154 uses simplified forward TOF model in tracker assembly and works fine). 
Difference between #154 and #191 are tracker.xml and tof_endcap.xml

wdconinc:
@yezhenyu2003 @nschmidtALICE I'm assuming this can be closed now that https://github.com/eic/epic/pull/466 has been merged. Please reopen if we still need this.

PR #196, opened 2022-10-11
76 outer mpgd layer geometry change 1

mposik1983:
Change outer MPGD layer to bar geometry and build using assembly.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

Yes. Outer MPGD layer is not implemented into ACTS

Yes. Outer MPGD layer is not implemented into ACTS

### DIFF
--- a/compact/tracker.xml
+++ b/compact/tracker.xml
@@ -28,7 +28,7 @@
     <comment> barrel length: 2R/tan(angle) - dz </comment>
 
     <comment> Position in center of space behind DIRC </comment>
-    <constant name="OuterMPGDBarrel_rmin"            value="DIRC_rmin + DIRC_thickness"/>
+    <constant name="MPGDDIRC_rmin"            value="DIRC_rmin + DIRC_thickness"/>
     <comment> barrel asymmetryc between forward and backward lengths </comment>
 
     <comment> Service/Support setup </comment>
--- a/compact/tracker.xml
+++ b/compact/tracker.xml
@@ -133,7 +133,7 @@
   </documentation>
   <plugins>
     <plugin name="epic_FileLoader">
-      <arg value="cache:$DETECTOR_PATH"/>
+      <arg value="cache:$DETECTOR_PATH:/opt/detector"/>
       <arg value="file:calibrations/materials-map.cbor"/>
       <arg value="url:https://eicweb.phy.anl.gov/EIC/detectors/athena/uploads/56b64d544442a93904488e7292aa509d/material-maps.cbor"/>
     </plugin>
--- a/compact/tracker.xml
+++ b/compact/tracker.xml
@@ -139,11 +139,11 @@
     </plugin>
   </plugins>
 
-	     <include ref="tracker/vertex_barrel.xml"/>
-	     <include ref="tracker/silicon_barrel.xml"/>
-	     <include ref="tracker/mpgd_barrel.xml"/>
-	     <include ref="tracker/mpgd_dirc.xml"/>
-	     <include ref="tracker/silicon_disks.xml"/>
-	     <include ref="tracker/support_service_assembly.xml"/>
+  <include ref="tracker/vertex_barrel.xml"/>
+  <include ref="tracker/silicon_barrel.xml"/>
+  <include ref="tracker/mpgd_barrel.xml"/>
+  <include ref="tracker/mpgd_dirc.xml"/>
+  <include ref="tracker/silicon_disks.xml"/>
+  <include ref="tracker/support_service_assembly.xml"/>
 
 </lccdd>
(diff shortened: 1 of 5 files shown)


compact/tracker.xml
```diff
@@ -133,16 +133,17 @@
   </documentation>
   <plugins>
     <plugin name="epic_FileLoader">
-      <arg value="cache:$DETECTOR_PATH:/opt/detector"/>
+      <arg value="cache:$DETECTOR_PATH"/>
```
wdconinc, line 136: Was there a reason for removing this? It is required to find the material maps in the container.

compact/tracker.xml
```diff
@@ -133,16 +133,17 @@
...
-  <include ref="tracker/silicon_barrel.xml"/>
-  <include ref="tracker/mpgd_barrel.xml"/>
-  <include ref="tracker/silicon_disks.xml"/>
-  <include ref="tracker/support_service_assembly.xml"/>
+	     <include ref="tracker/vertex_barrel.xml"/>
+	     <include ref="tracker/silicon_barrel.xml"/>
+	     <include ref="tracker/mpgd_barrel.xml"/>
+	     <include ref="tracker/mpgd_dirc.xml"/>
+	     <include ref="tracker/silicon_disks.xml"/>
+	     <include ref="tracker/support_service_assembly.xml"/>
```
wdconinc, line 147: No tabs please :-)

compact/tracker/mpgd_dirc.xml
```diff
@@ -0,0 +1,185 @@
...
+  <limits>
+  </limits>
+
+  <regions>
+  </regions>
+
+  <display>
+	  </display>
+
+	  <detectors>
```
wdconinc, line 63: More tabs to be removed. Probably good to apply that generally.

src/MPGDDIRC_geo.cpp
```diff
@@ -0,0 +1,243 @@
...
+      //  ------------------- MPGD foil 
+      //  --               -- Frame
+      //  --  gas volume   -- Frame
+      //  --               -- Frame
+      //  ------------------- MPGD foil
+
+      //Look for gas modules to subtract frame thickness from
+      //FIXME: these module names are hard coded for now. Should find 
+      //a way to set a arribut via the moduel tag to flag what components 
+      //need to have frame thickness subtracted.
```
wdconinc, line 118: Is this why this is implemented as a different geometry plugin? Mainly I'm hoping we can avoid code duplication between this and the tracker geometry plugin.

compact/tracker.xml
mposik1983: Not sure why that is changed. I did not explicitly change it, oh maybe it was from sourcing my local build directory? Anyhow I think keeping it as is ($DETECTOR_PATH:/opt/detector ) is fine

src/MPGDDIRC_geo.cpp
mposik1983: This is one reason. Another is that this is the only tracking code which uses rectangular geometry. The other right now I believe use  cylindrical geometries.

compact/tracker.xml
mposik1983: Ok, let me remove the tabs and I will update

compact/tracker.xml
mposik1983: Fixed

compact/tracker.xml
mposik1983: fixed

compact/tracker/mpgd_dirc.xml
mposik1983: fixed

compact/tracker/mpgd_dirc.xml
```diff
@@ -0,0 +1,185 @@
...
+  <regions>
+  </regions>
+
+  <display>
+  </display>
+
+   <detectors>
+   <detector id="MPGDDIRC_0_ID" name="MPGDDIRC_0" type="epic_MPGDDIRC" readout="MPGDDIRCHits" vis="TrackerVis">
+	    <dimensions width="MPGDDIRCModule_width" length="MPGDDIRCModule_length" height="MPGDDIRCModule_allowed_space" />
+	    <position x="0" y="0" z="MPGDDIRCModule_offset" />
```
wdconinc, line 66: There are still tabs :-)

wdconinc: `xmllint --format compact/tracker/mpgd_dirc.xml` or `sed -i `s/\t/  /g' compact/tracker/mpgd_dirc.xml` but some postprocessing would still be needed.

mposik1983: Thanks. I think the tabs are gone now😬

PR #186, opened 2022-10-07
Adding versioning to run with 275 (arches) or 41 (brycecanyon) GeV beam with B0 ECal

rahmans1:
Adding versioning to run with 275 (arches) or 41 (brycecanyon) GeV beam with B0 ECal

- [x] This should be merged following https://github.com/eic/ip6/pull/45

### What kind of change does this PR introduce?
- [ ] Bug fix (issue #__)
- [X] New feature (issue #160)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [X] Tests for the changes have been added
- [X] Documentation has been added / updated
- [X] Changes have been communicated to collaborators

### Does this PR introduce breaking changes? What changes might users need to make to their code?
No
### Does this PR change default behavior?
Introduces versioning in ip6 geometry (arches uses ecal in B0 and brycecanyon uses preshower)

PR #197, opened 2022-10-11
194 ensure correct config file yml format

rahmans1:
Ensures consistent yml format for config files. Removes the use of '-' for subattributes. 

- [X] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [X] Tests for the changes have been added

No. 

No.


wdconinc (dismissed):
Thanks! This should work better in the jinja templates.

configurations/pfrich_only.yml
```diff
@@ -1,4 +1,4 @@
 features:
-- beampipe
-- pid:
-    - pfrich
+  beampipe:
```
c-dilks, line 2:
```suggestion
  # beampipe:
```
Prefer to not have the beampipe by default, to be consistent with the dRICH. We use these `_only.yml` configurations for standalone RICH simulation tests.

c-dilks: Ah nevermind, I have that in another open PR, no need to do it here

rahmans1: Sounds good

> Ah nevermind, I have that in another open PR, no need to do it here

configurations/tracking_only.yml
```diff
@@ -1,3 +1,3 @@
 features:
-- beampipe
-- tracking
+  beampipe:
```
wdconinc, line 2: If `_only` is to mean 'no beampipe' then we should be consistent.
```suggestion
```

configurations/vertex_only.yml
```diff
@@ -1,3 +1,3 @@
 features:
-- beampipe
-- vertex
+  beampipe:
```
wdconinc, line 2: And maybe here too?
```suggestion
```

configurations/dirc_only.yml
```diff
@@ -1,4 +1,4 @@
 features:
-- beampipe
-- pid:
-    - dirc
+  beampipe:
```
wdconinc, line 2: Same here.
```suggestion
```

configurations/drich_only.yml
```diff
@@ -1,4 +1,4 @@
 features:
   # beampipe:
```
wdconinc, line 2: Remove commented lines
```suggestion
```

configurations/tracking_only.yml
rahmans1: Do you want it fixed with this pr or a separate one?

wdconinc (approved):

configurations/drich_only.yml
```diff
@@ -1,4 +1,3 @@
 features:
-  # beampipe:
   pid:
-    drich:
+    drich
```
c-dilks, line 3: Sorry to break the pattern here, but the dRICH actually needs the colon. This is because it's used a bit differently in the `jinja` template to allow loading of customized dRICH compact files on the fly (for optimizers)
```suggestion
    drich:
```

c-dilks (changes requested):

c-dilks:
Other than that, everything else looks good from what I can tell. I checked all of the resulting XML files in `jsroot`, and the detectors appear as expected. The DIRC does not appear, for issues unrelated to this PR

configurations/drich_only.yml
wdconinc: Yeah, everything should need the colon.

rahmans1:
> Other than that, everything else looks good from what I can tell. I checked all of the resulting XML files in `jsroot`, and the detectors appear as expected. The DIRC does not appear, for issues unrelated to this PR

I see a difference between epic_full.xml and epic.xml. In principle, these should be identical, right? Are the tofs expected to be in there or not?
```
diff epic_full.xml epic.xml
108,111d107
<   <include ref="${DETECTOR_PATH}/compact/tof_endcap.xml"/>
<
<   <include ref="${DETECTOR_PATH}/compact/tof_barrel.xml"/>
<
```

c-dilks:
Actually I'm not sure why we have separate `epic.xml` and `epic_full.xml`... @wdconinc do you know?

wdconinc:
> Actually I'm not sure why we have separate `epic.xml` and `epic_full.xml`... @wdconinc do you know?

Historical. We wanted to the template to parse without input and produce a 'default' epic.xml geometry. There was already a 'full' geometry, so they became two geometries.

My goal is to move away from having the template support a default geometry that doesn't require an input file (which allows us to make the template a lot simpler). And there should be no 'full' geometry either, just codenamed arches, brycecanyon, and the subdetector-only configs.

wdconinc:
So, in short, the eternal battle between backwards compatibility and forward sensibility.

rahmans1:
> > Actually I'm not sure why we have separate `epic.xml` and `epic_full.xml`... @wdconinc do you know?
> 
> Historical. We wanted to the template to parse without input and produce a 'default' epic.xml geometry. There was already a 'full' geometry, so they became two geometries.
> 
> My goal is to move away from having the template support a default geometry that doesn't require an input file (which allows us to make the template a lot simpler). And there should be no 'full' geometry either, just codenamed arches, brycecanyon, and the subdetector-only configs.

Just wanted to point out  that epic_imaging and epic_sciglass also don't have the tof in the yml. Should they be added there? On that note, is this ready for merge?

c-dilks:
> Just wanted to point out that epic_imaging and epic_sciglass also don't have the tof in the yml

I think this will be resolved when we move to `arches.yml` and `brycecanyon.yml` (see https://github.com/eic/epic/pull/126 or https://github.com/eic/epic/pull/186)

c-dilks (dismissed):

rahmans1:
Is this ready to merge? @wdconinc

wdconinc (approved):

wdconinc:
Updated, conflicts resolved, approved..

wdconinc (approved):

Issue #201, opened 2022-10-12
Solenoid needs -10*cm offset


wdconinc:
**Is your feature request related to a problem? Please describe.**
[Current offset position](https://github.com/eic/epic/blob/main/compact/definitions.xml#L380) is 0*cm, but should be -10*cm according to [detector geometry table](https://eic.jlab.org/Geometry/Detector/Detector-20220929172703.html).

**Additional context**
The magnetic field is already shifted by this -10*cm in the fieldmap.

PR #205, opened 2022-10-12
feat: update solenoid offset and length

wdconinc:
The solenoid is now slightly longer (5 cm) and shifted to the backward region by -10 cm, according to https://eic.jlab.org/Geometry/Detector/Detector-20220929172703.html. The magnetic field remains unchanged.

- [ ] Bug fix (issue #__)
- [X] New feature (issue #201)
- [ ] Documentation update
- [ ] Other: __

No.

Yes, slightly different solenoid material locations.

### DIFF
--- a/compact/definitions.xml
+++ b/compact/definitions.xml
@@ -374,10 +374,10 @@ Examples:
     </documentation>
 
     <comment>Solenoid option (BaBar magnet)</comment>
-    <constant name="Solenoid_length"           value="3790.0*mm"/>
+    <constant name="Solenoid_length"           value="3840.0*mm"/>
     <constant name="Solenoid_rmin"             value="1420.0*mm"/>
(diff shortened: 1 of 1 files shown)


osbornjd (commented):
Looks good to me, I would propose we bring it up at GD/I tomorrow just in case Elke or Renuka has any other comments, since they will be discussing the solenoid updates anyway at the start of the meeting.

wdconinc:
Material map shouldn't include solenoid anyway since outside tracking detectors.

osbornjd:
> Material map shouldn't include solenoid anyway since outside tracking detectors.

Fair enough, I just recall Elke mentioning that the material would change slightly but not in a significant way. But in any case, lets discuss tomorrow (today?).

osbornjd (approved):
Looks good to me, but I don't think I have write access so my review may not pass the CI check

PR #207, opened 2022-10-13
ecal_barrel_sciglass.xml: increase wedge box thickness from 2 to 3 mm

veprbl:

This increases the wedge box thickness to the nominal thickness.

The wedge box side walls will only correspond to 1.5 mm after this. Doubling the thickness there to nominal 3 mm and incorporating an expected 6 mm air gap is blocked by overlaps under investigation.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No


wdconinc (approved):

sly2j:
I'm a little confused about the 6mm airgap. This morning during the calorimeter meeting Elke talked about 2cm, and Tanja then wrote 1.5cm air gap behind the aluminum in the Zoom chat. Is this the same air gap behind the 3mm aluminum?

veprbl:
> I'm a little confused about the 6mm airgap. This morning during the calorimeter meeting Elke talked about 2cm, and Tanja then wrote 1.5cm air gap behind the aluminum in the Zoom chat. Is this the same air gap behind the 3mm aluminum?

I just received a number from Tanja that it should be about 17 mm. The PR used to suggest it was 7 mm, the new placement is now implemented.

wdconinc:

veprbl:
And, here, I was hoping to get rid of `BarrelExtraSpace_thickness`

sly2j:
Do we understand if there are any issues with the fact that, according to the table, the start of the ECal overlaps (0.45cm) with the DIRC support?

Screen Shot 2022-10-13 at 6 52 57 PM

veprbl:
@mariakzurek The last commit moves Imaging ECal 1cm closer. I assume that's okay, as menagerie suggests it's even 2 mm closer than that.

mariakzurek:
> @mariakzurek The last commit moves Imaging ECal 1cm closer. I assume that's okay, as menagerie suggests it's even 2 mm closer than that.

Yes, this should be all good.

PR #209, opened 2022-10-13
MPGDDIRC: fix compilation error and some format strings

veprbl:
```output
src/MPGDDIRC_geo.cpp:126:52: error: cannot pass object of non-trivial type 'std::__1::string' (aka 'basic_string<char, char_traits<char>, allocator<char>>') through variadic function; call will abort at runtime [-Wnon-pod-varargs]
        printout(DEBUG, "MPGDDIRC_geo", "gas: %f", comp_name);
                                                   ^
```
- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### DIFF
--- a/src/MPGDDIRC_geo.cpp
+++ b/src/MPGDDIRC_geo.cpp
@@ -123,13 +123,13 @@ static Ref_t create_MPGDDIRC_geo(Detector& description, xml_h e, SensitiveDetect
         max_component_length = box_length;
         gas_thickness += x_comp.thickness();
         c_box = {box_width / 2, box_length / 2, x_comp.thickness() / 2};
-        printout(DEBUG, "MPGDDIRC_geo", "gas: %f", comp_name);
+        printout(DEBUG, "MPGDDIRC_geo", "gas: %s", comp_name.c_str());
         printout(DEBUG, "MPGDDIRC_geo", "box_width: %f", box_width);
-        printout(DEBUG, "MPGDDIRC_geo", "box_length: %%f", box_length);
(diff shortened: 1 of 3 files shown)


mposik1983 (dismissed):

src/MPGDDIRC_geo.cpp
```diff
@@ -123,13 +123,13 @@ static Ref_t create_MPGDDIRC_geo(Detector& description, xml_h e, SensitiveDetect
         max_component_length = box_length;
         gas_thickness += x_comp.thickness();
         c_box = {box_width / 2, box_length / 2, x_comp.thickness() / 2};
-        printout(DEBUG, "MPGDDIRC_geo", "gas: %f", comp_name);
+        printout(DEBUG, "MPGDDIRC_geo", "gas: %s", comp_name.c_str());
```
wdconinc, line 126: Hmm. it seems like our CI should have caught this... Maybe gcc vs clang? Suggestions on how to avoid?

wdconinc (dismissed):

src/MPGDDIRC_geo.cpp
veprbl: This was using clang

wdconinc: #215 should hopefully avoid this

wdconinc (approved):

wdconinc:
Fixed two more unused variables which prevented `-Werror` compilation with clang (and clang-format for pre-commit). This will now allow #215 to succeed.

PR #219, opened 2022-10-14
ci: run tgeo overlap check with sampling too

wdconinc:
This introduces another TGeo overlap check, based on the sampling approach. This is more similar to the Geant4 approach, but faster. It catches more than the default mesh-based approach. 

- [ ] Bug fix (issue #__)
- [X] New feature (issue: additional overlap checks)
- [ ] Documentation update
- [ ] Other: __

No.

No.


.github/workflows/linux-eic-shell.yml
```diff
@@ -221,6 +221,9 @@ jobs:
 
   check-overlap-tgeo:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        option: ['', 's']
```
wdconinc, line 226: Apparently the empty string is failing to get parsed correctly... but any letter will do.
```suggestion
        option: ['m', 's']
```

wdconinc:

Issue #127, opened 2022-09-22
pfRICH: rescale the geometry for the `BryceCanyon` configuration


c-dilks:
**Is your feature request related to a problem? Please describe.**
Rescale the pfRICH for the `BryceCanyon` configuration (see #126)

**Describe the solution you'd like**

Rescaled envelope parameters (TBD):
```
Length ~44 cm
Proximity gap 30 cm
Services (LAPPD) 10 cm
Z position (end): -172.7 cm
```
See [Elke's slides](https://indico.bnl.gov/event/17081/) for more details

For reference, the parameters from ATHENA:
```
PFRICH_Length                  =       58.000 = BackwardRICHRegion_length
PFRICH_aerogel_thickness       =        3.000 = 3.0*cm
PFRICH_num_px                  =        8.000 = 8
PFRICH_rmax                    =       93.000 = BackwardPIDRegion_rmax - 2*cm
PFRICH_rmin0                   =        5.006 = BackwardPIDRegion_tan * BackwardRICHRegion_zmin
PFRICH_rmin1                   =        6.942 = BackwardPIDRegion_tan * (BackwardRICHRegion_zmin + BackwardRICHRegion_length)
PFRICH_sensor_active_size      =        2.400 = PFRICH_sensor_active_size_default
PFRICH_sensor_dist             =       40.000 = 40*cm
PFRICH_sensor_full_size        =        2.580 = PFRICH_sensor_full_size_default
PFRICH_sensor_thickness        =        0.150 = 1.5*mm
PFRICH_wall_thickness          =        0.500 = 0.5*cm
PFRICH_window_thickness        =        0.100 = 0.1*cm
PFRICH_zmax                    =     -208.000 = PFRICH_zmin - PFRICH_Length
PFRICH_zmin                    =     -150.000 = -BackwardRICHRegion_zmin
```

wdconinc:
We can just roll this out into the main branch before #126 and pick it up in a rebase there. (Same will need to happen for mRICH too.)

PR #223, opened 2022-10-18
Fix tracker support overlaps

### LINKED ISSUE #208: Fix outer tracker service routing overlaps
The outer tracker service routing overlaps the pfRICH (and probably mRICH) in the negative end, and the 90 degree turn conflicts with the imaging ecal in the positive end. I suggest (pragmatically) to:
* shorten second projective cone on the negative end to the face of the pfRICH, and have a discontinuous step to a cylinder outside of the pfRICH/mRICH
* replace 90 degree turn with an extension of the cone near the dRICH in the positive end for the dRICH end

sly2j:
Remove outer detail from forward tracker support, and make backward tracker support discontinuous

Also fixes some of the central detector volume allocations.

Closes #208

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### DIFF
--- a/compact/mrich.xml
+++ b/compact/mrich.xml
@@ -3,7 +3,7 @@
   <define>
     <constant name="MRICH_rmin"   value="8*cm"/>
     <constant name="MRICH_rmax"   value="min(63*cm, BackwardPIDRegion_rmax)"/>
-    <constant name="MRICH_length" value="BackwardPIDRegion_length"/>
+    <constant name="MRICH_length" value="BackwardPIDRegion_length-28.87*cm"/>
     <constant name="MRICH_zmax"   value="BackwardPIDRegion_zmax"/>
     <constant name="MRICH_zmin"   value="MRICH_zmax-MRICH_length"/>
 
--- a/compact/mrich.xml
+++ b/compact/mrich.xml
@@ -3,7 +3,7 @@
   <define>
     <constant name="MRICH_rmin"   value="8*cm"/>
     <constant name="MRICH_rmax"   value="min(63*cm, BackwardPIDRegion_rmax)"/>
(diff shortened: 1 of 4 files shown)


compact/mrich.xml
```diff
@@ -3,7 +3,7 @@
   <define>
     <constant name="MRICH_rmin"   value="8*cm"/>
     <constant name="MRICH_rmax"   value="min(63*cm, BackwardPIDRegion_rmax)"/>
-    <constant name="MRICH_length" value="BackwardPIDRegion_length"/>
+    <constant name="MRICH_length" value="BackwardPIDRegion_length-28.87*cm"/>
```
wdconinc, line 6: This I don't understand.

sly2j: This shifts the mrich back by the difference in length between the mrich and the pfrich to ensure they are both placed at the back of the PID region. Not sure how else to do this, except to maybe introduce an extra parameter in the top level definitions  file (which would be a bit cleaner)

c-dilks: I should double-check the `dawn` views again, but I think we resolved the pfRICH/mRICH placement in #127 (just look at `definitions.xml` and `{pf,m}rich.xml`, there are several workaround/test commits in that PR that need to be reverted)

wdconinc: > This shifts the mrich back by the difference in length between the mrich and the pfrich

That's where I am confused: the difference is 11.4 cm (pfRICH 54.1 cm - mRICH 42.7 cm), not 28.87 cm.

c-dilks: Let's resolve this in #158

wdconinc: Really? I'm just done here.

PR #225, opened 2022-10-18
Updated ZDC ids to point to older ZDC ATHENA geometry for this campaign.

ajentsch:
The PR switches the 5 hit IDs (163 - 167) used the FoCal ZDC back to the 2 IDs used (163-164) for the old ATHENA ZDC. This is because the digitization is problematic for the FoCal design at the moment.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [x] Documentation update
- [x] Other: Revert to older geometry to match existing digitization.

It should not - it should AVOID a breaking issue at digitization.

It should make the default behavior work properly.


wdconinc (approved):
If this passes the checks🤞 it looks good to me. I fear it won't because it may now be assuming to find the new hit collections.

wdconinc:
https://github.com/eic/ip6/pull/48 merged and all checks resubmitted here.

PR #228, opened 2022-10-18
epic.xml.jinja2: fix logic for farforward

veprbl:
Fixes an issue introduced in https://github.com/eic/epic/pull/186

The current logic enables farforward whenever `'farforward' not in features` which makes it harder to disable farforward.


veprbl:
My PR should work in case when features are defined, but farforward is not

c-dilks:
The rendered `xml` files are nearly identical to #222, the only difference is a blank line.

c-dilks:
I guess https://github.com/eic/epic/issues/229 is the reason the CI checks fail

c-dilks (approved):

wdconinc (approved):

PR #232, opened 2022-10-19
fix: extend HcalEndcapN r segments to rmin/rmax

wdconinc:
We are encountering issues with tracks that return numerous warnings like this:
```console
ERROR: Hit Position (1.32962276827011169900e+01) is below the acceptance (1.99431000000000011596e+01) of the segmentation
Position: Pre (                       (8.35509855972314454675e+01,-1.03429629682137914415e+02,-3.34259732049230251505e+03)) Post (                       (8.35453556905108314368e+01,-1.03439123309562916120e+02,-3.34259315786786373792e+03)) 
Momentum:  Pre (                       (-9.01134991035029281337e-02,-1.52323523319139481069e-01,6.67403254221919650657e-02))  Post (                       (-0.00000000000000000000e+00,-0.00000000000000000000e+00,0.00000000000000000000e+00)) 
```
These positions are in the HcalEndcapN. This extends the r segmentation to the rmin and rmax.

- [X] Bug fix (issue: segmentation issues in HcalEndcapN)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [X] Changes have been communicated to collaborators @lkosarz 

No.

No.


lkosarz (commented):
That was intentional to keep the geometry as it is, but limit the segmentation to a smaller area. It's not perfect and gives warnings. As far as I know, the additional inner and outer segments are not needed. Me preference is to keep the segmentation as it is. If needed we can shrink the volume to match the segmentation. The size of segments increases linearly with eta, so adding the remaining outer and inner parts may not match this scaling.

wdconinc:
We cannot not run with these warnings. They are numerous enough that they make the logs useless. So we will have to change the volume to be exactly along those r values.

lkosarz:
I understand, then I need to check with others whether to extend the segments o shrink the volume. I'll get back to it.

lkosarz:
I made a final update to instead shrink the volume to the segments. This is preferred, because adding additional segments will affect the clustering etc. That is why I updated the hcal-backward-segmentation-r branch.

wdconinc:
Ok, let me know when there is a pull request.

lkosarz:
I created the pull request now: https://github.com/eic/epic/pull/255
Thanks!

PR #233, opened 2022-10-19
ci: build faster with -j 2

wdconinc:
Building the geometry plugins is on the critical path. Faster jobs will result in faster pipeline completion. We have two CPUs on runners, so we should use them.

- [ ] Bug fix (issue #__)
- [X] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


wdconinc:
Minor change, builds succeeded, force merging.

PR #236, opened 2022-10-19
fix: decrease negative eta of cutout sciglass envelope

wdconinc:
The DIRC and Ecal barrel sciglass envelope volume were overlapping about 2 cm at the DIRC readout box. This change from -1.9 to -1.87 does not introduce overlaps in the sciglass and removes overlaps between the DIRC and the sciglass. With an MCP height of 23.746 cm starting at 70 cm radius, the DIRC is well within the 100 cm outer radius, so I think modifying the sciglass is the right approach here.

- [X] Bug fix (issue overlap fixed)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [X] Tests for the changes have been added: DIRC enabled in default `epic.xml`
- [X] Changes have been communicated to collaborators @veprbl 

No.

No.

### DIFF
--- a/compact/ecal_barrel_sciglass.xml
+++ b/compact/ecal_barrel_sciglass.xml
@@ -11,7 +11,7 @@
       The current numbers are not based on any design in particular, but
       provide some extra space for other detectors.
     </comment>
-    <constant name="EcalBarrel_etamin" value="-1.9"/>
+    <constant name="EcalBarrel_etamin" value="-1.87"/>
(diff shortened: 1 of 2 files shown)


wdconinc:
@veprbl I'm going to force-merge this in the interest of moving forward with other issues. The pipelines succeed, no overlaps internal or external to the sciglass.

veprbl:
Reducing the negative eta cut is definitely okay as we have "extra" towers there.

PR #238, opened 2022-10-19
feat: use upstream DD4hep materials

wdconinc:
DD4hep installs a materials.xml file. We should use the materials there if available.

- [X] Bug fix (issue: material duplication with potential for inconsistent definitions)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/templates/epic.xml.jinja2
+++ b/templates/epic.xml.jinja2
@@ -55,6 +55,7 @@
 
   <includes>
     <gdmlFile ref="${DD4hepINSTALL}/DDDetectors/compact/elements.xml"/>
+    <gdmlFile ref="${DD4hepINSTALL}/DDDetectors/compact/materials.xml"/>
     <gdmlFile ref="${DETECTOR_PATH}/compact/materials.xml"/>
     <file     ref="${DETECTOR_PATH}/compact/optical_materials.xml"/>
   </includes>


veprbl:
I thought about this one and wasn't sure if we want to use materials from DD4hep. Unlike the periodic table, the materials are not constant and unambiguous, I'm afraid this will bite us sooner or later.

wdconinc:
> the materials are not constant and unambiguous

I get your concern, but shouldn't well-defined materials be constant and unambiguous, in a sense that they asymptotically get close to their 'true' value? I would rather have a conscious discussion about why we have three differrent definitions (with up to 20% density differences) for what is essentially quartz, only differing in their generic name, than someone adding a fourth definition. Requiring people to justify addition of materials by not finding it in an externally curated data base means there is a bit of effort that goes into that. I wish we could refer to the entire NIST database, which at least has a provenance record.

veprbl:
From the QA perspective, expecting possible material changes on each DD4hep update is an unreasonable burden. If they are not immutable, then how are we going to monitor and let the experts review the changes? To say that they will only "improve" is to require guarantees from the upstream that we don't yet have.

I expect, the materials that are to be used will come with datasheets from the manufacturer, there will be nothing "true" about those compositions. They may get discontinued and disappear from the market before some next detector is built.Not to mention the optical properties that may often come from some in-house measurements.

I think, it would be understandable, if we were using DD4hep materials for our prototype geometries with intent to replace them with ePIC-specific ones for the future productions. But doing it in assumption that DD4hep will know better is a bit optimistic IMO.

wdconinc:
Closing since @veprbl points out some very legitimate issues with this. Not sure how to 'sanitize' or treatment and duplication of materials, but I agree that this is unlikely to be it.

Issue #234, opened 2022-10-19
Geant4VolumeManager INFO  +++   Bad volume Geant4 Path: /AV_836!MPGDDIRC_0_31#31!component0_0#0


wdconinc:
### Environment: (where does this bug occur, have you tried other environments)
- Which branch (`main` for latest released): `main`
- Which revision (`HEAD` for the most recent): 
- Any specific OS or system where the issue occurs? eic-shell
- Any special versions of ROOT or Geant4? eic-shell

### Steps to reproduce: (give a step by step account of how to trigger the bug)
See, e.g. https://eicweb.phy.anl.gov/EIC/benchmarks/physics_benchmarks/-/jobs/1093877#L900

No reproducer yet.

### Expected Result: (what do you expect when you execute the steps above)
Expect no warnings.

### Actual Result: (what do you get when you execute the steps above)
Getting warnings.

mposik1983:
@wdconinc I just took a closer look at some ddsim events I generated and I do see this message in the output as well. I must have missed it the first time due to all of the other output text that get printed.

mposik1983:
@sly2j That was my first thought as well and where I was going to start. I will try to find some time today and take a shot at fixing it myself, but I may need some help in which case I'll take you up on your offer. I'll let you know tonight where I am. Thanks!

wdconinc:
This was resolved by #241

PR #241, opened 2022-10-20
Mpgddirc geo fix

mposik1983:
Fixes Bad volume warnings listed in Issue #234 
https://github.com/eic/epic/issues/234 

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No, it should not
No

### DIFF
--- a/compact/tracker.xml
+++ b/compact/tracker.xml
@@ -162,7 +162,6 @@
 	    <include ref="tracker/silicon_barrel.xml"/>
 	    <include ref="tracker/mpgd_barrel.xml"/>
             <include ref="tracker/mpgd_dirc.xml"/>
-            <include ref="dirc.xml"/>
 	    <include ref="tracker/silicon_disks.xml"/>
 	    <include ref="tracker/support_service_assembly.xml"/>
 	    <include ref="tof_barrel.xml"/>


compact/tracker.xml
```diff
@@ -158,13 +158,14 @@
...
-  <include ref="tracker/mpgd_dirc.xml"/>
-  <include ref="tracker/silicon_disks.xml"/>
-  <include ref="tracker/support_service_assembly.xml"/>
-  <include ref="tof_barrel.xml"/>
-  <include ref="tof_endcap.xml"/>
+	    <include ref="tracker/vertex_barrel.xml"/>
+	    <include ref="tracker/silicon_barrel.xml"/>
+	    <include ref="tracker/mpgd_barrel.xml"/>
+            <include ref="tracker/mpgd_dirc.xml"/>
+            <include ref="dirc.xml"/>
```
wdconinc, line 165: Isn't this going to load the DIRC twice?

compact/tracker.xml
```diff
@@ -158,13 +158,14 @@
...
+	    <include ref="tracker/silicon_disks.xml"/>
+	    <include ref="tracker/support_service_assembly.xml"/>
+	    <include ref="tof_barrel.xml"/>
```
wdconinc, line 168: Same with TOF.

compact/tracker.xml
mposik1983: Yes, I had it there to see the dirc with the tracking system. I forgot to remove it before I pushed

compact/tracker.xml
wdconinc: Nevermind this was in tracker already

mposik1983:
@wdconinc is this pr failing the epic_brycecanyon benchmarks because the mpgd_dirc is not supposed to be in that configuration (the one with the silicon layer in the barrel calorimeter)?

wdconinc:
I don't think that silicon layer is implemented, and it wasn't on Silvia's latest presentation. I think both arches and brycecanyon are using the DIRC MPGD detector.

wdconinc:
This is failing because brycecanyon doesn't have mrich, and until yesterday we had the mrich in brycecanyon by mistake.

wdconinc:
I'll fix the physics benchmarks to do this correctly.

mposik1983:
Ok, thanks for the clarification. I didn't realize the dirc mpgd was going to be in both configurations. Good to know :)

wdconinc:
This is what I have from Silvia's presentation. Also the detector parameter table has the DIRC MPGD in the baseline and no alternative.

image

sly2j:
That is a mistake. BryceCanyone does NOT have the MPGD behind the DIRC. The imaging ECAL first layer serves the same role.

sly2j:
Screen Shot 2022-10-20 at 12 08 04 PM
FYI, this is the spreadsheet from GD&I. It is important we do this right to have a meaningful comparison.

wdconinc (approved):
Looks good. Thanks for fixing (I assume it's fixed and no more warnings).

Issue #138, opened 2022-09-27
Default `epic.xml` does not have DIRC or TOF


wdconinc:
### Environment: (where does this bug occur, have you tried other environments)
- Which branch (`main` for latest released): `main`
- Which revision (`HEAD` for the most recent): `HEAD`
- Any specific OS or system where the issue occurs? All
- Any special versions of ROOT or Geant4? All

### Steps to reproduce: (give a step by step account of how to trigger the bug)
1. `cmake` and `make install`

### Expected Result: (what do you expect when you execute the steps above)
The `epic.xml` was thought to be the default configuration, which you get when running the jinja2 template without a specific configuration yaml file. The logic in the template is intended to allow this, but that's not always easy. In several places we have to add things like `if features is not defined` and stuff to allow running without `features`. That's quickly turning into a cumbersome approach, and it is not clear to those editing the template that that is what's going on.

### Actual Result: (what do you get when you execute the steps above)
We would like to think think that the DIRC is in the default, but it isn't included in `epic.xml` because the logic is inverted there.

### Probable Solution:
We should probably just move away from a default `epic.xml`, or at least require an explicit configuration for it. This can probably most naturally be achieved by picking a default `config` argument in `bin/make_detector_configuration`. We may end up with `epic_default.xml` and `epic.xml` being the same, but that's likely clear enough.

wdconinc:
Right, here is why this was such a head ache last time:
```yaml
features:
- beampipe
- pid:
  - dirc
```
parses as
```python
['beampipe', {'pid': ['dirc']}]
```
so `'beampipe' in features` is true, but `'pid' in features` is false.

wdconinc:
Ah, and:
```yaml
features:
  beampipe:
  pid:
    dirc:
```
parses as
```python
{'beampipe': None, 'pid': {'dirc': None}
```
(a dict!)

while
```yaml
features:
  beampipe
  pid
    dirc
```
parses as
```python
beampipe pid dirc
```
(a string!)

PR #244, opened 2022-10-20
feat: define default features inside jinja template, and assume features are file names

wdconinc:
Having to allow for a non-existing `features` variable inside the jinja template has been an annoying headache. It means that the default configuration (that gives us `epic.xml`) is defined in an entirely different way than the configuration files. This PR introduces a single `features` definition at the top of the template for when `features` is not defined (it is still slightly different because jinja is not yaml, but it is much closer). The rest of the template can just assume that `features` exists and legitimately crash when it doesn't.

The other aspect that this introduces is looping over the entries in each detector type to load them. Rather than checking whether the `pid` includes `drich` and then including `drich.xml`, we now just load everything in the `pid` features, directly, interpreted as an xml file. If you create a working prototype `trich.xml` in `compact/pid/` (with triple radiators), you can simply add `trich` to the configurations.

- [x] Bug fix (issue #138)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No. This should result in the exact same xml files as before.

No. This should result in the exact same xml files as before.

TODO
- [ ] double check when all commits in this branch are final


templates/epic.xml.jinja2
```diff
@@ -9,6 +9,44 @@
...
+  'ecal': {
+    'forward_homogeneous': '',
+    'barrel_sciglass': '',
+    'backward_PbWO4': '',
+  },
+  'solenoid': '',
+  'farforward': {
+    'arches': '',
+  },
+  'farbackward': ''
```
wdconinc, line 46: Needs content since looped over.

templates/epic.xml.jinja2
```diff
@@ -9,6 +9,44 @@
    without any other variables defined.
 -#}
 
+{# Absolute default configuration, defines features -#}
+{% if features is not defined -%}
+{%   set features = ( {
```
veprbl, line 14: This only covers the case when template is instantiated by an empty file? Might as well keep the full feature file as a reference instead.

wdconinc: Not sure I understand what you mean.

wdconinc: It is indeed what we get for an empty yaml file, which is what we produce as the default `epic.xml` that most checks are run against (i.e. the reference design without the alternatives). We are already hitting the limits of GitHub actions, so we can't double all of the pipelines. Originally, we wanted the template to be simply a way to avoid having lots of nearly duplicated xml files. The idea was always that running without input would give the full detector in the current main reference design. I think that's probably something to maintain. It is possible that after arches and brycecanyon we will run more codenamed versions, but I think if we make a decision that we for sure will use the mRICH for example, then we'd update the default epic.xml to reflect that.

Not sure if that answers the second part of your comment. Maybe you are suggesting we should keep a `default.yml` file around to then parse into epic.xml, or treat`full.yml` like that? That's another option indeed, but means that the jinja template requires context whereas this was an attempt to make the template useful as a standalone entity.

veprbl: I guess, my main concern is that this gets rid of `features is not defined or ` but does nothing for `'foo' is in features['bar'] or `.

veprbl: Using `full.yml` might be better

c-dilks:
Do I understand correctly that the idea from https://github.com/eic/epic/pull/14 will be made easier with this? That is:
```yaml
features:
  pid:
    my_custom_drich
```
would look for `compact/pid/my_custom_drich.xml`? 

See that PR for more context on how this is useful for optimization scripts.

wdconinc:
> Do I understand correctly that the idea from https://github.com/eic/epic/pull/14 will be made easier with this?

That is the idea, indeed.

c-dilks:
Could you try to update this branch w.r.t. `main`? I'm not sure how to resolve the conflicts for the `tracker` compact files inclusion.

wdconinc:
> Could you try to update this branch w.r.t. main?

Will do after #251 .

templates/epic.xml.jinja2
wdconinc: I think I understand the concern better now: this PR attempts to handle the absence of `features` more gracefully but it does not handle the absence of an interable as `features['bar']`.

I think that's ok, since the goal is to have a sensible default that does not have a suffix (which is the user contract for files in `configurations/`), not to handle malformed input configuration (which should rightfully fail). Therefore, I think it is ok to fail when getting `features: pid:` without specifying *what* is in `pid:`.

To silently accept an empty `features: pid:`, we could use `{% if 'pid' in features and features['pid'] is iterable %}`, but I feel that the intent of including `features: pid:` is likely to include all default PID.

In any case, I'm thinking about how to make this a bit easier to manage, more expressive, and less error prone.

veprbl: Yes, from you original motivation

> It means that the default configuration (that gives us `epic.xml`) is defined in an entirely different way than the configuration files.

The approach from this PR will only further the discrepancy, by making the `epic.xml` into an one-off case.

wdconinc: So, you're proposing to get rid of epic.xml entirely?

wdconinc: And then we'd only have named configurations left.

wdconinc: Removed defaults in jinja; using just `full.yml` as default which installs as `epic.xml`.

templates/epic.xml.jinja2
```diff
@@ -92,136 +88,79 @@
      - B0 tracker
      - Off-momentum tracker
      - Far forward roman pots
-     - Zero Degree Calorimeter
+     - Zero Degree Calorimetergit restore --staged
```
veprbl, line 91: This needs some attention

wdconinc: The disadvantage of Lenovo track points...

wdconinc:
```suggestion
    - Zero Degree Calorimeter
```

.github/workflows/linux-eic-shell.yml
```diff
@@ -207,6 +219,30 @@ jobs:
...
         GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+  check-geometry-configs:
+    runs-on: ubuntu-latest
+    needs:
+      - build-gcc
+      - list-detector-configs
+    strategy:
+      matrix:
+        detector_config: ${{ fromJson(needs.list-detector-configs.outputs.configs) }}
```
wdconinc, line 229: I think this probably only works in dispatched workflows, not inside the same workflow. I have to look back at the other times I've used this.

veprbl: Should we just write a for-loop in bash, for now?

veprbl: Or provide the list manually?

wdconinc: Yes that would be the easy way out. But I'd like to reuse the list for spinning up parallel geoviewer jobs, and for parallel geometry checking jobs without fail-fast. The serial model surfaces one error at a time. The parallel model surfaces them all at once.

Anyway, if it takes me more than 10 minutes to figure this out when I'm at a computer, then I'm not going to bother and just use the bash for loop.

templates/epic.xml.jinja2
```diff
@@ -97,131 +93,74 @@
...
 
-{% if features is not defined or 'ecal' in features %}
+{% if 'ecal' in features %}
   <documentation level="10">
     ## Central EM calorimetry
   </documentation>
-  {% if features is not defined or features['ecal'] is none or 'forward_homogeneous' in features['ecal'] %}
-  <include ref="${DETECTOR_PATH}/compact/ecal_forward_homogeneous.xml"/>
-  {% elif 'forward_scifi' in features['ecal'] -%}
-  <include ref="${DETECTOR_PATH}/compact/ecal_forward_scfi.xml"/>
```
veprbl, line 159: So which one it is, "scifi" or "scfi"? :)

wdconinc: Historically? ScFi for imaging ecal. But we can't define what the forward ecal calls it...

veprbl (approved):
Looks good

PR #246, opened 2022-10-24
test: check overlaps with reduced tolerance

c-dilks:
Overlap checks with reduced tolerance to check for dRICH sensor-sensor overlaps.

Do not merge.


c-dilks:
dRICH sensor-sensor overlaps, as expected. But there are also the following overlaps:
```
Overlap is detected for volume wrapper_vol_3:3 (G4SubtractionSolid) with its mother volume module_vol (G4Box)
```

PR #247, opened 2022-10-24
fix(dRICH): increase gap between sensors

c-dilks:
### Briefly, what does this PR introduce?
This should stop transient dRICH overlaps, since there are likely still small overlaps that are not being detected by overlap checks at the default tolerance (0.1 mm).

### What kind of change does this PR introduce?
- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [x] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [x] Changes have been communicated to collaborators

### Does this PR introduce breaking changes? What changes might users need to make to their code?
no

### Does this PR change default behavior?
no

PR #248, opened 2022-10-24
test: check overlaps with reduced tolerance, after increasing dRICH sensor gaps

c-dilks:
cf. #246 from before increasing dRICH sensor gaps (#247)

Do not merge.


c-dilks:
In `check-overlap-geant4-fast`, only #249 remains.

PR #252, opened 2022-10-24
feat: de-duplicate imaging/sciglass EcalBarrelHits

wdconinc:
A collection name should be used for all hits that are similar. `EcalBarrelHits` was used for both sciglass and imaging (silicon pixel) hits. Those are not (at all) comparable and are more appropriately given different names.

- [X] Bug fix (issue: name clashes in EICrecon tag resolution)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [X] Tests for the changes have been added:
  - [ ] TODO: will need changes to juggler and eicweb benchmarks (sigh)
- [X] Changes have been communicated to collaborators @veprbl @mariakzurek @faustus123 

Yes. Calorimeter simulated hit collections are being renamed without backwards compatibility.

Yes.

### DIFF
--- a/compact/ecal_barrel_interlayers.xml
+++ b/compact/ecal_barrel_interlayers.xml
@@ -176,7 +176,7 @@
   </detectors>
 
   <readouts>
-    <readout name="EcalBarrelHits">
+    <readout name="EcalBarrelImagingHits">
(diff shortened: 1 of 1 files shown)


wdconinc:
This will temporarily (and necessarily) break benchmarks, so I'm going to force-merge.

Issue #243, opened 2022-10-20
Adjust hole dimensions for insert after fixed z position


rymilton:
**Is your feature request related to a problem? Please describe.**
With the correct z-position for the ECal and HCal inserts, the hole positions and sizes need to be adjusted. Will depend on [PR #168](https://github.com/eic/epic/pull/168).

**Describe the solution you'd like**
Adjust the hole position and size to maintain 4 cm of clearance w.r.t. beampipe.

**Describe alternatives you've considered**
N/A

**Additional context**
Will hopefully get this done tomorrow night.

PR #254, opened 2022-10-25
Update insert hole parameters, fix backplate attribute bug

rymilton:
Updated the forward insert's hole initial/final radius and x-position after the z positions of the forward endcap were changed to managerie values (329.6 cm for ECal, 359.6 for HCal).
Fixed bug where the attribute backplate_thickness wasn't being read in.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #243 )
- [ ] Documentation update
- [ ] Other: __

None
Adjusts default shape of the forward inserts

### DIFF
--- a/src/InsertCalorimeter_geo.cpp
+++ b/src/InsertCalorimeter_geo.cpp
@@ -49,7 +49,7 @@ static Ref_t createDetector(Detector& desc, xml_h handle, SensitiveDetector sens
   const std::pair<double, double> hole_y_parameters(hole_y_initial, hole_y_final);
 
   // Getting thickness of backplate
-  auto backplate_thickness = dd4hep::getAttrOrDefault<double>(detElem, _Unicode(backplate_thickness), 0. * cm);
+  auto backplate_thickness = detElem.hasChild(_Unicode(backplate))? detElem.child(_Unicode(backplate)).attr<double>(_Unicode(thickness)) : 0.;
 
   // Function that returns a linearly interpolated hole radius, x-position, and y-position at a given z
   auto get_hole_rxy = [hole_radii_parameters, hole_x_parameters, hole_y_parameters, length,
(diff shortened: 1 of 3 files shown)


src/InsertCalorimeter_geo.cpp
```diff
@@ -49,7 +49,7 @@ static Ref_t createDetector(Detector& desc, xml_h handle, SensitiveDetector sens
   const std::pair<double, double> hole_y_parameters(hole_y_initial, hole_y_final);
 
   // Getting thickness of backplate
-  auto backplate_thickness = dd4hep::getAttrOrDefault<double>(detElem, _Unicode(backplate_thickness), 0. * cm);
+  auto backplate_thickness = detElem.hasChild(_Unicode(backplate))? detElem.child(_Unicode(backplate)).attr<double>(_Unicode(thickness)) : 0.;
```
wdconinc, line 52: This looks a bit janky (ternary operator is not very readable). I assume there's somewhere else another `if` to then place the backplate or not.

rymilton: Ah, no there's not actually. The backplate is included in the layer section that's placed via the existing for loop. But I need the thickness of the final layer (which is the backplate) before the for loop to make a function that calculates the radius and position of the beampipe hole.

rymilton: Err if that wasn't clear, the hole shape is determined by a linear interpolation from the front of the insert to the front of the final layer. So I need the (total length - final layer length) to calculate the hole info.

wdconinc: Nah, this is fine. Maybe just add a comment here to clarify.

wdconinc (approved):

wdconinc:
Restarted failing benchmarks after benchmarks bug fix.

PR #258, opened 2022-10-26
SciGlassCalorimeter_geo: implement double walls for wedge box and 6 mm gap

veprbl:
Replaces a single 3 mm wall with two 3 mm walls separated by a 6 mm Air Gap
image

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

Yes

### DIFF
--- a/compact/ecal/barrel_sciglass.xml
+++ b/compact/ecal/barrel_sciglass.xml
@@ -49,7 +49,7 @@
 	surfaces of individual wedge boxes with a single shared tube at a fixed
         `inner_r`, which allows for ~17 mm air gap before the towers.
       </comment>
-      <wedge_box inner_r="78.55 * cm" thickness="3 * mm" vis="becal_wedge" material="Aluminum5083" />
+      <wedge_box inner_r="78.55 * cm" thickness="3 * mm" gap="6 * mm" vis="becal_wedge" material="Aluminum5083" />
 
       <sectors number="24" phi0="0." deltaphi="pi / 12">
         <comment>
--- a/compact/ecal/barrel_sciglass.xml
+++ b/compact/ecal/barrel_sciglass.xml
@@ -56,7 +56,7 @@
           Rows within a sector are supposed to have smaller gaps than edge rows of two adjacent sectors
         </comment>
 
-        <rows number="5" deltaphi="2.92 * degree">
+        <rows number="5" deltaphi="2.811 * degree">
           <dimensions inner_r="EcalBarrel_rmin + EcalBarrel_inner_margin"  gap="1 * mm" />
 
           <comment>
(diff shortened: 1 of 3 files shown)


wdconinc:

wdconinc:
FYI: Merge conflicts due to #244.

compact/ecal/barrel_sciglass.xml
```diff
@@ -67,23 +67,23 @@
...
-          <family dir_sign="-1" x1="2 * cm" y1="2 * cm" z_length="45.5 * cm" number="6"  flare_angle_polar="1.045 * degree" flare_angle_at_face="0.7058 * degree" vis="family2" />
-          <family dir_sign="-1" x1="2 * cm" y1="2 * cm" z_length="45.5 * cm" number="5"  flare_angle_polar="0.929 * degree" flare_angle_at_face="0.8892 * degree" vis="family3" />
-          <family dir_sign="-1" x1="2 * cm" y1="2 * cm" z_length="45.5 * cm" number="5"  flare_angle_polar="0.82 * degree"  flare_angle_at_face="1.0258 * degree" vis="family4" />
-          <family dir_sign="-1" x1="2 * cm" y1="2 * cm" z_length="45.5 * cm" number="5"  flare_angle_polar="0.70 * degree"  flare_angle_at_face="1.126 * degree"  vis="family5" />
-          <family dir_sign="-1" x1="2 * cm" y1="2 * cm" z_length="45.5 * cm" number="6"  flare_angle_polar="0.586 * degree" flare_angle_at_face="1.1914 * degree" vis="family6" />
+          <family dir_sign="+1" x1="1.925 * cm" y1="2 * cm" z_length="45.5 * cm" number="10" flare_angle_polar="1.16 * degree"  flare_angle_at_face="0.3059 * degree" vis="family1" />
```
wdconinc, line 70: Is `x1` then something like `2 * cm - gap * trig_function(angle)` for some `trig_function` and some `angle`? They seem correlated from this PR.

veprbl: In principle, there may be a merit to your "subtractive" parametrization (e.g. for numerical optimizations). However, with a limitation of towers having to come in families of identical shapes, well defined gaps are not quite possible. On the other hand, the additive parametrization can be compared to the Molière radius.

wdconinc: Yeah, I understand that these 'perfect' parametrizations are not possible. But I have to keep asking anyway to prevent people (not you) from adding hard-coded constant everywhere :-)

At this point we have a geometry that is guaranteed to change several more times between now and when stuff is actually built in the real world. I want those changes to be as painless as possible, in particular by being as automatic as possible. Change one number and the rest follows.

The other part of this is the ai/ml side of detector optimization. That only works if you can continuously change the input parameters and still end up with a reasonable geometry. Of course, I'm not saying we'll want to add the air gap thickness as input into an ai/ml optimization of the sciglass calorimeter, but I could see us adding block size at some point. And in that case having derived values rather than values hard-coded in different places may be useful.

That's my motivation, which is mostly focused on future efforts and I understand that in the interest of expediency right now we will want to just put in some numbers and be done with it.

veprbl: I appreciate your effort to keep thinking about improving this. The points I agree with, except that this parametrization is not constructed for expediency, it matches what the engineers communicate to us, and what makes sense for the geometry. By the way, the invalid parameters (such as "negative" gaps) can be detected by running a quick overlap check.

wdconinc: I apologize, the comment about expediency was not applicable to this case. That was a more general comment for the kind of scenarios where we may choose to deviate from parametrization.

wdconinc (approved):

PR #261, opened 2022-10-28
fix: split overlap check jobs for barrel hcal

wdconinc:
With the merge of the barrel hcal, we have temprorarily failing overlap checks due to (it appears) a bug in the underlying classes. While this is being solved, we want to continue having a working CI system, so we disable the barrel hcal from some jobs. This PR moves stuff around so at least the eicweb benchmarks pipelines are started again (even if they may fail).

- [X] Bug fix (issue: eicweb benchmark pipelines are not started)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -372,7 +372,33 @@ jobs:
     strategy:
       matrix:
         option: ['m'] #, 's'] # FIXME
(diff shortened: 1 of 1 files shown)


wdconinc:
All required checks are passing; fialing checks are expected due to the barrel hcal.

PR #266, opened 2022-10-31
Simplified tile shapes for outer HCAL

johnlajoie:
Simplified description of tile solid shapes for outer HCAL

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [X] Other: Optization

No breaking changes found/expected

The tile shapes have been reduced to about 10% of their original complexity.  Most of the complexity in terms of the facets describing the tiles was associated with small mounting holes in the tiles SiPM mount/light blocker. This is unnecessary for the description of the tiles in simulations and adds much more complexity, which increases the time required for G4 to decide where it is in a solid.  This should result in a reduced time in traversing the oHCAL geometry and simplify things like the overlap checks. 

Unfortunately, this does not affect at all the spurious overlap errors reported in G4 between disparate objects that are clearly not overlapping and nowhere near to each other.  In addition, the TGeo overlap checks also still fail with tons of bizzare messages. It was hoped that the simplified geometry would address this, but that does not seem to be the case, although the simplified geometry should make the debugging easier.  This would seem to indicate that the issue does not lie with the description of the solids themselves but with the internal handling in dd4hep/TGeo/G4.  

Holes


wdconinc (approved):
Thanks. Too bad this did not resolve the overlaps.

Issue #288, opened 2022-11-10
Edit LowQ2 detector and Vacuum geometries


simonge:
**Is your feature request related to a problem? Please describe.**
The current implementation doesn't allow a simple export of the beampipe and detectors into a cad format. Additionally nesting of detectors from separate xml files and code goes against the conventions and causes extra problems so the vacuum and taggers should be combined into a single detector description.

PR #290, opened 2022-11-10
288 edit lowq2 detector and vacuum geometries

simonge:
An updated Low Q2 tagger implementation which is described using single xml/cpp files and can be exported into CAD

- [x] Bug fix (issue #288 )
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

Changed the detector ID for taggers to a single value


simonge:
Are these failing due to overlaps unrelated to the PR?

wdconinc:
> Are these failing due to overlaps unrelated to the PR?

 There are still (incorrect) reports of overlaps in the barrel hcal. Those are not marked as "required", so as long as the required check are fine, you should be good.

simonge:
> > Are these failing due to overlaps unrelated to the PR?
> 
> There are still (incorrect) reports of overlaps in the barrel hcal. Those are not marked as "required", so as long as the required check are fine, you should be good.

Thanks, I'll stop hunting for problems I'm causing then!

wdconinc (approved):
This looks good. No comments.

PR #291, opened 2022-11-11
epic_SciGlassCalorimeter -> epic_EcalBarrelSciGlass, BarrelSciGlassCa…

veprbl:
…lorimeter_geo.cpp -> BarrelCalorimeterSciGlass_geo.cpp

Rename to use same conventions as Interlayers

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

Yes. They might need to edit their private compact files

No


src/BarrelCalorimeterSciGlass_geo.cpp
```diff
@@ -410,4 +410,4 @@ static Ref_t create_detector(Detector &lcdd, xml_h handle,
   return det_element;
 }
 
-DECLARE_DETELEMENT(epic_SciGlassCalorimeter, create_detector)
+DECLARE_DETELEMENT(epic_EcalBarrelSciGlass, create_detector)
```
wdconinc, line 413: We've been trying, to first order, to keep the name of the source file the same as the name of the element. Would we be able to use EcalBarrelSciGlass_geo.C?

veprbl: Yes, I saw the inconsistency, but the naming convention for other files is not like that yet, so I opted to conform to that first.

veprbl: Can you approve please.

wdconinc: We should keep the name of the plugin source file consistent with the definition. Here you are moving from consistent to inconsistent names.

wdconinc: Just to be clear, I like the rename of the plugin type and DECLARE_DETELEMENT, as it is clearer, but just would prefer the name of the src file to be `EcalBarrelSciGlass_geo.cpp` to be consistent with the others.

veprbl: No, we shouldn't. The names of the files are not part of the visible API. Renaming all calorimeters is beyond the scope of this PR.

veprbl: > Just to be clear, I like the rename of the plugin type and DECLARE_DETELEMENT, as it is clearer, but just would prefer the name of the src file to be `EcalBarrelSciGlass_geo.cpp` to be consistent with the others.

It's pretty consistent with others:
https://github.com/eic/epic/blob/15b2e4485ec8eafee6084889ca1d6eac20f47937/src/BarrelCalorimeterInterlayers_geo.cpp#L452

wdconinc: Ok, let's make them consistent in a future PR then.

wdconinc (approved):

PR #293, opened 2022-11-11
SciGlassCalorimeter: place towers in a sector volume

veprbl:
This places towers inside a sector volume instead of placing them directly in the envelope.

### TODO
 - [x] See if there is a performance improvement
 - [ ] ~~(optional) Avoid doubling the number of longitudinal carbon fibers~~


veprbl:
This does help with performance:
Canvas_1

wdconinc:
Curious: does this show up as an improvement in real-time running?

veprbl:
Good point, let me check real quick

wdconinc (approved):
This looks great! Great to see the improvement in the metrics.

veprbl:
Before:
```
real	8m41.722s
user	8m34.180s
sys	0m3.974s
```
After:
```
real	6m23.088s
user	6m14.730s
sys	0m3.818s
```

PR #188, opened 2022-10-08
SciGlassCalorimeter: revert back to sector-row-tower readout

veprbl:
The sector-module-slice was temporarily introduced due to a failing benchmark.

### What kind of change does this PR introduce?
- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [ ] Changes have been communicated to collaborators

### Does this PR introduce breaking changes? What changes might users need to make to their code?
Yes

### Does this PR change default behavior?
Yes

PR #296, opened 2022-11-12
SciGlassCalorimeter: specify tower material in the compact file

veprbl:
Introduces an option to specify the tower material.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No

### DIFF
--- a/compact/ecal/barrel_sciglass.xml
+++ b/compact/ecal/barrel_sciglass.xml
@@ -44,6 +44,11 @@
                   zmax="EcalBarrelForward_zmax" zmin="-EcalBarrelBackward_zmax"
                   etamin="EcalBarrel_etamin" etamax="EcalBarrel_etamax"/>
 
+      <documentation>
+        Pick the material for the towers
+      </documentation>
+      <material name="SciGlass" />
+
       <comment>
 	The current implementation of the wedge box approximates flat inner
 	surfaces of individual wedge boxes with a single shared tube at a fixed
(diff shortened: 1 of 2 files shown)


compact/ecal/barrel_sciglass.xml
```diff
@@ -44,6 +44,11 @@
                   zmax="EcalBarrelForward_zmax" zmin="-EcalBarrelBackward_zmax"
                   etamin="EcalBarrel_etamin" etamax="EcalBarrel_etamax"/>
 
+      <documentation>
+        Pick the material for the towers
+      </documentation>
+      <material name="SciGlass" />
```
wdconinc, line 50: A more typical approach is to add `material` as an attribute to the structure that defines the object that should receive that material. Without that, this sort of stands on its own without context so it may be unclear what the material exactly refers to.

veprbl: Yeah, thought about this. The other option is to put the material as an attribute into each family, or add a `<towers />` tag.

wdconinc (approved):
I'll approve this and leave it up to you as to whether you want to define materials for individual parts of the geometry.

veprbl:
I don't think we'd want or need to define materials for each tower family.

Another question would be to see how the material definitions could be kept more uniform across all the other calorimeters. I'd move forward with this option, as it's quite niche one, we can always change things later.

wdconinc:
I'm all for uniformity in material definitions! 

I'll just force-merge this rather than wait for another pipeline sequence after #188 was merged.

PR #301, opened 2022-11-14
fix: use consistent dd4hep cmake options

wdconinc:
The recent upgrade of some underlying dependency (not sure which one...) has surfaced an issue in how we use the DD4hep cmake helper macros. This has resulted in cases with duplicate library entries in `ldd` where only one is resolved, e.g.
```
        libTree.so.6.26 => not found
        libRIO.so.6.26 => not found
        libTree.so.6.26 => /opt/software/linux-debian-x86_64/gcc-12.2.0/root-6.26.06-vjvesxyjadhyrz3darz6wlyqhz4visgv/lib/root/libTree.so.6.26 (0x00007f5884aa7000)
        libRIO.so.6.26 => /opt/software/linux-debian-x86_64/gcc-12.2.0/root-6.26.06-vjvesxyjadhyrz3darz6wlyqhz4visgv/lib/root/libRIO.so.6.26 (0x00007f5884190000)
```
This PR fixes the macro calling by properly including the correct DD4hep cmake files and calling the required macros.

- [x] Bug fix (issue: build failures on updated containers)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,10 +47,15 @@ list(TRANSFORM DD4hep_required_libraries PREPEND DD4hep::)
 
 # Dependencies
 find_package(DD4hep REQUIRED COMPONENTS ${DD4hep_required_components})
+include(${DD4hep_DIR}/cmake/DD4hep.cmake)
+include(${DD4hep_DIR}/cmake/DD4hepBuild.cmake)
+dd4hep_configure_output()
+dd4hep_set_compiler_flags()
 if(${DD4hep_VERSION} VERSION_LESS 1.21)
   message(WARNING "DD4hep before 1.21 does not write collections correctly \n"
                   "More info at https://github.com/AIDASoft/DD4hep/pull/922")
 endif()
+
 find_package(ActsDD4hep)
 if(ActsDD4hep_FOUND)
   add_compile_definitions(USE_ACTSDD4HEP)
@@ -59,21 +64,16 @@ else()
   find_package(Acts REQUIRED COMPONENTS Core PluginIdentification PluginTGeo PluginDD4hep)
   set(ActsDD4hep ActsCore ActsPluginDD4hep)
 endif()
+
 find_package(fmt REQUIRED)
 
 #-----------------------------------------------------------------------------------
-set(a_lib_name ${PROJECT_NAME})
-
-dd4hep_configure_output()
 
 file(GLOB sources CONFIGURE_DEPENDS src/*.cpp)
-dd4hep_add_plugin(${a_lib_name}
+dd4hep_add_plugin(${PROJECT_NAME}
   SOURCES ${sources}
-  USES ${ActsDD4hep} ROOT::Core ROOT::Gdml
-  )
-target_link_libraries(${a_lib_name}
-  PUBLIC ${DD4hep_required_libraries} fmt::fmt
-  )
+  USES ${ActsDD4hep} ${DD4hep_required_libraries} ROOT::Core ROOT::Gdml fmt::fmt
+)
 
 #-----------------------------------------------------------------------------------
 # Parse jinja templates: once by default, and once for all yml files
--- a/src/FileLoaderHelper.h
+++ b/src/FileLoaderHelper.h
@@ -15,12 +15,13 @@
 
 namespace fs = std::filesystem;
 
-using namespace dd4hep;
-
 // Function to download files
 inline void EnsureFileFromURLExists(std::string url, std::string file, std::string cache_str = "",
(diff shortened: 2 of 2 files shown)


CMakeLists.txt
```diff
@@ -47,10 +47,15 @@ list(TRANSFORM DD4hep_required_libraries PREPEND DD4hep::)
 
 # Dependencies
 find_package(DD4hep REQUIRED COMPONENTS ${DD4hep_required_components})
+include(${DD4hep_DIR}/cmake/DD4hep.cmake)
+include(${DD4hep_DIR}/cmake/DD4hepBuild.cmake)
+dd4hep_configure_output()
+dd4hep_set_compiler_flags()
```
veprbl, line 53: I've never seen anyone doing this before. Can we discuss this in terms of change to the compilation flags?

wdconinc: We can and should definitely discuss this (for one: I have no mac to test this on).

But this is a bit of a hotfix situation. At this point the builds of the geometry stopped functioning in the container, with no clear reason as to why (can't reproduce locally, but can reproduce in docker containers), see e.g. https://github.com/eic/epic/actions/runs/3466153267/jobs/5789693825 (i.e. the current HEAD on this PR, so this PR doesn't even fix it). The symptom is that the ROOT libraries are not found by the `libepic.so` anymore, so the build completes but the geometry checks fail.

This identical issue first showed up in NPDet (where we don't use the geometry plugins, but where it caused the build to fail since we create executables that link against the problematic libraries). There, the build system was a bit more of a mess (with more homegrown than DD4hep-based cmake stuff), and the issue was resolved by aligning the cmake calls to the approach that DD4hep examples take. No such luck here.

veprbl:
```
WARNING: cannot load libepic.so for factory epic_Solenoid
[92](https://github.com/eic/epic/actions/runs/3466153267/jobs/5789693825#step:5:94)
WARNING: libGdml.so.6.26: cannot open shared object file: No such file or directory
```
Didn't you recently enable linking with Gdml for ip6?

veprbl: Not finding libGdml may be because it's missing in installation of ROOT, if all other libraries are found.

wdconinc: This has worked fine after the IP6 merge and libGdml.so is installed. I think this is one of those cases where something used to work, but coincidentally. ROOT installs to a prefix and is linked into /usr/local (as are all other dependencies). We used to have the ROOT libraries in LD_LIBRARY_PATH, but only because podio adds its prefix to LD_LIBRARY_PATH. The default for installing ROOT into GNU install dirs was changed and is now enabled, so the ROOT libraries end up in /usr/local/lib/root, which is not in LD_LIBRARY_PATH anymore. I think one of the epic dependencies and ROOT dependents is not using rpath linking (as we try to encourage) and therefore we were always accidentally picking up the LD_LIBRARY_PATH with the ROOT libraries but aren't anymore. The solution is likely simply adding an entry to the LD_LIBRARY_PATH and searching offline for what is failing to use rpath linking.

wdconinc: Essentially, something is linking against ROOT without using the `root-config --libs` flags (which would enforce rpath linking), `-Wl,-rpath,/opt/software/linux-debian-x86_64/gcc-12.2.0/root-6.26.06-vjvesxyjadhyrz3darz6wlyqhz4visgv/lib/root`. So we end up having to resort to LD_LIBRARY_PATH shenanigans.

veprbl: Usually, ROOT folks don't use rpath, but rely on thisroot.sh to set the LD_LIBRARY_PATH.

veprbl: I wonder if LD_DEBUG can shed some light on this: https://github.com/eic/epic/pull/304

wdconinc: Yes, they like their non-standard approaches...

wdconinc: At least I don't have many other software packages on my system that require me to call `this<program>.sh` if I want to use it.

wdconinc: > I wonder if LD_DEBUG can shed some light on this: https://github.com/eic/epic/pull/304

I am pretty sure this is resolved by simply setting LD_LIBRARY_PATH. I am not sure yet who exactly drops the ball on supporting rpath (by now my money is on DD4hep in its forwarding of `ROOT::` targets), but that's likely not something we'll be able to fix in this repo anyway.

veprbl: Okay. It would seem like you've resolved this elsewhere.

Issue #302, opened 2022-11-15
Field map files disappear


veprbl:
> trying https://wiki.bnl.gov/EPIC/images/3/3a/MARCO_v.6.4.1.1.2_1.7T_Magnetic_Field_Map_2022_10_05_rad_coords_cm_T.txt
       >   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
       >                                  Dload  Upload   Total   Spent    Left  Speed
       >   0   287    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
       > curl: (22) The requested URL returned error: 404
       > error: cannot download MARCO_v.6.4.1.1.2_1.7T_Magnetic_Field_Map_2022_10_05_rad_coords_cm_T.txt from any mirror

This is bad for reproducibility.

Makes me wonder if we should store those in a separate repo, with git-annex or LFS.

PR #303, opened 2022-11-15
solenoid.xml: use magentic field versioned in eic/epic-data

veprbl:
Fixes: #302


wdconinc (approved):
We should have done this from the start...

PR #312, opened 2022-11-18
fix: clarify error on failing download or link; fail when empty file downloaded

wdconinc:
Improve error messages when FileLoader encounters issues.

- [X] Bug fix (issue: many questions on mattermost...)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


sly2j (approved):

src/FileLoaderHelper.h
```diff
@@ -151,4 +157,12 @@ inline void EnsureFileFromURLExists(std::string url, std::string file, std::stri
     printout(ERROR, "FileLoader", "check permissions and retry");
     std::_Exit(EXIT_FAILURE);
   }
+
+  // final check of the file size
+  if (fs::file_size(file_path) == 0) {
```
veprbl, line 162: I see, in https://en.cppreference.com/w/cpp/filesystem/file_size they say

> The result of attempting to determine the size of a directory (as well as any other file that is not a regular file or a symlink) is implementation-defined.

I'm not super familiar with std::filesystem, but perhaps it would be tiny bit safer to have:
```suggestion
  if (fs::file_size(fs::canonical(file_path)) == 0) {
```

veprbl: Oh, misread it. I guess it says that symlinks are fine.

veprbl: > symlinks are followed

Yeah

src/FileLoaderHelper.h
```diff
@@ -19,7 +19,7 @@ using namespace dd4hep;
 
 // Function to download files
 inline void EnsureFileFromURLExists(std::string url, std::string file, std::string cache_str = "",
-                                    std::string cmd = "curl --retry 5 --location --fail {0} --output {1}")
+                                    std::string cmd = "curl --silent --retry 5 --location --fail {0} --output {1}")
```
veprbl, line 22:
```suggestion
                                    std::string cmd = "curl --retry 5 --location --fail {0} --output {1}")
```

veprbl (approved):
Looks good. I would not silence curl though. It's useful to know that it is being ran (imagine user unknowingly running downloads inside the jobs), or what was the error.

src/FileLoaderHelper.h
wdconinc: Yeah, I agree. I wish there was a way to get rid of the progress bar which renders poorly and only get the final line.

veprbl (approved):

wdconinc:

Issue #309, opened 2022-11-17
Misplacement of the ZDC (ecce-style)


shimasnd:
### Environment: (where does this bug occur, have you tried other environments)
- Which branch (`main` for latest released): main 
- Which revision (`HEAD` for the most recent): HEAD
- Any specific OS or system where the issue occurs? no
- Any special versions of ROOT or Geant4? no

### Steps to reproduce: (give a step by step account of how to trigger the bug)
1. Use ZDC.xml instead of ZDC_Athena.xml

### Actual Result: (what do you get when you execute the steps above)
ZDC is not on a straight line.

PR #316, opened 2022-11-22
Modification of ZDC (ecce-style) placement. 

shimasnd:
Modification of the ecce-style ZDC placement. 
NOTE: it will not affect the current default ZDC behavior (Athena style).
Misplacement is fixed. 
ZDC is shifted to the place in the CAD figure. 

- [x] Bug fix (issue #309)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No. Only if one uses ZDC.xml instead of ZDC_Athena.xml, the position of the ZDC will be modified.


compact/far_forward/ZDC.xml
```diff
@@ -25,47 +24,52 @@
       Zero Degree Calorimeter - ECAL
       -------------------------------
     </comment>
-    <constant name="ZDC_1stSilicon_z_pos"        value="ZDC_z_pos"/>
-    <constant name="ZDC_1stSilicon_x_pos"        value="ZDC_1stSilicon_z_pos * ionCrossingAngle"/>
+    <constant name="ZDC_1stSilicon_r_pos"        value="ZDC_r_pos+0.5*cm"/>
+    <constant name="ZDC_1stSilicon_z_pos"        value="ZDC_1stSilicon_r_pos * cos(ionCrossingAngle)"/>
+    <constant name="ZDC_1stSilicon_x_pos"        value=" -1. * ZDC_1stSilicon_r_pos * sin(ionCrossingAngle)"/>
```
wdconinc, line 29: I'm surprised we need an explicit `-1` here, since I thought the so-called "big sign flip" made sure that `ionCrossingAngle` contained the correct sign to be able to treat this without additional factors of `-1` in the far forward region. @ajentsch @sly2j

shimasnd: Thanks for pointing this out and sorry that I messed up. I will upload a new xml file today.

compact/far_forward/ZDC.xml
```diff
@@ -1,6 +1,3 @@
-<!-- SPDX-License-Identifier: LGPL-3.0-or-later -->
```
wdconinc, line 1: Don't remove the copyright headers.

compact/far_forward/ZDC.xml
```diff
@@ -11,11 +8,10 @@
       Zero Degree Calorimeter General Parameters
       -------------------------------------------
     </comment>
-    <constant name="ZDC_z_pos"         value="3560.0 * cm"/> <!-- need to shift FORWARD toward IP to account for 50cm shift -->
-    <constant name="ZDC_x_pos"         value="ZDC_z_pos * ionCrossingAngle"/>
+    <constant name="ZDC_r_pos"         value="3550.0 * cm"/> <!-- Make it match to the CAD figure 22/Nov/22 -->
     <constant name="ZDC_y_pos"         value="0.0 * cm"/>
     <constant name="ZDC_rotateX_angle" value="0.0 * rad"/>
-    <constant name="ZDC_rotateY_angle" value="ionCrossingAngle"/>
+    <constant name="ZDC_rotateY_angle" value="-1.* ionCrossingAngle"/>
```
wdconinc, line 14: This doesn't really address the issue I pointed out. It just moves the minus sign somewhere else. Why is a minus sign needed at all?

ajentsch:
Isn't this just because of the coordinate system? In the z-x plane, the
clockwise rotation is -25 mrad.

compact/far_forward/ZDC.xml
shimasnd: Due to historical reason, local rotation of volumes in the detector templates (cpp files) under src/ is defined as RotationZYX(rot.z(), -rot.y(), rot.x()) in all the ZDC related cpp files (including the ATHENA ones). Current -1 factor for the rotation in the ZDC.xml file gives proper alignment. 
I will check the situation for ATHENA ZDC as it could have wrong alignment as well. 
I and/or Po-Ju (@pjlin) will reverse the rotation in cpp files of both ECCE and ATHENA type and take out -1 factors from XML file during this week.

compact/far_forward/ZDC.xml
pjlin: The copyright headers have been put back into the ZDC.xml.

ajentsch (approved):

PR #317, opened 2022-11-22
Lfhcal incusion

FriederikeBock:
This PR introduces the detailed geometry of the LFHCal modules and their correct placement.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __


src/LFHCAL_geo.cpp
```diff
@@ -0,0 +1,915 @@
...
+  }
+
+  // create 4M modules
+  if (xpos4M.size() != ypos4M.size() || xpos4M.size() != zpos4M.size()){
+    std::cout << xpos4M.size() << "\t" <<ypos4M.size() <<  "\t" << zpos4M.size() << std::endl;
+    std::cout <<  "idiot you can't count" << std::endl;
+  } else {
+    for (int f = 0; f < (int)xpos4M.size(); f++){
+      std::cout <<  "LFHCAL placing 4M module: " << f << "/"<< (int)xpos4M.size() << "\t" << xpos4M[f] << "\t" << ypos4M[f] << "\t" << zpos4M[f]<< std::endl;
+      Assembly  fourMassembly = createFourMModule ( desc, fourM_params, slice_Params,  moduleID, length, sens);
```
veprbl, line 903: If these assemblies are identical, you can create their logical volumes just once and use `placeVolume` to put copies of it in different locations as needed. Same goes for all the nested pieces down the hierarchy.

wdconinc (commented):
I think the reason why you are seeing too many volume is because there is no layering of volume hierarchy due to the assembly volumes.

An assembly is essentially just a bookkeeping device that groups some volumes but without actually introducing a new node in the geometry tree. The grouped volumes are still all placed at the same level in the geometry tree. So, if you place assemblies of assemblies of assemblies it can multiply the number of volumes at that level very rapidly.

Many volumes at the same level of the geometry tree (in particular for materials where there are showers) is very detrimental to the performance. Anytime any particle leaves a volume in that assembly, the geant4 navigator will check if the track is entering a new volume by looping over all 'sibling' volumes. If there are 1 million sibling volumes, and if this has to be done for each particle in the shower, that takes a substantial amount of time. I have added some comments to specific places where you could define a volume instead of using an assembly.

The other comment I would have is that we don't need to reconstruct the geometry in the same way we create the physical detector. It is perfectly acceptable in simulation-land to define a volume completely made up of steel and then place scintillators inside of that volume. This again cuts down the number of volumes by a factor of 2, typically, and makes the navigator faster by another factor 2 because we don't need to go out of one volume, into the parent, then immediately inside another volume at every transition.

src/LFHCAL_geo.cpp
FriederikeBock: I tried that, but unfortunately that didn't work as I needed to assign each active volume the correct IDs

veprbl: I'm submitting a similar change for EcalBarrel right now. Here is how it's done:
https://github.com/eic/epic/pull/293/commits/8cedd81efcfedff3d18c3abc90a1d9a93d490269#diff-3eed1d7a563d681a4ff46f655d092d270bb844260f59462fa5aecdbdc845a904L282
https://github.com/eic/epic/pull/293/commits/8cedd81efcfedff3d18c3abc90a1d9a93d490269#diff-3eed1d7a563d681a4ff46f655d092d270bb844260f59462fa5aecdbdc845a904R434

wdconinc: https://github.com/eic/epic/pull/293/files?diff=unified&w=1#diff-3eed1d7a563d681a4ff46f655d092d270bb844260f59462fa5aecdbdc845a904 (this should hide the whitespace changes)

compact/forwardHcal_standalone.xml
```diff
@@ -0,0 +1,12 @@
...
+  <display>
+  </display>
+  <define>
+  </define>
+
+  <documentation level="10">
+    ### Hcal configuration
+  </documentation>
+  <include ref="hcal/hcal_longforward.xml"/>
+</lccdd>
```
wdconinc, line 12: We should not include this file. The entry points should be handled through the configurations.

compact/hcal/lfhcal.xml
```diff
@@ -0,0 +1,179 @@
+<lccdd>
+
+  <info name="lfhcal.xml"
+        title="ECCE LFHCAL with insert cutout"
```
wdconinc, line 4: I assume you mean EPIC here.

src/LFHCAL_geo.cpp
FriederikeBock: > I'm submitting a similar change for EcalBarrel right now. Here is how it's done: [8cedd81#diff-3eed1d7a563d681a4ff46f655d092d270bb844260f59462fa5aecdbdc845a904L282](https://github.com/eic/epic/commit/8cedd81efcfedff3d18c3abc90a1d9a93d490269#diff-3eed1d7a563d681a4ff46f655d092d270bb844260f59462fa5aecdbdc845a904L282) [8cedd81#diff-3eed1d7a563d681a4ff46f655d092d270bb844260f59462fa5aecdbdc845a904R434](https://github.com/eic/epic/commit/8cedd81efcfedff3d18c3abc90a1d9a93d490269#diff-3eed1d7a563d681a4ff46f655d092d270bb844260f59462fa5aecdbdc845a904R434)

That would work if one of my modules would be the equivalent of a readout unit, unfortunately this is not the case.  One 8M modules has right now 8 * 65 readout channels.

compact/hcal/lfhcal_wInsert.xml
```diff
@@ -0,0 +1,179 @@
+<lccdd>
+
+  <info name="lfhcal.xml"
+        title="ECCE LFHCAL with insert cutout"
+        author="@FriederikeBock"
+        url="https://github.com/FriederikeBock"
+        status="development"
+        version="1.0"
+  ><comment/></info>
```
wdconinc, line 9: We should avoid having two parallel implementations that only differ in the presence (or not) of the insert. It is likely to lead to headaches as parameters are changed in one and not the other.

compact/materials.xml
```diff
@@ -334,6 +334,25 @@
     <fraction n="0.528" ref="Quartz"/>
     <fraction n="0.472" ref="Fr4_Epoxy"/>
   </material>
+  <material name="Tyvek">
+    <D type="density" value="0.3529" unit="g/cm3"/>
+    <fraction n="0.528" ref="Quartz"/>
+    <fraction n="0.472" ref="Fr4_Epoxy"/>
+  </material>
```
wdconinc, line 341: Is tyvek really a mixture of quartz and epoxy???

templates/epic.xml.jinja2
```diff
@@ -139,14 +139,36 @@
...
+
+    ## Central H calorimetry
+  {% if features is not defined or features['hcal'] is none or 'barrel_gdml' in features['hcal'] %}
+  <include ref="${DETECTOR_PATH}/compact/hcal/barrel_gdml.xml"/>
+  {% endif -%}
 
-  {% for hcal in features['hcal'] -%}
-  <include ref="${DETECTOR_PATH}/compact/hcal/{{ hcal }}.xml"/>
-  {% endfor -%}
 {% endif -%}
```
wdconinc, line 172: Please don't change this file. The relevant files for the forward hcal should be included by specifying their path and fil name in the configuration.

src/LFHCAL_geo.cpp
```diff
@@ -0,0 +1,915 @@
...
+  pvm = modAbsAssembly.placeVolume(nodge_t3_vol, Position(0, h_mod/2-t_mod_tp-0.5*h_nodge, 0 ));
+  pvm = modAbsAssembly.placeVolume(nodge_t4_vol, Position(-(w_mod-2*t_mod_sp)/4, h_mod/2-t_mod_tp-0.5*h_nodge, 0 ));
+  pvm = modAbsAssembly.placeVolume(nodge_t5_vol, Position(-w_mod/2+t_mod_sp+0.5*w_nodgeC, h_mod/2-t_mod_tp-0.5*h_nodge, 0 ));
+
+  // bottom row of nodges
+  pvm = modAbsAssembly.placeVolume(nodge_b1_vol, Position(w_mod/2-t_mod_sp-0.5*w_nodgeC, -(h_mod-2*t_mod_tp)/2+0.5*h_nodge, 0 ));
+  pvm = modAbsAssembly.placeVolume(nodge_b2_vol, Position((w_mod-2*t_mod_sp)/4, -(h_mod-2*t_mod_tp)/2+0.5*h_nodge, 0 ));
+  pvm = modAbsAssembly.placeVolume(nodge_b3_vol, Position(0, -(h_mod-2*t_mod_tp)/2 +0.5*h_nodge, 0 ));
+  pvm = modAbsAssembly.placeVolume(nodge_b4_vol, Position(-(w_mod-2*t_mod_sp)/4, -(h_mod-2*t_mod_tp)/2 +0.5*h_nodge, 0 ));
+  pvm = modAbsAssembly.placeVolume(nodge_b5_vol, Position(-w_mod/2+t_mod_sp+0.5*w_nodgeC, -(h_mod-2*t_mod_tp)/2 +0.5*h_nodge, 0 ));
```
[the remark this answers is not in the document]
FriederikeBock: So the ExtrudedPolygon worked. I think if we wanna be nice we help the dd4hep community with an example though, as the code isn't selfexplanatory.

configurations/lfhcal_withInlay.yml
```diff
@@ -0,0 +1,4 @@
+features:
+  hcal:
+    lfhcal_wInsert:
+
```
wdconinc, line 4: Rename file to have `_only` suffix since it includes only a single detector.

src/LFHCAL_geo.cpp
```diff
@@ -0,0 +1,915 @@
...
+//************************** Assembly for absorber plates for 8M modules *************************************
+//************************************************************************************************************
+Assembly createAbsorberPlateEightM(Detector& desc,
+                                   std::string basename,
+                                   double h_mod,
+                                   double w_mod,
+                                   double t_mod_tp,
+                                   double t_mod_sp,
+                                   double t_slice,
+                                   double h_nodge,
```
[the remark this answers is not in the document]
FriederikeBock: yes fixed.

configurations/forwardNew.yml
```diff
@@ -0,0 +1,18 @@
...
+    tof_barrel:
+    tof_endcap:
+  pid:
+    dirc:
+    drich:
+  ecal:
+    forward_homogeneous:
+    forward_insert_homogeneous:
+  hcal:
+    lfhcal:
```
wdconinc, line 18: I don't think forwardNew is a clear name (or even correct as this includes also barrel detectors). Can we rename this to `full_hcal_only_forward` or so?

compact/materials.xml
wdconinc, line 341: Is tyvek really a mixture of quartz and epoxy???

FriederikeBock: Tyvek is very similar in its composition to a PCB just lower density (that's why I went with that.)

compact/hcal/lfhcal_wInsert.xml
FriederikeBock: the problem is the positioning of the modules can be different and not easily calculable. That's why having them separate makes sense.

compact/hcal/lfhcal.xml
FriederikeBock: fixed in next commit.

compact/materials.xml
wdconinc: Tyvek is not similar in composition to PCBs and it for sure is not > 50% quartz (or glass); there is no Si at all in tyvek. Please use the correct composition. When you make Tyvek available as a material, others may use it assuming it is at least somewhat correct.

src/LFHCAL_geo.cpp
veprbl: It is possible to further define values sub-segmentation for sub volumes. This is done in many other detectors, including SciGlass linked above.

src/LFHCAL_geo.cpp
```diff
@@ -0,0 +1,915 @@
...
+                              int modID,
+                              double length,
+                              SensitiveDetector sens
+){
+
+  std::string baseName = "LFHCAL_4M"+_toString(modID, "_%d");
+
+  // assembly definition
+  Assembly moduleAssembly(baseName);
+  moduleAssembly.setVisAttributes(mod_params.mod_visStr);
```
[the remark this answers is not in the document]
FriederikeBock: done

veprbl:
@FriederikeBock I've pushed a fix for the issue you've brought up on Mattermost. This should unblock addressing rest of the comments.

wdconinc:
Since this was started as a branch on a fork, we should first merge this into a branch inside the upstream repository so we can then run CI checks before we can merge into main. I'll change the base here since we cannot change the branch itself without losing the discussion history.

FriederikeBock:
> Since this was started as a branch on a fork, we should first merge this into a branch inside the upstream repository so we can then run CI checks before we can merge into main. I'll change the base here since we cannot change the branch itself without losing the discussion history.

sure no problem!

src/LFHCAL_geo.cpp
wdconinc: @veprbl Can you resolve this conversation if this looks good to you?

compact/materials.xml
```diff
@@ -336,6 +336,25 @@
...
+
+  <material name="Ti02Epoxy">
+    <D type="density" value="2.3" unit="g/cm3"/>
+    <fraction n="0.67" ref="Epoxy"/>
+    <fraction n="0.33" ref="TiO2"/>
+  </material>
+
+
+
```
wdconinc, line 357:
```suggestion
```

templates/epic.xml.jinja2
```diff
@@ -139,11 +139,11 @@
   {% endfor -%}
 {% endif -%}
 
-{% if 'hcal' in features %}
+
+{% if features is not defined or 'hcal' in features %}
```
wdconinc, line 143:
```suggestion
{% if 'hcal' in features %}
```

templates/epic.xml.jinja2
```diff
@@ -139,11 +139,11 @@
...
   <documentation level="10">
-    ## hadronic calorimetry
+    ## Forward H calorimetry
```
wdconinc, line 145:
```suggestion
    ## Hadronic calorimetry
```
This includes more than forward calorimetry.

wdconinc:
@FriederikeBock Can you address the missing tests?
- Add SPDX license headers to the source code
- Address the inlay geometry failure, https://github.com/eic/epic/actions/runs/4066563797/jobs/7002918367

src/LFHCAL_geo.cpp
veprbl: No, this has not been implemented. Not addressing this will result in higher memory use (and slower simulation?). If that's okay, we could merge without it.

veprbl: Nevermind. See new commits. Looks better, resolved.

wdconinc: Could we assess the time it takes to simulate per event, before and after, and the memory it takes after?

compact/definitions.xml
```diff
@@ -197,6 +197,7 @@ The unused IDs below are saved for future use.
     <constant name="HCalEndcapN_ID"     value="113"/>
     <constant name="PassiveSteelRingEndcapP_ID" value="114"/>
     <constant name="HCalEndcapPInsert_ID" value="115"/>
+    <constant name="LFHCAL_ID"     value="116"/>
```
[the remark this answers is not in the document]
nschmidtALICE: Hi Wouter, could you please provide some reasoning for this proposed change?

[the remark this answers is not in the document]
wdconinc: The LFHCAL is the forward endcap hcal. There is an ID already allocated for it: `HCalEndcapP_ID`. We should use that ID. We only have a finite number of IDs (255).

configurations/lfhcal_withInlay.yml
```diff
@@ -0,0 +1,4 @@
+features:
+  hcal:
+    lfhcal_insertGeom:
```
veprbl, line 3: Looks like we are missing `compact/hcal/lfhcal_insertGeom.xml`

FriederikeBock:
Will pull from different branch.

Issue #328, opened 2022-12-03
dRICH and pfRICH: use `VariantParameters` plugin to streamline geometry access from reconstruction


c-dilks:
**Is your feature request related to a problem? Please describe.**
Sometimes it is difficult to access geometry parameters from the main `TGeo` handle

**Describe the solution you'd like**
Use https://github.com/AIDASoft/DD4hep/pull/911 to assign `map`s to any `DetElement`. For example, the normal vectors of the sensor surfaces could be determined in the `src/.cpp` file, and stored in a `map`.

c-dilks:
Let's just do this for the sensor positions and normals, which can be tricky to extract after all the `Transform3D`s. The other constants, such as mirror position and radius, can remain as DD4hep constants, since those are convenient to have in `npdet_info` output.

Issue #175, opened 2022-10-06
dRICH: add sensor services


c-dilks:
**Is your feature request related to a problem? Please describe.**
Request from Roberto Preghenella:

> I would like to ask if there is any chance to try putting some services also behind the dRICH sensor plane.
> somewhat in the same fashion as done for the pfRICH.
>
> my feeling from your drawings is that there is not enough room to put services and electronics behind the dRICH sensors. it is true that a real model for that is still missing, but a possible start is what was put in the drawings for the mRICH. Namely a 20 cm box behind the sensors that represents the current-guess for space needed for services and electronics.
>
> would it be possible to have a look at that?
> I have a drawing that I attach to this email to explain better my message.
my rough estimate based on numbers extracted from this drawing (need to see how it is in 3D) is that there is a lack of 10 cm in space with the current optics layout. 

![Image-DR4HT1](https://user-images.githubusercontent.com/7843751/194400649-05f318ee-fb2c-44d4-a441-3e1779f681c9.png)

preghenella:
hi @c-dilks thanks for opening this issue.
can one not increase the thickness of the sensor volume to achieve what I am asking in a quick-and-dirty way?

c-dilks:
If we try to extrude the sensor bases, they might start overlapping each other. I think an easier approach is to add a `Sphere`, placed with the same center as the sphere used to generate the sensor positions. We can make the inner and outer radii such that it's 20 cm thick, and the outer radius can be made to be as close to the sensor bases as possible.

Then we can take the boolean `IntersectionSolid` of this `Sphere` and the gas volume, cutting the shape so it fits in the vessel. Since we will have to move the sensors forward by ~10 cm, the optics will have to be re-tuned, but that will be straightforward.

What material(s) should we use? We could also make concentric `Sphere`s, if we want layers of materials. By comparison, the pfRICH currently uses 3 sheets of aluminum, 0.5 cm thick, equally spaced spanning 15 cm behind the sensor plane, with air volumes between.

preghenella:
as a start I would not mind of overlaps, but just try to understand how much the readout could violate the boundaries of the dRICH envelope. that is why I was suggesting to increase the dz of the sensor box.
a possible readout block might have something like a 4x4 cm² of area behing the 5x5 cm² of the sensors, similar to this pic
![image](https://user-images.githubusercontent.com/5835025/194618285-3e752cc9-6a44-4265-9d1f-dfa8ee6fdd98.png)

[...]

Issue #111, opened 2022-09-13
dRICH: use true sensor thickness and add resin base


c-dilks:
**Is your feature request related to a problem? Please describe.**
The sensor should be made more realistic.

**Describe the solution you'd like**
- Add resin base to dRICH sensors, following the [S13361-3050NE-08 datasheet](https://www.hamamatsu.com/content/dam/hamamatsu-photonics/sites/documents/99_SALES_LIBRARY/ssd/s13361-3050_series_kapd1054e.pdf)
- Use the correct sensor thickness

c-dilks:
@chchatte92 @alexander-kiselev @ajitkmaurya is this the right SiPM?

alexander-kiselev:
which one? I see no attachment

________________________________

c-dilks:
S13361-3050NE-08
https://www.hamamatsu.com/content/dam/hamamatsu-photonics/sites/documents/99_SALES_LIBRARY/ssd/s13361-3050_series_kapd1054e.pdf
I suppose we could also use S13361-3050AE-08 which AFAIK just has a different base pinout and form factor.
(BTW, if you are reading this as an email it's better to view the issue on Github at https://github.com/eic/epic/issues/111)

alexander-kiselev:
Hi Chris,

  historically I first used the specs of S13361-3050AE-08, but for our purposes AE and NE are the same type.
Roberto would know better what they are considering to use.

  Cheers,
    Alexander.

________________________________

JunhuaiXu:
structure of S13361-3050NE-08

Number of channels: 64 (8 × 8)

sensor thickness=real thickness + resin=1.35mm
sensor size: 25.8mm



```
<!-- sensor geometry; model = S13361-3050NE-08 SiPM -->
<constant name="DRICH_sensor_size"           value="25.8*mm"/> <!-- full length of the sensor side FIXME: resin base should be this size -->
<constant name="DRICH_pixel_gap"             value="0.2*mm"/> <!-- size of gaps between adjacent pixels AND gaps between edge pixels and sensor side -->
<constant name="DRICH_sensor_sensitive_size" value="DRICH_sensor_size - DRICH_pixel_gap"/> <!-- side length of photosensitive surface (to be segmented) -->
<constant name="DRICH_sensor_thickness"      value="1.35*mm"/>  <!-- sensor thickness FIXME: use the real thickness + resin -->
[...]

PR #331, opened 2022-12-05
feat(dRICH): assemble sensors in Photodetector Units (PDUs)

JunhuaiXu:

This is a big change for the dRICH sensor geometry. It was initially started by @JunhuaiXu and continued by @c-dilks with the advice of Marco Contalbrigo, et al. 

#### Summary of changes:
- adjust some thicknesses, such as vessel walls and acrylic filter
- refactor of the sensor geometry (which on `main` is just a thin `Box` for each SiPM)
  - add the resin backing, following [the S13361-3050NE-08 SiPM datasheet](https://www.hamamatsu.com/eu/en/product/optical-sensors/mppc/mppc_mppc-array/S13361-3050NE-08.html); the photosensitive volume plus the resin backing forms a "SiPM assembly"
  - make a 2x2 matrix of SiPM assemblies, forming a Photodetector Unit (PDU); include service materials behind each 2x2 matrix forming the full PDU assembly
    - `frontservices`, currently just a box
    - `boards`: 5 circuit boards, coming out of the `frontservices` box
    - `backservices`, currently just another box
    - all of these are made of Aluminum, for now
  - The spherical tiling algorithm now tiles PDU assemblies, instead of the original SiPM `Box`es
- Change the readout bit fields, so sensors are identified by a PDU number `pdu` and a SiPM number (0-3) `sipm`, instead of just a single `module` number
- Extrude the annular region of the vessel (the part outside the aerogel snout) toward the -z direction, to allow the sensor services to fit; this is a step toward modelling sensor boxes, which will house each sector's set of PDUs (outside the scope of this PR)

#### Pictures and Geometry details:
See `sensors.pdf` from <https://indico.bnl.gov/event/19848/>

---

- [ ] Bug fix (issue #__)
- [X] New feature, closing the following:
  - close https://github.com/eic/epic/issues/111
  - close https://github.com/eic/epic/issues/175
  - close https://github.com/eic/epic/issues/333
  - close https://github.com/eic/epic/issues/328
- [ ] Documentation update
- [ ] Other: __

---

---

**YES**, addressed by these PRs; all of them should be merged simultaneously:
- https://github.com/eic/EICrecon/pull/727
- https://github.com/eic/drich-dev/pull/83

#### Things to check downstream:
- [x] pixel gap cuts
- [x] event display and readout
- [x] are the photons stopped by the resin, or do they still penetrate through?
- [x] sensor thickness and IRT geometry conversion
- [x] number of photons / interaction of photons at the resin/sensor/air surface
- [x] PID

---

yes, photons will stop within the resin volume (currently on `main` they pass through and stop at the vessel wall behind the sensors); this should also prevent photons causing hits from behind (unlikely, but can happen) and from the sensor sides (because of the resin walls)


c-dilks:
Hmm... somehow this looks like it's based off an older commit on the `main` branch, but Github tells me it's indeed based off the most recent commit on `main`. In other words, it's not possible to update your branch with respect to `main` since Github thinks it is already as up-to-date as possible. Let me see if I can fix it with `git rebase` or something...

c-dilks:
Fixed! I created temporary branch `before-rebase` to backup what you had before the fix, in case anything got lost in the rebase.

c-dilks:
If I understand correctly, the full sensor thickness should be 1.35mm, all resin. The first 0.1mm of that is the sensitive part (currently we use `AirOptical` as the material, for technical reasons), so we should modify this to place the sensitive volume inside a resin volume, rather than two separate volumes.

With this change, we will have to check the number of photons carefully, so perhaps we delay merging this PR until then.

2022-12-05-122925_741x554_scrot

c-dilks:
Full schematic:
2023-02-28-122444_1276x623_scrot

c-dilks:
This is starting to look good, however, we need to check things carefully downstream (checklist added to PR description above).

Here is a picture (grey=resin, blue=sensor active area)
2023-02-28-151155_501x558_scrot

src/DRICH_geo.cpp
```diff
@@ -37,90 +37,105 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
 
   // attributes, from compact file =============================================
```
c-dilks, line 38: The majority of changes in this following block are `double` -> `auto` and formatting changes. There are some new variables though, for the PDU and sensor geometry.

c-dilks:
@chchatte92 have you tested this yet? Shall we open this PR for review?

wdconinc:
Quick questions: 
- Is there a way adjust visualization settings so the services behind the SiPMs are not shown?
- Will the same PDUs be used by pfRICH? Does it make sense to start extracting them as a separate source code unit (header + implementation) with the eye on future reuse?

chchatte92:
@wdconinc I guess the PDU will be different for pfRICH.

wdconinc:
> changing a true to a false somewhere

Yeah, usually a `showdaughters=false` is sufficient.

> Somewhat related: root-project/jsroot

Our geoviewer, https://github.com/eic/epic/blob/main/_layouts/geoviewer.html, live at https://eic.github.io/epic/geoviewer, seems to be up to date.

> can provide the exact numbers

Ok, thanks for the info. No numbers needed, or at least not in this PR.

wdconinc:
> linux-eic-shell / dawn-view (view1, epic_brycecanyon) (pull_request) Successful in 2m

As long as this doesn't change, the 2 minutes are fine and we don't need to modify the visualization, I think.

> linux-eic-shell / check-overlap-geant4 (epic_brycecanyon) (pull_request) Successful in 69m

This always takes long, and probably isn't your fault anyway :-D

src/DRICH_geo.cpp
```diff
@@ -494,63 +502,297 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
...
+           *       ||  ||  ||  ||  ||
+           *       ||  ||  ||  ||  || front-end and
+           *       ||  ||  ||  ||  || readout boards
+           *       ||  ||  ||  ||  ||
+           *       ||  ||  ||  ||  ||
+           *               ||
+           *               ||
+           *       +----------------+
+           *       | Sockets, etc.  |  backservices
+           *       +----------------+
```
wdconinc, line 565: May also be useful in xml, though I understand the desire to avoid duplication.

wdconinc (approved):
This looks nice, and adds a great amount of detail!

alexander-kiselev:
Hi Chris,

  no, we don't. pfRICH sensor layout s completely different.

  Cheers,
    Alexander.

________________________________

src/DRICH_geo.cpp
c-dilks: Moved these diagrams to the xml in 1e43f01.

wdconinc:
> This should be merged at the same time as https://github.com/eic/EICrecon/pull/727.

Right! thanks :)

PR #332, opened 2022-12-14
310 improve the neemc acceptance at high eta region

johnny8266:
- [ ] Bug fix (issue #__)
- [X] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No
No


johnny8266:
@wdconinc  I made the pull request yesterday and I thought I didn't do it correctly. Before I made the pull request, I had already made lots commits. I think I should make the pull request after my first commit, right? Is it the reason that make the conflicts?

wdconinc:
You have made your changes against an older version of the geometry. That is why there are conflicts. Please run `git rebase origin/main` and address the conflicts. https://swcarpentry.github.io/git-novice/09-conflict/index.html for info.

wdconinc:
If it says you are this many commits behind the main branch, then you need to make those updates. In your case you are still basing your commits on a main branch that is about 1 month old (from November 16).
image

johnny8266:
> If it says you are this many commits behind the main branch, then you need to make those updates. In your case you are still basing your commits on a main branch that is about 1 month old (from November 16). image

Thanks, I made another pull request to update the changes from main to my branch.

johnny8266:
> > If it says you are this many commits behind the main branch, then you need to make those updates. In your case you are still basing your commits on a main branch that is about 1 month old (from November 16). image
> 
> Thanks, I made another pull request to update the changes from main to my branch.

I saw the comments on the other pull request that I made. And I know that I shouldn't solve the issue in this way.  
I used the git rebase origin/main command and it starts to show the conflict from my first commit to the final one. Can I just skip the first 8 ones by "git rebase --skip" then go to the final one and run "git add src/HomogeneousCalorimeter_geo.cpp" then run git rebase --continue

wdconinc:
> Can I just skip the first 8 ones by "git rebase --skip"

Wish it were that easy... but no. With rebase you have to go through all of them.

If you get too many conflicts early on, I suggest you do `git merge origin/main` instead. We allow that here (not every project does) and it will give you a single set of conflicts to resolve once.

wdconinc (changes requested):
Please remove the commits which included local log and calibrations files.

johnny8266:
I remove the files by the following commands:

git reset --soft HEAD^ 
git restore --staged path/to/unwanted_file  [2 files I guess, log(I used that for debug) and 1calibration file]
git commit -c ORIG_HEAD  
git push -f

src/HomogeneousCalorimeter_geo.cpp
```diff
@@ -406,58 +406,114 @@ static std::tuple<int, int> add_disk(Detector& desc, Assembly& env, xml::Collect
...
+  //=============================
+  // The mother volume of modules
+  //=============================
+  bool has_envelope = dd4hep::getAttrOrDefault<bool>(plm, _Unicode(envelope), false);
+  PolyhedraRegular solid_world(12, 0., r12min, calo_module_length);
+  EllipticalTube  solid_sub(7.5*cm, 5.5*cm, calo_module_length/2.);
+  Transform3D subtract_pos = RotationZYX(-15.*degree, 0., 0.) * Translation3D(0., 0., 0.);
+  SubtractionSolid calo_subtract(solid_world, solid_sub, subtract_pos);
+  Volume      env_vol(std::string(env.name()) + "_envelope", calo_subtract, outer_ring_material);
+  Transform3D tr_global = RotationZYX(15.*degree, 0., 0.) * Translation3D(0., 0., 0.);
```
wdconinc, line 505: This should not use hardcoded constants like 7.5 cm, 5.5 cm, 15 degrees. The motivation of the compact files is that they contain the parameters, and the plugin contains what is done with those parameters.

src/HomogeneousCalorimeter_geo.cpp
```diff
@@ -406,58 +406,114 @@ static std::tuple<int, int> add_disk(Detector& desc, Assembly& env, xml::Collect
...
+
+
+  // // Version2: elliptical shape
+  // //
+  // EllipticalTube   inner_elliptical_support_a(9.*cm, 7.5*cm, calo_module_length/2.);
+  // EllipticalTube   inner_elliptical_support_b(8.5*cm, 7.*cm, calo_module_length/2.);
+  // SubtractionSolid inner_elliptical_support_substract(inner_elliptical_support_a, inner_elliptical_support_b, Position(0., 0., 0.));
+  // Volume           inner_elliptical_vol("inner_elliptical", inner_elliptical_support_substract, inner_ring_material);
+  // inner_elliptical_vol.setVisAttributes(desc.visAttributes(plm.attr<std::string>(_Unicode(vis_struc))));
+  // Transform3D tr_global_Iring_elli = RotationZYX(-15.*degree, 0., 0.) * Translation3D(2.05*cm, 0., 0.);
```
wdconinc, line 469: Extensive commented sections should be removed. Since these sections are commented out, they are never compiled and when other parts of this file are changed they can and will become uncompilable or even wrong.

src/HomogeneousCalorimeter_geo.cpp
```diff
@@ -406,58 +406,114 @@ static std::tuple<int, int> add_disk(Detector& desc, Assembly& env, xml::Collect
...
+  inner_support_vol.setVisAttributes(desc.visAttributes(plm.attr<std::string>(_Unicode(vis_struc))));
+  Transform3D tr_global_Iring_elli = RotationZYX(-15.*degree, 0., 0.) * Translation3D(0., 0., 0.);
+
+
+  // // The vacuum inside the inner structure
+  // //
+  // EllipticalTube   inner_elliptical_vacuum(7.5*cm, 4.5*cm, calo_module_length/2.);
+  // Volume           inner_elliptical_vacuum_vol("inner_elliptical_vacuum_vol", inner_elliptical_vacuum, hole_material);
+  // inner_elliptical_vacuum_vol.setVisAttributes(desc.visAttributes(plm.attr<std::string>(_Unicode(vis_struc))));
+  // Transform3D tr_global_Iring_elli_vacuum = RotationZYX(-15.*degree, 0., 0.) * Translation3D(1.025*cm, 0., 0.);
```
wdconinc, line 493: Remove commented sections.

src/HomogeneousCalorimeter_geo.cpp
```diff
@@ -467,60 +523,85 @@ static std::tuple<int, int> add_12surface_disk(Detector& desc, Assembly& env, xm
...
+      // env.placeVolume(cabling_support12_V, tr_global_csfront);  // Place the front cabling frame(Only the holes are visible)
+      // env.placeVolume(cabling_support12_V, tr_global_csback);   // Place the back cabling frame(Only the holes are visible)
+
+      // for(int i = 0 ; i < 12 ; i++)
+      //   {
+      //     hole_pos = RotationZYX((15. + i * 30.)*degree, 0., 0.) * Translation3D(82.5*mm, 675.*mm, 0.);
+      //     cabling_support12_V.placeVolume(hole_V, hole_pos);
+      //     hole_pos = RotationZYX((15. + i * 30.)*degree, 0., 0.) * Translation3D(-82.5*mm, 675.*mm, 0.);
+      //     cabling_support12_V.placeVolume(hole_V, hole_pos);
+      //   }
```
wdconinc, line 544: Remove commented sections.

src/HomogeneousCalorimeter_geo.cpp
```diff
@@ -467,60 +523,85 @@ static std::tuple<int, int> add_12surface_disk(Detector& desc, Assembly& env, xm
...
+    }
+
+  // Add the modules manually in the gap [Outer]
+  //
+  const int add_N_mod_outer = 56;
+  float addX_outer[add_N_mod_outer] = {9.225, 11.275, 13.325, 15.375, 17.425, 19.475, 44.075, 46.125, 60.475, 60.475, 62.525, 62.525, 62.525, 62.525,  -9.225, -11.275, -13.325, -15.375, -17.425, -19.475, -44.075, -46.125, -60.475, -60.475, -62.525, -62.525, -62.525, -62.525,  -9.225, -11.275, -13.325, -15.375, -17.425, -19.475, -44.075, -46.125, -60.475, -60.475, -62.525, -62.525, -62.525, -62.525,  9.225, 11.275, 13.325, 15.375, 17.425, 19.475, 44.075, 46.125, 60.475, 60.475, 62.525, 62.525, 62.525, 62.525};
+  float addY_outer[add_N_mod_outer] = {62.525, 62.525, 62.525, 62.525, 60.475, 60.475, 46.125, 44.075, 19.475, 17.425, 15.375, 13.325, 11.275, 9.225,  62.525, 62.525, 62.525, 62.525, 60.475, 60.475, 46.125, 44.075, 19.475, 17.425, 15.375, 13.325, 11.275, 9.225,  -62.525, -62.525, -62.525, -62.525, -60.475, -60.475, -46.125, -44.075, -19.475, -17.425, -15.375, -13.325, -11.275, -9.225, -62.525, -62.525, -62.525, -62.525, -60.475, -60.475, -46.125, -44.075, -19.475, -17.425, -15.375, -13.325, -11.275, -9.225};
```
wdconinc, line 580: Remove hardcoded positions. Avoid arrays and use vectors instead.

src/HomogeneousCalorimeter_geo.cpp
```diff
@@ -467,60 +523,85 @@ static std::tuple<int, int> add_12surface_disk(Detector& desc, Assembly& env, xm
...
+      auto modPV = (has_envelope ? env_vol.placeVolume(modVol, add_local) : env.placeVolume(modVol, tr_global * add_local));
+      modPV.addPhysVolID("sector", sector_id).addPhysVolID("module", total_id);
+      total_id++;
+    }
+
+  // Add the modules manually in the gap [Inner]
+  //
+  const int add_N_mod_inner = 36;
+  double addX_inner[add_N_mod_inner] = {1.025,1.025,3.075,3.075,5.125,7.175,7.175,9.225,9.225, -1.025,-1.025,-3.075,-3.075,-5.125,-7.175,-7.175,-9.225,-9.225, -1.025,-1.025,-3.075,-3.075,-5.125,-7.175,-7.175,-9.225,-9.225, 1.025,1.025,3.075,3.075,5.125,7.175,7.175,9.225,9.225};
+  double addY_inner[add_N_mod_inner] = {7.175,9.225,7.175,9.225,7.175,7.175,5.125,1.025,3.075, 7.175,9.225,7.175,9.225,7.175,7.175,5.125,1.025,3.075, -7.175,-9.225,-7.175,-9.225,-7.175,-7.175,-5.125,-1.025,-3.075, -7.175,-9.225,-7.175,-9.225,-7.175,-7.175,-5.125,-1.025,-3.075};
```
wdconinc, line 594: Remove hardcoded positions. Avoid arrays and use vectors instead.

src/HomogeneousCalorimeter_geo.cpp
```diff
@@ -467,60 +523,85 @@ static std::tuple<int, int> add_12surface_disk(Detector& desc, Assembly& env, xm
...
+  int mid = 0, total_id = 0;
+  float half_modx = modSize.x() * 0.5, half_mody = modSize.y() * 0.5;
+  auto points = epic::geo::fillRectangles({half_modx, half_mody}, modSize.x(), modSize.y(), rmin, rmax, phimin, phimax);
+  for (auto& p : points)
+    {
+      Transform3D tr_local = RotationZYX(-15.*degree, 0.0, 0.0) * Translation3D(p.x(), p.y(), 0.0);
+      auto modPV = (has_envelope ? env_vol.placeVolume(modVol, tr_local) : env.placeVolume(modVol, tr_global * tr_local));
+      modPV.addPhysVolID("sector", sector_id).addPhysVolID("module", total_id);
+      total_id = id_begin + mid++;
+      // std::cout << total_id << " " << p.x() << " " << p.y() << std::endl;  // use for debug
```
wdconinc, line 573: Remove commented debugging lines.

src/HomogeneousCalorimeter_geo.cpp
```diff
@@ -467,60 +523,85 @@ static std::tuple<int, int> add_12surface_disk(Detector& desc, Assembly& env, xm
...
-    //     hole_pos = RotationZYX((15. + i * 30.)*degree, 0., 0.) * Translation3D(-82.5*mm, 675.*mm, 0.);
-    //     cabling_support12_V.placeVolume(hole_V, hole_pos);
-    //   }
-  }
+  if (has_envelope)
+    {
+      env.placeVolume(env_vol, tr_global);                          // Place the mother volume for all modules
+      env.placeVolume(ring12_vol, tr_global_Oring);                 // Place the outer supporting frame
+      // env.placeVolume(Sring12_vol, tr_global_Iring);                // Place the version1 inner supporting frame
+      // env_vol.placeVolume(inner_elliptical_vol, tr_global_Iring_elli);  // Place the version2 inner supporting frame
```
wdconinc, line 531: Remove commented lines.

src/HomogeneousCalorimeter_geo.cpp
```diff
@@ -406,58 +406,114 @@ static std::tuple<int, int> add_disk(Detector& desc, Assembly& env, xml::Collect
...
+  Box subtract_corner(1.025*cm, 1.025*cm, calo_module_length/2.);
+  EllipticalTube   subtract_a(7.5*cm, 5.5*cm, calo_module_length/2.);
+  SubtractionSolid inner_support_substractb1(inner_support_main, subtract_corner, Position(7.175*cm, 5.125*cm, 0.));
+  SubtractionSolid inner_support_substractb2(inner_support_substractb1, subtract_corner, Position(7.175*cm, -5.125*cm, 0.));
+  SubtractionSolid inner_support_substractb3(inner_support_substractb2, subtract_corner, Position(-7.175*cm, 5.125*cm, 0.));
+  SubtractionSolid inner_support_substractb4(inner_support_substractb3, subtract_corner, Position(-7.175*cm, -5.125*cm, 0.));
+  SubtractionSolid inner_support_substracta(inner_support_substractb4, subtract_a, Position(0., 0., 0.));
+  Volume           inner_support_vol("inner_support_vol", inner_support_substracta, inner_ring_material);
+  inner_support_vol.setVisAttributes(desc.visAttributes(plm.attr<std::string>(_Unicode(vis_struc))));
+  Transform3D tr_global_Iring_elli = RotationZYX(-15.*degree, 0., 0.) * Translation3D(0., 0., 0.);
```
wdconinc, line 485: Move hardcoded constants to compact file.

src/HomogeneousCalorimeter_geo.cpp
```diff
@@ -406,58 +406,114 @@ static std::tuple<int, int> add_disk(Detector& desc, Assembly& env, xml::Collect
...
-  //
+  Box inner_support_main(8.2*cm, 6.15*cm, calo_module_length/2.);  // Original size
+  // Box inner_support_main(8.2*cm, 5.125*cm, calo_module_length/2.);  // Adapted size
+  Box subtract_corner(1.025*cm, 1.025*cm, calo_module_length/2.);
+  EllipticalTube   subtract_a(7.5*cm, 5.5*cm, calo_module_length/2.);
+  SubtractionSolid inner_support_substractb1(inner_support_main, subtract_corner, Position(7.175*cm, 5.125*cm, 0.));
+  SubtractionSolid inner_support_substractb2(inner_support_substractb1, subtract_corner, Position(7.175*cm, -5.125*cm, 0.));
+  SubtractionSolid inner_support_substractb3(inner_support_substractb2, subtract_corner, Position(-7.175*cm, 5.125*cm, 0.));
+  SubtractionSolid inner_support_substractb4(inner_support_substractb3, subtract_corner, Position(-7.175*cm, -5.125*cm, 0.));
+  SubtractionSolid inner_support_substracta(inner_support_substractb4, subtract_a, Position(0., 0., 0.));
```
wdconinc, line 482: Avoid serial solid subtractions because they are slow, in particular in high-occupancy regions. Use ExtrudedPolygon instead.

wdconinc (changes requested):

src/HomogeneousCalorimeter_geo.cpp
johnny8266: Thanks for the advice, I think it will be more elegant to do this. Do you have any example can help me how to implement  ExtrudedPolygon? I tried to search this, but they are not clear to me.

src/HomogeneousCalorimeter_geo.cpp
```diff
@@ -406,58 +406,114 @@ static std::tuple<int, int> add_disk(Detector& desc, Assembly& env, xml::Collect
 static std::tuple<int, int> add_12surface_disk(Detector& desc, Assembly& env, xml::Collection_t& plm,
                                                SensitiveDetector& sens, int sid)
 {
```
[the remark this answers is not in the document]
johnny8266: I was thinking how to arrange the module tightly along the volume border, but I realize it would be difficult math problem and not possible/very difficult to fill the rectangle in the polygon. So, I used the defaulted fillrectangle function and fill the gap manually. I know it is not convenient for people want to parameterized the size of module for some tests as they will just get overlap issue with bigger module. I think, instead of parameterizing the position of modules at edge, I could built a boolean flag for people to turn modules at the edge on or off.

src/HomogeneousCalorimeter_geo.cpp
johnny8266: I remove the hardcoded positions and read them by vectors

src/HomogeneousCalorimeter_geo.cpp
johnny8266: remove a serial of solid subtractions and use the ExtrudedPolygon instead.

src/HomogeneousCalorimeter_geo.cpp
johnny8266: Remove the those number and put them into the configuration files

src/HomogeneousCalorimeter_geo.cpp
johnny8266: got it, I remove those comment region in the last commit

johnny8266:
@wdconinc I resolve the conversation by remove the comments, hardcoded parameters and read the data by vector instead of array. could you help check is good to merge? Thanks!

wdconinc:
Will give another review today.

johnny8266:
> Will give another review today.

Hi Wouter,  do you have chance to give it a look?

src/HomogeneousCalorimeter_geo.cpp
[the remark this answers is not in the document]
wdconinc: This is still unresolved. You took a general geometry plugin allowing for general specification homogeneous calorimeters, and you turned it into the N EEMC plugin (for good reason, I don't have any quibble with that). There are potentially other subsystems using this plugin, or considering it (there are far foward calorimeters that don't need more than the general approach). It's not maintainable to keep adding NEEMC details to the general cpp plugin. Please split off your changes to a different source file with only the parts you are actually using, and name it so it is clear that it is specific to the negative endcap.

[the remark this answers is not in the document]
johnny8266: I got it. Should I keep this specified plugin in the general NEEMC file when I split off this configuration?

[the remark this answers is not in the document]
johnny8266: @wdconinc Could I make another pull request for this? This pull request is about the updated and improved geometry close to beam pipe. And the work of spin off the this configuration from the homogenouscalorimeter would be a lots and has nothing to do with this pull request. Could I merge this into the main if there is no problem other than making another geometry plugin for NEEMC?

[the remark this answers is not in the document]
johnny8266: And sorry that I was not able to think about this recently, but I will open the new pull request about this, if you agreed with me, and back to finish this.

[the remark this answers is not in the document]
johnny8266: Thanks for giving another solution for this conversation. I change the way to assign cell ID in my last commit by using the row and column id instead of the sequential ID. @wdconinc could you help me do another review. Thanks!

compact/ecal/backward_PbWO4.xml
```diff
@@ -80,7 +99,7 @@
     </comment>
     <readout name="EcalEndcapNHits">
       <segmentation type="NoSegmentation" key="sector"/>
-      <id>system:8,sector:4,module:20</id>
+      <id>system:8,sector:4,row:10,column:10,module:20</id>
```
veprbl, line 102: 256 cell rows/columns should be more than enough
```suggestion
      <id>system:8,sector:4,row:8,column:8</id>
```

src/HomogeneousCalorimeter_geo.cpp
```diff
@@ -288,311 +177,181 @@ std::tuple<Volume, Position> build_module(Detector& desc, xml::Collection_t& plm
...
-        trans.push_back(Position{trans[i].x(), trans[i].y(), -trans[i].z()});
-      }
-    }
+  // Place the modules, use row and column ID for each modules
+  // row and column ID start from the top left corner
+  //
+  int row = 0, column = 0;
+  for (auto& p : points) {
+    column = (p.x() - min_ptsx) / modSize.x();
+    row = (max_ptsy - p.y()) / modSize.y();
```
veprbl, line 337: Would be nice to configure cells starting from integer row/column pairs and then multiplying by modSize. If you insist on doing it backwards, at least let's
```suggestion
    column = std::round((p.x() - min_ptsx) / modSize.x());
    row = std::round((max_ptsy - p.y()) / modSize.y());
```

compact/ecal/backward_PbWO4.xml
johnny8266: Thanks for correct :)

src/HomogeneousCalorimeter_geo.cpp
johnny8266: I started from the position as there is a function to fill the circle with box, which will output the positions of boxes.

compact/ecal/backward_PbWO4.xml
```diff
@@ -39,11 +49,20 @@
...
+            inner_outer_add_posy="7.175 9.225 7.175 9.225 7.175 7.175 5.125 1.025 3.075 7.175 9.225 7.175 9.225 7.175 7.175 5.125 1.025 3.075 -7.175 -9.225 -7.175 -9.225 -7.175 -7.175 -5.125 -1.025 -3.075 -7.175 -9.225 -7.175 -9.225 -7.175 -7.175 -5.125 -1.025 -3.075 62.525 62.525 62.525 62.525 60.475 60.475 46.125 44.075 19.475 17.425 15.375 13.325 11.275 9.225 62.525 62.525 62.525 62.525 60.475 60.475 46.125 44.075 19.475 17.425 15.375 13.325 11.275 9.225 -62.525 -62.525 -62.525 -62.525 -60.475 -60.475 -46.125 -44.075 -19.475 -17.425 -15.375 -13.325 -11.275 -9.225 -62.525 -62.525 -62.525 -62.525 -60.475 -60.475 -46.125 -44.075 -19.475 -17.425 -15.375 -13.325 -11.275 -9.225"
```
wdconinc, line 62: Keeping the x and y coordinates in separate lists is not ideal. The values that belong together are not aligned, and no one can figure out which of them fit together without significant effort. If you want to store arrays of structured info, you should store your arrays as structured info. Take a look at https://github.com/eic/epic/blob/main/compact/pid/mrich.xml for another detector that specifies lots of positions in the compact file.

compact/ecal/backward_PbWO4.xml
johnny8266: solve in the last commit

src/HomogeneousCalorimeter_geo.cpp
johnny8266: solved in the last commit

src/HomogeneousCalorimeter_geo.cpp
```diff
@@ -288,311 +177,181 @@ std::tuple<Volume, Position> build_module(Detector& desc, xml::Collection_t& plm
...
+  for (auto& p : points) {
+    if (p.x() < min_ptsx)
+      min_ptsx = p.x();
+    if (p.x() > max_ptsx)
+      max_ptsx = p.x();
+    if (p.y() < min_ptsy)
+      min_ptsy = p.y();
+    if (p.y() > max_ptsy)
+      max_ptsy = p.y();
+  }
```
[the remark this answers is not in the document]
johnny8266: Thanks for the source!!!
implemented in the last commit.

src/HomogeneousCalorimeter_geo.cpp
```diff
@@ -288,311 +177,181 @@ std::tuple<Volume, Position> build_module(Detector& desc, xml::Collection_t& plm
...
-  // Place the modules in the circle
-  for (auto& p : points) {
-    Transform3D tr_local = RotationZYX(-15. * degree, 0.0, 0.0) * Translation3D(p.x(), p.y(), 0.0);
-    modPV    = (has_envelope ? env_vol.placeVolume(modVol, tr_local) : env.placeVolume(modVol, tr_global * tr_local));
-    total_id = id_begin + mid++;
-    modPV.addPhysVolID("sector", sector_id).addPhysVolID("module", total_id);
-  }
+  // Add the modules followd the fillRectangles function
+  //
+  int   mid = 0;
```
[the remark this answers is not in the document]
johnny8266: @wdconinc this number is used for the module id, and I use row and column to replaced it. Could you point me how other sub-detectors return row and column ID? I didn't remove this variable in my last commit since I need to change lots of things related if I do so. Thanks!

src/HomogeneousCalorimeter_geo.cpp
```diff
@@ -288,311 +177,181 @@ std::tuple<Volume, Position> build_module(Detector& desc, xml::Collection_t& plm
...
+  std::vector<double> inner_outer_posx;
+  std::vector<double> inner_outer_posy;
+
+  pos = 0;
+  while ((pos = iposx.find(delimiter)) != std::string::npos) {
+    token = iposx.substr(0, pos);
+    inner_outer_posx.push_back(atof(token.c_str()));
+    iposx.erase(0, pos + delimiter.length());
   }
+  inner_outer_posx.push_back(atof(iposx.c_str()));
```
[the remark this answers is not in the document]
johnny8266: Solved in the last commit.

src/HomogeneousCalorimeter_geo.cpp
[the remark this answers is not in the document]
wdconinc: Instead of `int mid`, you are essentially using `int row, column`, so I think you should modify `add_12surface_disk` to return  `std::tuple<int, int, int>` and use it as
```cpp
  for (xml::Collection_t disk_12surface(plm, _Unicode(disk_12surface)); disk_12surface; ++disk_12surface) {
    auto [sector, row, column] = add_12surface_disk(desc, assembly, disk_12surface, sens, sector_id++);
    addModuleNumbers(sector, row, column);
  }
```
where then both `addModuleNumbers` and its loop would need some changes too.

src/HomogeneousCalorimeter_geo.cpp
```diff
@@ -70,23 +70,41 @@ static Ref_t create_detector(Detector& desc, xml::Handle_t handle, SensitiveDete
   // module placement
   xml::Component     plm = detElem.child(_Unicode(placements));
   std::map<int, int> sectorModuleNumbers;
```
johnny8266, line 72: It looks I can't put sector ID, row and column information into the map container here.

src/HomogeneousCalorimeter_geo.cpp
```diff
@@ -70,23 +70,41 @@ static Ref_t create_detector(Detector& desc, xml::Handle_t handle, SensitiveDete
...
   }
 
-  for (auto [sector, nmods] : sectorModuleNumbers) {
-    desc.add(Constant(Form((detName + "_NModules_Sector%d").c_str(), sector), std::to_string(nmods)));
+  // for (auto [sector, nmods] : sectorModuleNumbers) {
+  //   desc.add(Constant(Form((detName + "_NModules_Sector%d").c_str(), sector), std::to_string(nmods)));
+  // }
+
+  for (auto [sector, nrow, ncolumn] : sectorModuleNumbers) {
+    desc.add(Constant(Form((detName + "_NModules_Sector%d").c_str(), sector), std::to_string(nrow), std::to_string(ncolumn)));
```
johnny8266, line 105: Could I naively add one more argument at here?

src/HomogeneousCalorimeter_geo.cpp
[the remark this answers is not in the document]
johnny8266: Thanks for answering. I tried to adapt it in my last commit, but I also have some questions about how those sector ID, row and column information pass between these functions.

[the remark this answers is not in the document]
johnny8266: Solved in the last commit.

src/HomogeneousCalorimeter_geo.cpp
```diff
@@ -70,23 +70,41 @@ static Ref_t create_detector(Detector& desc, xml::Handle_t handle, SensitiveDete
...
+  //     sectorModuleNumbers[sector] = nmod;
+  //   }
+  // };
+
+  auto               addModuleNumbers = [&sectorModuleNumbers](int sector, int nrow, int ncolumn) {
     auto it = sectorModuleNumbers.find(sector);
     if (it != sectorModuleNumbers.end()) {
       it->second += nmod;
     } else {
       sectorModuleNumbers[sector] = nmod;
```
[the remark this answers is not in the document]
johnny8266: solved by std::pair

compact/ecal/backward_PbWO4.xml
johnny8266: done by the last commit

johnny8266:
@wdconinc would you be available to give this a review? I just solve all these conversations. Thanks!

wdconinc:
I'll review today but it's my teaching day.

compact/ecal/backward_PbWO4.xml
```diff
@@ -14,9 +14,19 @@
     <constant name="EcalEndcapN_structure_ring_max" value="65.00*cm"/>
     <constant name="EcalEndcapN_structure_ring_min" value="EcalEndcapN_structure_ring_max - 0.9*cm"/>
 
+    <comment>Basic hardcoded parameters about rotation and shift</comment>
+    <constant name="EcalEndcapN_Protate" value="15.*degree"/>
```
wdconinc, line 18: Add more documentation to these variables. It's unclear what Protate or Nrotate are supposed to mean.

compact/ecal/backward_PbWO4.xml
```diff
@@ -14,9 +14,19 @@
...
+    <constant name="EcalEndcapN_Nrotate" value="-15.*degree"/>
+    <constant name="EcalEndcapN_Oring_relative_shift" value="-20.*cm"/>
```
wdconinc, line 20: Same thing about Iring and Oring being unclear.

compact/ecal/backward_PbWO4.xml
```diff
@@ -39,11 +49,135 @@
             r12max="EcalEndcapN_structure_ring_max"
             SFlength="EcalEndcapN_structure_ring_length"
             CMlength="EcalEndcapN_CrystalModule_length + 2.*EcalEndcapN_CrystalModule_wrap_thickness"
+            NEEMC_PR="EcalEndcapN_Protate"
```
wdconinc, line 52: Stick to one variable name scheme. You have some that are all caps NEEEMC_PR, some that are lowercase underscore. Don't be redundant in fields: that this is the NEEMC is clear from the plugin so just focus on the parameters inside the NEEMC.

compact/ecal/backward_PbWO4.xml
```diff
@@ -39,11 +49,135 @@
...
-            vis_struc="RPVis"
-            vis_steel_gap="AnlGray"
             ring_material="StainlessSteel"
-            sector="1">
+            sector="1"
+            vis_struc="RPVis"
+            vis_steel_gap="AnlGray">
+
+          <comment>
+            The position information[cm] of inner supporting structure [extrudedpolygon shape, 12 points in total]
```
wdconinc, line 64: You shouldn't specify the dimensions in a comment. Specify it explicitly in the value, eg `12*cm`.

compact/ecal/backward_PbWO4.xml
```diff
@@ -39,11 +49,135 @@
...
+            <position x="-6.15" y="4.1"/>
+            <position x="-6.15" y="6.15"/>
+          </positions>
+
+          <comment>
+            The position information[cm] of additional modules to fill the gap
+            between the sensitive region and inner/outer frame structure. [92 posints in total]
+          </comment>
+          <positions name="additionalmodule_pos">
+            <position x="1.025"  y="7.175"/>
```
wdconinc, line 86: Same here, all positions should have an explicit unit multiplied.

compact/ecal/backward_PbWO4.xml
```diff
@@ -39,11 +49,135 @@
...
             ring_material="StainlessSteel"
-            sector="1">
+            sector="1"
+            vis_struc="RPVis"
+            vis_steel_gap="AnlGray">
+
+          <comment>
+            The position information[cm] of inner supporting structure [extrudedpolygon shape, 12 points in total]
+          </comment>
+          <positions name="extrudedpolygon_pos">
```
wdconinc, line 66: Don't call this positions but something descriptive.

compact/ecal/backward_PbWO4.xml
```diff
@@ -39,11 +49,135 @@
...
+            <position x="-8.2"  y="4.1"/>
+            <position x="-6.15" y="4.1"/>
+            <position x="-6.15" y="6.15"/>
+          </positions>
+
+          <comment>
+            The position information[cm] of additional modules to fill the gap
+            between the sensitive region and inner/outer frame structure. [92 posints in total]
+          </comment>
+          <positions name="additionalmodule_pos">
```
wdconinc, line 85: Calling this positions is probably ok.

src/HomogeneousCalorimeter_geo.cpp
```diff
@@ -40,101 +34,14 @@ using namespace dd4hep;
...
 
 // headers
-static std::tuple<int, int> add_individuals(Detector& desc, Assembly& env, xml::Collection_t& plm,
-                                            SensitiveDetector& sens, int id);
-static std::tuple<int, int> add_array(Detector& desc, Assembly& env, xml::Collection_t& plm, SensitiveDetector& sens,
-                                      int id);
-static std::tuple<int, int> add_disk(Detector& desc, Assembly& env, xml::Collection_t& plm, SensitiveDetector& sens,
-                                     int id);
-static std::tuple<int, int> add_12surface_disk(Detector& desc, Assembly& env, xml::Collection_t& plm,
+// static std::tuple<int, int> add_12surface_disk(Detector& desc, Assembly& env, xml::Collection_t& plm,
```
wdconinc, line 41: Remove unused commented lines. Git history has them.

src/HomogeneousCalorimeter_geo.cpp
```diff
@@ -40,101 +34,14 @@ using namespace dd4hep;
...
+//                                                SensitiveDetector& sens, int id);
+static std::tuple<int, std::pair<int, int>> add_12surface_disk(Detector& desc, Assembly& env, xml::Collection_t& plm,
```
wdconinc, line 43: Tuple of int and pair of ints is kinda redundant. Why not a tuple of three ints?

src/HomogeneousCalorimeter_geo.cpp
```diff
@@ -288,311 +184,153 @@ std::tuple<Volume, Position> build_module(Detector& desc, xml::Collection_t& plm
...
+  double      rmax                   = plm.attr<double>(_Unicode(rmax));
+  double      r12min                 = plm.attr<double>(_Unicode(r12min));
+  double      r12max                 = plm.attr<double>(_Unicode(r12max));
+  double      structure_frame_length = plm.attr<double>(_Unicode(SFlength));
+  double      calo_module_length     = plm.attr<double>(_Unicode(CMlength));
+  double      NEEMC_Prot             = plm.attr<double>(_Unicode(NEEMC_PR));
+  double      NEEMC_Nrot             = plm.attr<double>(_Unicode(NEEMC_NR));
+  double      NEEMC_OR_shift         = plm.attr<double>(_Unicode(NEEMC_OR_RS));
+  double      NEEMC_IR_a             = plm.attr<double>(_Unicode(Inner_a));
+  double      NEEMC_IR_b             = plm.attr<double>(_Unicode(Inner_b));
```
wdconinc, line 204: Avoid prefix NEEMC here. That's redundant in this plugin.

src/HomogeneousCalorimeter_geo.cpp
```diff
@@ -288,311 +184,153 @@ std::tuple<Volume, Position> build_module(Detector& desc, xml::Collection_t& plm
...
+  //=========================================================
+  for (xml_coll_t x_positions_i(plm, _Unicode(positions)); x_positions_i; ++x_positions_i){
+    xml_comp_t x_positions = x_positions_i;
+
+    for (xml_coll_t x_position_i(x_positions, _U(position)); x_position_i; ++x_position_i){
+      xml_comp_t x_position = x_position_i;
+
+      if( x_positions.nameStr() == "extrudedpolygon_pos" ){  // retrieve the position of extrudedpolygon shape supporting structure
+        pt_innerframe_x.push_back((x_position.x())*cm);
+        pt_innerframe_y.push_back((x_position.y())*cm);
```
wdconinc, line 231: Don't store in two variables x and y, but in a tuple or pair.

src/HomogeneousCalorimeter_geo.cpp
```diff
@@ -288,311 +184,153 @@ std::tuple<Volume, Position> build_module(Detector& desc, xml::Collection_t& plm
...
+  //=========================================================
+  // Read the positions information from xml file
+  //=========================================================
+  for (xml_coll_t x_positions_i(plm, _Unicode(positions)); x_positions_i; ++x_positions_i){
+    xml_comp_t x_positions = x_positions_i;
+
+    for (xml_coll_t x_position_i(x_positions, _U(position)); x_position_i; ++x_position_i){
+      xml_comp_t x_position = x_position_i;
+
+      if( x_positions.nameStr() == "extrudedpolygon_pos" ){  // retrieve the position of extrudedpolygon shape supporting structure
```
wdconinc, line 229: Don't use the string comparison. Use a different name for the xml entity that contains these positions vis a vis the others.

src/HomogeneousCalorimeter_geo.cpp
```diff
@@ -288,311 +184,153 @@ std::tuple<Volume, Position> build_module(Detector& desc, xml::Collection_t& plm
...
-  // Volume hole_V("hole_V", hole, hole_material);
-  // Transform3D hole_pos = RotationZYX(0., 0., 0.) * Translation3D(0., 0., 0.);
+  // Version3: solid with elliptical inside
+  //
+  std::vector<double> sec_z  = {-calo_module_length / 2., calo_module_length / 2.};
+  std::vector<double> sec_x  = {0., 0.};
+  std::vector<double> sec_y  = {0., 0.};
+  std::vector<double> zscale = {1., 1.};
+
+  ExtrudedPolygon  inner_support_main(pt_innerframe_x, pt_innerframe_y, sec_z, sec_x, sec_y, zscale);
```
wdconinc, line 271: I can see the use for the separate x and y variables now.

wdconinc (commented):
Great progress! Few more comments. We should get this merged and we definitely plan on using it in the end of March simulation campaign.

compact/ecal/backward_PbWO4.xml
johnny8266: done

compact/ecal/backward_PbWO4.xml
johnny8266: done

src/HomogeneousCalorimeter_geo.cpp
johnny8266: I save them in 2 vectors because the first two input arguments of ExtrudedPolygon are vectors.

compact/ecal/backward_PbWO4.xml
johnny8266: done

compact/ecal/backward_PbWO4.xml
johnny8266: modified the variable names

src/HomogeneousCalorimeter_geo.cpp
```diff
@@ -288,311 +184,153 @@ std::tuple<Volume, Position> build_module(Detector& desc, xml::Collection_t& plm
...
+
+    for (xml_coll_t x_position_i(x_positions, _U(position)); x_position_i; ++x_position_i){
+      xml_comp_t x_position = x_position_i;
+
+      if( x_positions.nameStr() == "extrudedpolygon_pos" ){  // retrieve the position of extrudedpolygon shape supporting structure
+        pt_innerframe_x.push_back((x_position.x())*cm);
+        pt_innerframe_y.push_back((x_position.y())*cm);
+      }
+      else{  // retrieve the position of additional modules
+        auto add_point = epic::geo::Point((x_position.x())*cm, (x_position.y())*cm);
```
wdconinc, line 234: Related with adding dimensions like `cm` into the xml file, you would have to remove them from here. For physical quantities, always keep the values together with their dimension for clarity. This currently puts the dimension in the cpp file, but the values in the xml file.

johnny8266: Got it! this is the same issue that I need to assign unit in the xml file. I remove this in my next commit. I will push it soon.

wdconinc (approved):

wdconinc:
Now we just need to resolve the overlaps: https://github.com/eic/epic/actions/runs/4449730155/jobs/7814405594?pr=332. I'd suggest figuring out first if this is from the individually placed modules or the rest. The overlap of 500 um seems to be due to ignoring the thickness of VM2000 and carbon fiber (once on each neighboring module), likely placing the modules with a pitch given by the crystals, but then placing a module that is larger than that pitch by 500 um.

johnny8266:
> Now we just need to resolve the overlaps: https://github.com/eic/epic/actions/runs/4449730155/jobs/7814405594?pr=332. I'd suggest figuring out first if this is from the individually placed modules or the rest. The overlap of 500 um seems to be due to ignoring the thickness of VM2000 and carbon fiber (once on each neighboring module), likely placing the modules with a pitch given by the crystals, but then placing a module that is larger than that pitch by 500 um.

Got it! I checked the position and the sizes but I still not sure where the overlap is. I used the "checkOverlap -t 0.001 -c epic_NEEMC.xml" epic_NEEMC.xml is NEEMC standalone geometry, and the results show Number of illegal overlaps/extrusions : 0. 
I comment the additional modules and push it again to see where is the problem.

compact/ecal/backward_PbWO4.xml
```diff
@@ -110,7 +110,7 @@
             <position x="7.175*cm"  y="7.175*cm"/>
             <position x="7.175*cm"  y="5.125*cm"/>
             <position x="9.225*cm"  y="1.025*cm"/>
-            <position x="9.225*cm"  y="3.075*cm"/>
+            <position x="9.225*cm"  y="3.075*cm"/>           
             <position x="-1.025*cm" y="7.175*cm"/>
             <position x="-1.025*cm" y="9.225*cm"/>
             <position x="-3.025*cm" y="7.175*cm"/>
```
johnny8266, line 116: The problem is this value..... it supposed to be -3.075*cm, and it became -3.025*cm during copy paste..... it is totally eye test :)

johnny8266: I will fix this in my next commit because I want to see is there any other overlapped issue appears in my last commit.

wdconinc: This is of course *exactly* why we don't want to have to maintain these long lists of module positions in a file but prefer to calculate them directly in the code...

johnny8266: I totally agree and I just suffered from that...... I will try to think about how to fill the box in the 12-sides polygon by calculation but I think it is fine to fill the gap between inner side and beam pipe by hardcoded, as number of modules are relatively less and there might have new design of supporting structure. I will add some comments in xml file and put 1 or 2 line for spacing inner and outer additional modules.

johnny8266: correct the values in the last commit

wdconinc (approved):

veprbl (approved):

PR #337, opened 2022-12-20
Create CITATION.cff

simonge:
Adds a CITATION.cff so that the code repository has an easily accessible and standardised style for referencing from other documents.

Initially only a barebones there probably needs to be some discussion about what/who should be included.


wdconinc (dismissed):

CITATION.cff
```diff
@@ -0,0 +1,10 @@
+cff-version: 1.2.0
+title: 'EPIC Experiment Simulation '
```
veprbl, line 2:
```suggestion
title: 'DD4hep Geometry Description of the EPIC Experiment'
```

wdconinc (approved):

PR #341, opened 2022-12-21
fix: convert to gdml and tgeo (root) for export to gh-pages

wdconinc:
This adds exports to tgeo and gdml for the major epic_arches and epic_brycecanyon configurations (was only done for epic). This also introduces a github-pages geoviewer for these geometries. 

- [x] Bug fix (issue: incorrect geometry exported to gdml and tgeo)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


wdconinc:
This is demo'd at https://wdconinc.github.io/epic/ but links in the top level page are explicitly linking to the eic.github.io url because otherwise they won't work in the readme view on GitHub itself.

.github/workflows/linux-eic-shell.yml
```diff
@@ -313,17 +325,18 @@ jobs:
         setup: install/setup.sh
         run: |
           mkdir -p doc
-          npdet_info dump ${DETECTOR_PATH}/epic.xml | grep -v '^\s' | grep '=' | cut -d= -f1-2 | tee doc/constants.toml
-          python bin/make_detector_parameter_table | tee doc/DetectorParameterTable.csv
+          npdet_info dump ${DETECTOR_PATH}/epic.xml | grep -v '^\s' | grep '=' | cut -d= -f1-2 | tee doc/${{matrix.detector_config}}_constants.toml
```
c-dilks, line 328:
```suggestion
          npdet_info dump ${DETECTOR_PATH}/${{matrix.detector_config}}.xml | grep -v '^\s' | grep '=' | cut -d= -f1-2 | tee doc/${{matrix.detector_config}}_constants.toml
```

.github/workflows/linux-eic-shell.yml
```diff
@@ -264,17 +271,19 @@ jobs:
...
+        name: tgeo
+        path: ${{ matrix.detector_config }}.root
         if-no-files-found: error
 
   dump-constants:
     runs-on: ubuntu-latest
     needs: build-gcc
+    strategy:
+      matrix:
+        detector_config: [epic_arches, epic_brycecanyon]
```
c-dilks, line 286: any reason we don't use `matrix: ${{ fromJson(needs.list-detector-configs.outputs.configs) }}` here too?

.github/workflows/linux-eic-shell.yml
```diff
@@ -289,16 +298,19 @@ jobs:
...
+        name: constants
+        path: doc/${{matrix.detector_config}}_constants.out
         if-no-files-found: error
 
   dump-parameter-table:
     runs-on: ubuntu-latest
     needs: build-gcc
+    strategy:
+      matrix:
+        detector_config: [epic_arches, epic_brycecanyon]
```
c-dilks, line 313: and here?

wdconinc: For this one I'm pretty sure it wouldn't work because e.g. drich_only doesn't define a lot of constants that are used in the table template.

.github/workflows/linux-eic-shell.yml
wdconinc: Good catch.

.github/workflows/linux-eic-shell.yml
wdconinc: We could indeed do this here. At this point I'm trying to limit the number of matrix jobs due to the maximum of 200 simultaneous jobs in GitHub (which we routinely hit) and the practical matter of the interface being hard to navigate (I'm working on addressing that using workflow_call). Do you think it would be of use to get for example drich_only_constants.out and/or toml?

c-dilks: My bad, I hadn't realized this would get all the detectors, rather than just the two overall configs.

c-dilks:
My favorite `view1_top.png` artifact looks wrong, but that's certainly not this PR's fault. It used to show the full EPIC detector, but now it's zoomed in on part of the beam pipe. Anyway, now that we have a website we can make the views look better later.

c-dilks (approved):

wdconinc:
> My favorite `view1_top.png` artifact looks wrong, but that's certainly not this PR's fault.

Fixed in https://github.com/eic/epic/pull/345

PR #351, opened 2023-01-13
dRICH large sensor for focal point region studies

cpecar:
Adds two new optics debugging mode for dRICH:
   4 - creates large photosensor instead of default dRICH sensors, designed to catch all possible reflected photons. This is currently used with test 12 / 14 in drich-devs, throwing parallel beams of photons in order to determine the focal region of the dRICH optics.
   5 - from a local text file, creates representations of the calculated focal points from mode 4 and a script in drich-devs, and also creates default dRICH sensors to visually compare calculated focal point to sensor position.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No. 
No.

### DIFF
--- a/compact/pid/drich.xml
+++ b/compact/pid/drich.xml
@@ -44,7 +44,7 @@
 - `DRICH_debug_mirror`:    1 = draw full mirror shape for single sector; 0 = off
 - `DRICH_debug_sensors`:   1 = draw full sensor sphere for a single sector; 0 = off
 </comment>
-<constant name="DRICH_debug_optics"    value="5"/>
+<constant name="DRICH_debug_optics"    value="0"/>
 <constant name="DRICH_debug_mirror"    value="0"/>
 <constant name="DRICH_debug_sensors"   value="0"/>
 </define>
--- a/compact/pid/drich.xml
+++ b/compact/pid/drich.xml
@@ -38,11 +38,13 @@
 - `DRICH_debug_optics`:    1 = all components become vacuum, except for mirrors; test opticalphotons from IP
                            2 = all components become vacuum, except for mirrors and `gasvol`, test charged particles from IP
                            3 = all components become vacuum, except for mirrors and sensors, test reflected trajectories' hits
+			   4 = all components become vacuum, except for mirrors and sensors, oversized sensor created to catch all photons
+                           5 = all components created normally, calculated focal points from mode 4 drawn if environment variable set
                            0 = off
 - `DRICH_debug_mirror`:    1 = draw full mirror shape for single sector; 0 = off
 - `DRICH_debug_sensors`:   1 = draw full sensor sphere for a single sector; 0 = off
 </comment>
-<constant name="DRICH_debug_optics"    value="0"/>
+<constant name="DRICH_debug_optics"    value="5"/>
 <constant name="DRICH_debug_mirror"    value="0"/>
 <constant name="DRICH_debug_sensors"   value="0"/>
 </define>
--- a/src/DRICH_geo.cpp
+++ b/src/DRICH_geo.cpp
@@ -17,6 +17,7 @@
 #include "DD4hep/Printout.h"
 #include "DDRec/DetectorData.h"
 #include "DDRec/Surface.h"
+#include "DD4hep/CartesianGridXY.h"
 
 #include <XML/Helper.h>
 
@@ -124,6 +125,12 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
     case 3:
       vesselMat = aerogelMat = filterMat = gasvolMat = desc.material("VacuumOptical");
       break;
+    case 4:
+      vesselMat = aerogelMat = filterMat = gasvolMat = desc.material("VacuumOptical");
+      break;
+    case 5:
+      vesselMat = aerogelMat = filterMat = gasvolMat = desc.material("VacuumOptical");
+      break;
     default:
       printout(FATAL, "DRICH_geo", "UNKNOWN debugOpticsMode");
       return det;
@@ -131,6 +138,17 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
     aerogelVis = sensorVis = mirrorVis;
     gasvolVis = vesselVis = desc.invisible();
   };
+  
+  // sensor size rescaling for large sensor (catching all photons)
+  double sensorRescale = 0;
+  if (debugOpticsMode == 4) sensorRescale = 250;
+  if (sensorRescale > 0) {
+    auto seg = (CartesianGridXY) desc.readout(readoutName).segmentation();
+    seg.setGridSizeX(sensorRescale * seg.gridSizeX());
+    seg.setGridSizeY(sensorRescale * seg.gridSizeY());
+    seg.setOffsetX(sensorRescale * seg.offsetX());
+    seg.setOffsetY(sensorRescale * seg.offsetY());
+  }
 
   // readout coder <-> unique sensor ID
   /* - `sensorIDfields` is a list of readout fields used to specify a unique sensor ID
@@ -191,7 +209,7 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
   //  extra solids for `debugOptics` only
   Box vesselBox(1001, 1001, 1001);
   Box gasvolBox(1000, 1000, 1000);
-
+  
   // choose vessel and gasvol solids (depending on `debugOpticsMode` (0=disabled))
   Solid vesselSolid, gasvolSolid;
   switch (debugOpticsMode) {
@@ -201,6 +219,14 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
     break; // `!debugOptics`
   case 1:
   case 3:
+    vesselSolid = vesselBox;
+    gasvolSolid = gasvolBox;
+    break;
+  case 4:
+    vesselSolid = vesselBox;
+    gasvolSolid = gasvolBox;
+    break;
+  case 5:
     vesselSolid = vesselBox;
     gasvolSolid = gasvolBox;
     break;
@@ -301,17 +327,21 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
 
   // radiator z-positions (w.r.t. IP)
   double aerogelZpos = vesselPos.z() + aerogelPV.position().z();
-  double airgapZpos  = vesselPos.z() + airgapPV.position().z();
-  double filterZpos  = vesselPos.z() + filterPV.position().z();
-  desc.add(Constant("DRICH_aerogel_zpos", std::to_string(aerogelZpos)));
-  desc.add(Constant("DRICH_airgap_zpos", std::to_string(airgapZpos)));
-  desc.add(Constant("DRICH_filter_zpos", std::to_string(filterZpos)));
+  double airgapZpos, filterZpos;
+  if (!debugOptics) {
+    airgapZpos  = vesselPos.z() + airgapPV.position().z();
+    desc.add(Constant("DRICH_airgap_zpos", std::to_string(airgapZpos)));
+    filterZpos  = vesselPos.z() + filterPV.position().z();
+    desc.add(Constant("DRICH_filter_zpos", std::to_string(filterZpos)));
+  };
 
+  desc.add(Constant("DRICH_aerogel_zpos", std::to_string(aerogelZpos)));
   // radiator material names
   desc.add(Constant("DRICH_aerogel_material", aerogelMat.ptr()->GetName(), "string"));
   desc.add(Constant("DRICH_airgap_material", airgapMat.ptr()->GetName(), "string"));
   desc.add(Constant("DRICH_filter_material", filterMat.ptr()->GetName(), "string"));
   desc.add(Constant("DRICH_gasvol_material", gasvolMat.ptr()->GetName(), "string"));
+  
 
   // SECTOR LOOP //////////////////////////////////////////////////////////////////////
 
@@ -321,9 +351,8 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
   // double sensorCentroidX = 0;
   // double sensorCentroidZ = 0;
   // int    sensorCount     = 0;
-
+  
   for (int isec = 0; isec < nSectors; isec++) {
-
     // debugging filters, limiting the number of sectors
     if ((debugMirror || debugSensors || debugOptics) && isec != 0)
       continue;
@@ -455,7 +484,7 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
     auto sensorSphPos = Position(sensorSphCenterX, 0., sensorSphCenterZ) + originFront;
 
     // sensitivity
-    if (!debugOptics || debugOpticsMode == 3)
+    if (!debugOptics || debugOpticsMode == 3 || debugOpticsMode == 4)
       sensorVol.setSensitiveDetector(sens);
 
     // reconstruction constants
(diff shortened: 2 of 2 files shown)


c-dilks:
Looks good to me. Where should we store your code to calculate the focal points? Here or in `drich-dev`?

cpecar:
The focal point determination script is already in drich-dev (PR #55 in drich-dev), so I would say there makes more sense.

c-dilks (approved):

compact/pid/drich.xml
```diff
@@ -1,5 +1,5 @@
 <!-- SPDX-License-Identifier: LGPL-3.0-or-later -->
-<!-- Copyright (C) 2022 Christopher Dilks -->
+<!-- Copyright (C) 2022 Christopher Dilks, Connor Pecar -->
```
wdconinc, line 2:
```suggestion
<!-- Copyright (C) 2023 Christopher Dilks, Connor Pecar -->
```

src/DRICH_geo.cpp
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: LGPL-3.0-or-later
-// Copyright (C) 2022 Christopher Dilks, Junhuai Xu
+// Copyright (C) 2022 Christopher Dilks, Junhuai Xu, Connor Pecar
```
wdconinc, line 2:
```suggestion
// Copyright (C) 2023 Christopher Dilks, Junhuai Xu, Connor Pecar
```

src/DRICH_geo.cpp
```diff
@@ -546,29 +574,94 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
...
+      sensorSkin.isValid();
+    }; // end large sensor placement
+
+    // for debugOpticsMode 5: drawing calculated focal points
+    // of thrown beams of parallel photons
+    if( std::getenv("FPPLOT_FILE")!=NULL && debugOpticsMode == 5){
+      //std::cout << "drawing focal points " << std::endl;
+      Cone FPsolid(1.,0.25,0.5,1,1);
+      Volume FPvol(detName + "_FPpos_" + secName, FPsolid, aerogelMat);
+      FILE * fptxt = fopen(std::getenv("FPPLOT_FILE"),"r");
```
wdconinc, line 632: Please avoid C-style file reading. Please use C++ iostreams.

src/DRICH_geo.cpp
```diff
@@ -546,29 +574,94 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
...
+      DetElement sensorDE(det, Form("sensor_de%d_%d", isec, imod), 0);
+      sensorDE.setPlacement(sensorPV);
+
+      SkinSurface sensorSkin(desc, sensorDE, Form("sensor_optical_surface%d", isec), sensorSurf, sensorVolScaled);
+      sensorSkin.isValid();
+    }; // end large sensor placement
+
+    // for debugOpticsMode 5: drawing calculated focal points
+    // of thrown beams of parallel photons
+    if( std::getenv("FPPLOT_FILE")!=NULL && debugOpticsMode == 5){
```
wdconinc, line 628: Please avoid environment variables as an interface to specifying information. It is not manifest when running and leads to problems with reproducibility of results. Suggestion is to create a setting in the xml interface so the file name is stored.

src/DRICH_geo.cpp
```diff
@@ -546,29 +574,94 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
...
+      //std::cout << "drawing focal points " << std::endl;
+      Cone FPsolid(1.,0.25,0.5,1,1);
+      Volume FPvol(detName + "_FPpos_" + secName, FPsolid, aerogelMat);
+      FILE * fptxt = fopen(std::getenv("FPPLOT_FILE"),"r");
+      double fpx, fpy, fpz, dirx, diry, dirz;
+      int fpnum=0;
+
+      while(fscanf(fptxt,"%lf %lf %lf %lf %lf %lf", &fpx, &fpy, &fpz, &dirx, &diry, &dirz)!=EOF){
```
wdconinc, line 636: Please avoid files with 'just a bunch of numbers'. Use a file format that provides meta-data as part of the file. This could be xml as already supported by dd4hep, or it could be a json file. That will allow you to use more robust parameter files with key-value pairs.

src/DRICH_geo.cpp
```diff
@@ -546,29 +574,94 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
...
+          double xrot = std::atan2(sqrt(diry*diry+dirx*dirx),abs(dirz));
+	  xrot = 0;
+	  zrot = 0;
+          auto FPPlacement =
+            RotationZ(0) *
+            Translation3D(fpx, fpy, fpz-vesselPos.z())*
+            RotationX(xrot)*
+            RotationZ(zrot)
+            ;
+          auto FPPV = gasvolVol.placeVolume(FPvol, FPPlacement);
```
wdconinc, line 656: Variable names which are all-caps are considered reserved for preprocessor definitions, absolute constants, and environment variable names.

src/DRICH_geo.cpp
```diff
@@ -546,29 +574,94 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
               Translation3D(sensorSphRadius, 0., 0.) * // push radially to spherical surface
               RotationY(M_PI / 2) *                    // rotate sensor to be compatible with generator coords
               RotationZ(-M_PI / 2);                    // correction for readout segmentation mapping
-          auto sensorPV = gasvolVol.placeVolume(sensorVol, sensorPlacement);
+	    auto sensorPV = gasvolVol.placeVolume(sensorVol, sensorPlacement);
```
wdconinc, line 577: Avoid using tabs (and set your editor to insert spaces instead). Tab width is not necessarily the same on every system.

src/DRICH_geo.cpp
```diff
@@ -491,52 +518,53 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
...
-          // };
-
-          // placement (note: transformations are in reverse order)
-          // - transformations operate on global coordinates; the corresponding
-          //   generator coordinates are provided in the comments
-          auto sensorPlacement =
+    if(debugOpticsMode != 4){
+      for (int t = 0; t < (int)(nTheta + 0.5); t++) {
+	double thetaGen = t / ((double)nTheta) * M_PI;
+
```
wdconinc, line 524: Avoid C-style casts. In particular, integers are upcast to double when needed. It may be that `t` does not have to be an integer here; `for` loops work fine with non-integer types too.

src/DRICH_geo.cpp
```diff
@@ -455,7 +482,7 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
     auto sensorSphPos = Position(sensorSphCenterX, 0., sensorSphCenterZ) + originFront;
 
     // sensitivity
-    if (!debugOptics || debugOpticsMode == 3)
+    if (!debugOptics || debugOpticsMode == 3 || debugOpticsMode == 4)
```
wdconinc, line 485: Avoid relying implicitly on the boolean conversion of integers. Suggestion: use `debugOptics != 0`.

wdconinc: This applies in multiple places.

src/DRICH_geo.cpp
```diff
@@ -204,6 +222,14 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
...
     gasvolSolid = gasvolBox;
     break;
+  case 4:
+    vesselSolid = vesselBox;
+    gasvolSolid = gasvolBox;
+    break;
+  case 5:
+    vesselSolid = vesselBox;
+    gasvolSolid = gasvolBox;
+    break;
```
wdconinc, line 232: Identical case blocks can be combined for readability:
```suggestion
  case 4:
  case 5:
    vesselSolid = vesselBox;
    gasvolSolid = gasvolBox;
    break;
```

wdconinc: This may be applied beyond the context range where github allows suggested changes.

src/DRICH_geo.cpp
```diff
@@ -132,6 +139,17 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
...
 
+  // sensor size rescaling for large sensor (catching all photons)
+  double sensorRescale = 0;
+  if (debugOpticsMode == 4) sensorRescale = 250;
+  if (sensorRescale > 0) {
+    auto seg = (CartesianGridXY) desc.readout(readoutName).segmentation();
+    seg.setGridSizeX(sensorRescale * seg.gridSizeX());
+    seg.setGridSizeY(sensorRescale * seg.gridSizeY());
+    seg.setOffsetX(sensorRescale * seg.offsetX());
+    seg.setOffsetY(sensorRescale * seg.offsetY());
```
wdconinc, line 150: `seg` will go out of scope and disappear. What is its intention?

src/DRICH_geo.cpp
```diff
@@ -124,6 +125,12 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
     case 3:
       vesselMat = aerogelMat = filterMat = gasvolMat = desc.material("VacuumOptical");
       break;
+    case 4:
+      vesselMat = aerogelMat = filterMat = gasvolMat = desc.material("VacuumOptical");
+      break;
+    case 5:
+      vesselMat = aerogelMat = filterMat = gasvolMat = desc.material("VacuumOptical");
+      break;
```
wdconinc, line 133: Combine `case` statements with identical treatment. See also below.

src/DRICH_geo.cpp
```diff
@@ -132,6 +139,17 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
     gasvolVis = vesselVis = desc.invisible();
   };
 
+  // sensor size rescaling for large sensor (catching all photons)
+  double sensorRescale = 0;
+  if (debugOpticsMode == 4) sensorRescale = 250;
+  if (sensorRescale > 0) {
+    auto seg = (CartesianGridXY) desc.readout(readoutName).segmentation();
```
wdconinc, line 146: Avoid C-style casts. In this case the behavior is unspecified when a segmentation is returned that cannot be cast into `CartesianGridXY`.

c-dilks:
@cpecar I updated this branch w.r.t. `main` and resolved conflicts (hopefully correctly).

src/DRICH_geo.cpp
cpecar: It's created to alter the sensor grid size when creating an unrealistically large photosensor, as dd4hep crashes when running with the regular small grid size on a very large sensor. `seg` is then only needed in this scope to change those readout parameters. This fix was added by Chris, so he might have some better insight.

wdconinc: Do you even get non-const access to the segmentation? I really wouldn't change the segmentation from what's in the xml.

c-dilks: It's a workaround, since the sensor size is altered in this `cpp` file (from what is set in the `xml` file). We don't care about segmentation for this giant sensor, we just want to catch all the photons' real hit positions here. If we leave the segmentation as is, it will try to make 3x3 mm pixels on a giant sensor (scale of meters), overflowing the readout bit fields. Can we just turn off segmentation somehow if `debugOpticsMode == certain number(s)`? We don't run reconstruction based on this optics 'debugging' mode.

src/DRICH_geo.cpp
cpecar: How high of a priority is this? Just because it would require changes to the other scripts as well, and anything demonstrating usage of this feature would make it clear what script/simulation setting in `drich-dev` generates these points.

wdconinc: If I say that it's not a high priority, it won't ever get done, right?

wdconinc: How about the following minimal requirements?
- The file reader must be able to ignore comment (full) lines anywhere in the file starting with a comment character `#`, so a preceding line in any file can (should) contain a description of the fields.
- The file reading cannot use C style `FILE` access but must use `ifstream`.

src/DRICH_geo.cpp
```diff
@@ -204,6 +220,11 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
     vesselSolid = vesselBox;
     gasvolSolid = gasvolBox;
     break;
```
wdconinc, line 222: Combine?
```suggestion
```

src/DRICH_geo.cpp
```diff
@@ -124,6 +125,10 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
     case 3:
       vesselMat = aerogelMat = filterMat = gasvolMat = desc.material("VacuumOptical");
       break;
```
wdconinc, line 127: Combine?
```suggestion
```

src/DRICH_geo.cpp
```diff
@@ -204,6 +220,11 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
...
     gasvolSolid = gasvolBox;
     break;
+  case 4:
+  case 5:
+    vesselSolid = vesselBox;
+    gasvolSolid = gasvolBox;
+    break;
   case 2:
     vesselSolid = vesselBox;
     gasvolSolid = gasvolUnion;
```
wdconinc, line 230: Add `default` case (even though already handled, but code above may change and leave this as the first `switch`).

src/DRICH_geo.cpp
```diff
@@ -319,9 +341,8 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
   // int    sensorCount     = 0;
 
   for (int isec = 0; isec < nSectors; isec++) {
-
     // debugging filters, limiting the number of sectors
-    if ((debugMirror || debugSensors || debugOptics) && isec != 0)
+    if ((debugMirror || debugSensors || debugOpticsMode!=0) && isec != 0)
```
wdconinc, line 345:
```suggestion
    if ((debugMirror || debugSensors || debugOpticsMode != 0) && isec != 0)
```

src/DRICH_geo.cpp
```diff
@@ -487,84 +508,154 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
...
+	    // increment sensor module number
+	    imod++;
+
+	  } // end patch cuts
+	}   // end phiGen loop
+      }     // end thetaGen loop
+    }       // end if(debugOpticsMode != 4)
+
+    // large sensor placement (for focal point testing):
+    if (debugOpticsMode == 4){
```
wdconinc, line 593: `else` ?

src/DRICH_geo.cpp
wdconinc: Ok, I get this now. Resolving.

wdconinc:
@cpecar @c-dilks What is the status on this PR? Do we want to complete this before end of March to include in campaign?

c-dilks:
> @cpecar @c-dilks What is the status on this PR? Do we want to complete this before end of March to include in campaign?

This isn't needed for the campaign; it supports an ongoing study for sensor placement optimization.

@cpecar there are still some review comments to resolve.

wdconinc:
Is this branch still under development? Can you rebase, fix merge conflicts, and comment when it's ready?

wdconinc:
> Is this branch still under development?

Adding @chchatte92. Can you pick this up if needed, or close if not needed?

cpecar:
Sorry, I should have finished this a while ago and the PR has fallen behind a a good bit in terms of how the sensors are now defined. I’ll wait for word from Chandra on if this would still be a useful feature and if so I’ll bring it up to date and finalize it.

rahmans1:
@cpecar @chchatte92 We are tagging a new release for February simulation campaign? What's the status of this PR? Is it supposed to be merged or no?

PR #369, opened 2023-02-18
fix: set dd4hep compiler flags to suppress ABI warnings on arm

wdconinc:
As of version 1.24, DD4hep exports `-Wno-psabi` in its compiler flags to suppress gcc ABI warnings in the DD4hep header Shapes.h. This PR has us pick up these upstream compiler flags to pick this up. This means:
`-Wshadow -Wformat-security -Wno-long-long -Wdeprecated -fdiagnostics-color=auto -Wall -Wextra -pedantic -Wno-psabi`

- [x] Bug fix (issue: too much spurious output on arm)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -95,7 +95,7 @@ jobs:
         platform-release: "jug_xl:nightly"
         run: |
           CC=${{ matrix.CC }} CXX=${{ matrix.CXX }} cmake -B build -S . -DCMAKE_INSTALL_PREFIX=${PREFIX} -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
-          cmake --build build -- -j 2 install
+          cmake --build build -- -k -j 2 install
     - uses: actions/upload-artifact@v3
       with:
         name: build-${{ matrix.CC }}-full-eic-shell
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,6 +42,7 @@ find_package(fmt REQUIRED)
 set(a_lib_name ${PROJECT_NAME})
 
 dd4hep_configure_output()
+dd4hep_set_compiler_flags()
 
 file(GLOB sources CONFIGURE_DEPENDS src/*.cpp)
 dd4hep_add_plugin(${a_lib_name}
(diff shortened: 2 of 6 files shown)


wdconinc:
This will require some other fixes first...

.github/workflows/linux-eic-shell.yml
```diff
@@ -95,7 +95,7 @@ jobs:
         platform-release: "jug_xl:nightly"
         run: |
           CC=${{ matrix.CC }} CXX=${{ matrix.CXX }} cmake -B build -S . -DCMAKE_INSTALL_PREFIX=${PREFIX} -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
-          cmake --build build -- -j 2 install
+          cmake --build build -- -k -j 2 install
```
veprbl, line 98: What is `-k`?
https://cmake.org/cmake/help/git-master/manual/cmake.1.html
Please, use long option, if available.

wdconinc: The options after `--` are passed to the build system; they are not CMake options. The build system can be either `ninja` or `make` (and I think we prefer `ninja` in CI because it is faster than `make`) Both support `-k`, but only `make` has the longer version `--keep-going`; `ninja` has no long options. The purpose of this option is essentially the opposite of "fail fast", and it keeps compiling files (that aren't dependent on failed files) so errors in as many source files as possible are made visible in a single pass.

wdconinc: We don't specify `ninja` here. We can and probably should with `cmake -G ninja` (also no long form).

wdconinc: My statements on relative speed of `ninja` versus `make` are not based on my own tests or experience but on what for example dd4hep uses in its source-based distribution through spack.

wdconinc: So, long story short (and many words about one letter): yes, we should prefer long options always, but in this case the long option is not currently supported by one build system choice which we/dd4hep/cmake may now/later use by default.

veprbl: Just realized that myself, the `--` was a clue. I agree, we want `--keep-going` in CI.

veprbl (approved):

PR #374, opened 2023-02-24
Change Aerogel radius to 90 cm and retuning optics

chchatte92:
Optics re-tuning and reshaping aerogel.

- Optics re-tuning to achieve ATHENA like acceptance
- Aerogel size reduced to 90 cm

No 
No


compact/pid/drich.xml
```diff
@@ -15,10 +15,10 @@
 <constant name="DRICH_window_thickness"   value="0.1*cm"/>  <!-- thickness of entrance and exit walls -->
 <constant name="DRICH_num_sectors"        value="6"/>  <!-- number of azimuthal sectors -->
 <!-- snout geometry: cone with front radius rmax0 and back radius of rmax1 -->
-<constant name="DRICH_rmax0"              value="110.0*cm"/>
+<constant name="DRICH_rmax0"              value="90.0*cm"/>
 <constant name="DRICH_snout_length"       value="20.0*cm"/>
 <constant name="DRICH_snout_slope"        value="DRICH_rmax0 / DRICH_zmin"/> <!-- TODO: increase slope to allow more space for aerogel cones ? -->
-<constant name="DRICH_rmax1"              value="DRICH_rmax0 + DRICH_snout_length * DRICH_snout_slope"/>
+<constant name="DRICH_rmax1"              value="DRICH_rmax0 + DRICH_snout_length *(0.20+ DRICH_snout_slope)"/>
```
c-dilks, line 21: Small suggestion, just to prevent my own future confusion: remove this extra `DRICH_snout_slope` variable, since with the extra 200 mrad it is no longer the real slope. And we can remove the `TODO`:
```xml
<constant name="DRICH_rmax1"              value="DRICH_rmax0 + DRICH_snout_length *(0.20 + DRICH_rmax0 / DRICH_zmin")"/>
```

c-dilks (changes requested):

wdconinc (commented):
LGTM but deferring to @c-dilks

compact/pid/drich.xml
chchatte92: This is a good idea. Have you already implemented?

chchatte92:
Dear all, I see that there is an overlap with the Barrel Calorimeter. Is it possible to understand where exactly the overlap is happening? An image or something like that would be very helpful. 
Cheers,
Chandra

wdconinc:
The overlap is in the barrel HCal due to a bug in geant4. It doesn't apply here.

compact/pid/drich.xml
c-dilks: No I haven't, please go ahead

chchatte92: For example if we don't take the tangent into account the DRICH rmax for (90 cm, 20 cm) scenario will be 94 cm instead of 105 cm. This is substantially different.

c-dilks (approved):

PR #376, opened 2023-02-24
feat: use asymmetry tracker region in secondary particle handler

wdconinc:
Due to limitations in DD4hep we were using symmetric limits here, but DD4hep 1.24 allows asymmetric regions. This defines the "tracker region" which is the region where created secondaries will be stored in the MCParticles collection. This setting is backwards compatible in that `tracker_region_zmax` will be used as the symmetric limit for older DD4hep versions.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/compact/definitions.xml
+++ b/compact/definitions.xml
@@ -611,6 +611,7 @@ Service gaps in FW direction (before endcapP ECAL) and BW direction (before endc
     </comment>
 
     <constant name="tracker_region_rmax"   value="EcalBarrel_rmin"/>
-    <constant name="tracker_region_zmax"   value="EcalEndcapN_zmin"/>
+    <constant name="tracker_region_zmax"   value="EcalEndcapP_zmin"/>
+    <constant name="tracker_region_zmin"   value="EcalEndcapN_zmin"/>
 
   </define>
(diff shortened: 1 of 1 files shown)


wdconinc:
@c-dilks I think you mostly run without any particle handler, but I think this mostly affects the dRICH/pfRICH groups. The intent here is to store secondaries (including photons) for the entire asymmetric tracker region up to the front face of the Ecal in either endcap.

c-dilks:
> I think you mostly run without any particle handler, but I think this mostly affects the dRICH/pfRICH groups.

Yes, `npsim --part.userParticleHandler=''` to force the inclusion of Cherenkov `MCParticle` photons (which are not _necessary_, but extremely useful in algo development). Let me know if/when we no longer need to run `npsim` with this option.

c-dilks:
Looks good. Only change in the constants is:
```diff
< tracker_region_zmax            =      174.000 = EcalEndcapN_zmin
---
> tracker_region_zmax            =      329.600 = +EcalEndcapP_zmin
> tracker_region_zmin            =     -174.000 = -EcalEndcapN_zmin
```

c-dilks (approved):

wdconinc:
> Let me know if/when we no longer need to run npsim with this option.

Not sure it will change. We can override the default particle handler, but I am not sure if that makes sense for all contexts.

PR #377, opened 2023-02-24
feat: explicitly define a world volume and material

wdconinc:
By defining our own world volume, we gain some flexibility in defining differently shaped world volumes, e.g. a central detector hall with far forward and far backward tunnels. It also may allow us to treat the world as a geant4 region with association user limit setters.

@c-dilks This has required moving the `optical_materials.xml` inclusion outside the `<includes>` tag. When the file is included inside the `<includes>` tag, it requires that the world is created at that time since it is treated as a new translation unit. When the file is included with a simple `<include>` tag, it includes the contents in-place in the current translation unit. Can you please double-check that the change to the optical_materials include line does not affect optical properties, in particular your optical surface definitions?

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators @c-dilks 

No.

No.

### DIFF
--- a/compact/definitions.xml
+++ b/compact/definitions.xml
@@ -2,14 +2,17 @@
 <!-- Copyright (C) 2022 Whitney Armstrong, Sylvester Joosten, Wouter Deconinck, Zhenyu Ye -->
 
   <define>
-    <constant name="world_side" value="30*m"/>
-    <constant name="world_x" value="world_side"/>
-    <constant name="world_y" value="world_side"/>
-    <constant name="world_z" value="100*m"/>
-
     <constant name="Pi" value="3.14159265359"/>
     <constant name="mil" value="0.0254*mm"/>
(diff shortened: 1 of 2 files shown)


dhevang:
I did a quick check based on angle smearing of electrons flowing through the world volume in the far-bwd region.  One study had the default Air world volume.  The other had Vacuum.  I see a very clear difference.  So, it worked for me.  
Just one sanity check though.  I see the "dz" of the world volume is 100 m.  I want to make sure that this is the half-width of the world.  My study was at -65 m.

wdconinc:
`dz` is half-width indeed. It would show overlaps for placements outside of the world volume.

c-dilks (approved):

dhevang (approved):

PR #382, opened 2023-03-06
feat: import full HRPPD GDML as mRICH photo-detector

wdconinc:
Instead of implementing a detailed HRPPD model in xml with support in the geometry plugin, this just pulls in the GDML model from @alexander-kiselev.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators @osbornjd @msar15

No.

No.


wdconinc:
Looks like this (only one box shown with children):
image

wdconinc:
@c-dilks Asked you to review this. Maybe there are useful approaches here for the dRICH? I think you should be able to get volumes by name out of the gdml, and then assign them sensor/module/segmentation. It is possibly a good 'CAD-light' approach.

alexander-kiselev:
Perhaps. In a different context (and in a simple model case, where one can actually code them in OpenCascade directly) I'm using a different approach: have a CAD model and a gerber set for the matching PCB production done in the same ROOT script, in which case both models are automatically synchronized in terms of shapes, sizes, hole locations and such, as long as one does not make a coding mistake. It is obvious one can do the same for a CAD+GEANT output rather than a CAD+PCB combo. Essentially one is generating a GDML file and a CAD model of identical complexity at the same time. But then one can argue, why not use ROOT->CAD export, which is readily available. Anyway, my 2 cents.

c-dilks:
Looks great, however there are some mRICH-mRICH overlaps, e.g.:
```
Overlap is detected for volume MRICH_module1_39:39 (G4Trd) with MRICH_module1_35:35 (G4Trd)
```
```
Overlap is detected for volume MRICH_module1_59:59 (G4Trd) with its mother volume MRICH_Envelope (G4Tubs)
```

alexander-kiselev:
not sure what we are supposed to see here. looks like mRICH embedded into pfRICH vessel?

________________________________

wdconinc:
This branch (obviously) needs work since we changed the default to pfrich. There is no real value in having the mrich implemented in more detail, but there is value in having a proof of concept of having an imported gdml component to point to as an example.

wdconinc:
There are now several better examples of importing gdml in production code, so I'm going to close this.

Issue #383, opened 2023-03-07
DDG4 should be made optional again


faustus123:
**Is your feature request related to a problem? Please describe.**
Building on systems where only reconstruction is done, Geant4 is not needed. The `DDG4` module is not built by `DD4HEP` if Geant4 is not present. This repository should make it optional to include the `DDG4` module since it is not needed for building the reconstruction software.

This was [fixed at one point](https://github.com/eic/epic/commit/367894adbe22577815e9419b9ce26aaaf1e77d13) but later [removed over a dispute on licensing](https://github.com/eic/epic/pull/285).

**Describe the solution you'd like**
DDG4 should be made an optional component so `epic` can be built when Geant4 is not available.

**Describe alternatives you've considered**
None.

**Additional context**
None

PR #384, opened 2023-03-07
cmake: drop DDG4 dependency

veprbl:

Resolves: #383

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No


wdconinc:
DDG4 isn't needed by the src files, but it is required for several of the scripts that are installed. We can tweak this by fine-grained installs based on whether DDG4 (which is a bit of a time-consuming exercise), or we can install all scripts even those that we know won't run. What do you think?

veprbl:
Which scripts, though? They are not part of CI?

veprbl:
You mean that the point of `find_package(... DDG4)` is to check that DD4hep was built with its support? Shouldn't the script be able to crash and tell you, basically, the same?

wdconinc:
We don't actually install the scripts with cmake (my bad), so this is fine: we don't make any guarantees we can't keep.

PR #387, opened 2023-03-09
Add geometry configuration files for img calo

mariakzurek:
- [ ] Bug fix (issue #__)
- [x] New feature 
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators @Chao1009 
- [x] Tested locally (EICrecon + dd4sim particle gun)

3 Geometry Implementation:
9layers
6layers
4layers

1 Implementation with kB constant = 0.079 MeV/mm


compact/ecal/barrel_interlayers_9layers.xml
```diff
@@ -0,0 +1,192 @@
+<!-- SPDX-License-Identifier: LGPL-3.0-or-later -->
+<!-- Copyright (C) 2022 Whitney Armstrong, Chao Peng, Maria Zurek, Jihee Kim -->
```
wdconinc, line 2: This file is identical to `barrel_interlayers.xml`, except for `EcalBarrelImagingLayers_nMax` which is 9 instead of 6. This feels like we should approach this differently (under the "don't repeat yourself" mantra), and this is an instructive example to start the discussion with.

In this case, we could separate out the definition of  `EcalBarrelImagingLayers_nMax`, and put it in a file `barrel_interlayers_6layers.xml` with just this:
```xml
<lccdd>
  <define>
     <constant name="EcalBarrelImagingLayers_nMax"   value="6"/>
  </define>
</lccdd>
```
and similar for the `barrel_interlayers_9layers.xml`.

The current file `barrel_interlayers.xml` would become something like `barrel_interlayers_impl.xml` and would not have any `EcalBarrelImagingLayers_nMax` definition.

compact/ecal/barrel_interlayers_birks.xml
```diff
@@ -0,0 +1,205 @@
+<!-- SPDX-License-Identifier: LGPL-3.0-or-later -->
+<!-- Copyright (C) 2022 Whitney Armstrong, Chao Peng, Maria Zurek, Jihee Kim -->
+
```
wdconinc, line 3: Now we move on to the birks file, which only adds a new material and uses it instead, but keeps the rest of the structure of `barrel_interlayers.xml` unchanged. Again, this makes sense to put in a separate file, which now would contain:
```xml
<lccdd>
  <define>
     <constant name="EcalBarrelImagingLayers_nMax"   value="6"/>
     <constant name="EcalBarrelImagingLayers_FiberMaterial" value="SciFiPb_Scintillator_Birks"/>
  </define>
</lccdd>
```
The definition of the material itself should be placed in the overall materials file, and likely be given a name which includes the actual Birks value (it's only a matter of time before someone wants to try a differen value), so maybe `SciFiPb_Scintillator_Birks0079`.

wdconinc:
Can we close this and reopen later if necessary?

wdconinc:
@Chao1009 Can we close this, or just pick the 4-layer configuration for the June campaign since that is the default?

wdconinc:
@Chao1009 @mariakzurek I'm closing this. Reopen if needed.

Issue #389, opened 2023-03-10
Add functionality to swap electron magnet fields for different energy beams


simonge:
Allow the user to select a fields xml file for different electron beam energies, similar to the far forward implementation.

PR #390, opened 2023-03-10
389 add functionality to swap electron magnet fields for different energy beams

simonge:
Functionality to easily select beam energies and associated magnetic fields. Additionally moves the beamline geometry extensions into separate files and configuration to improve performance.

- [x] Bug fix (issue #__)
- [x] New feature (issue #389 )
- [ ] Documentation update
- [ ] Other: __

A different base xml file will need to be selected when investigating the overlaps with systems further along the beamline.

No


templates/epic.xml.jinja2
```diff
@@ -46,6 +46,8 @@
     </documentation>
     <include ref="${DETECTOR_PATH}/compact/far_forward/fields_{{ pbeam | default("275", true) }}.xml" />
     <include ref="${DETECTOR_PATH}/compact/far_forward/definitions.xml" />
+    <include ref="${DETECTOR_PATH}/compact/far_backward/fields_{{ ebeam | default("18", true) }}.xml" />
```
wdconinc, line 49: Do the fields only depend on the electron beam energy?

simonge: From the tables here:
https://indico.bnl.gov/event/10974/contributions/51260/

The fields as I understand them are just scaling with the electron energy to the first few decimal places, however there is variation on the same level when the ion beam energy is changed. If you know where a more up to date set of values it would be useful to check. I imagine these will be tweaked in the future to not just be simply scalable.

wdconinc: I mean that at this point you add fields_18.xml etc. But won't we need fields_18x275.xml at some point?

simonge: Good question, I'm not sure, the focussing might be changed slightly? Isn't it more likely the solenoid will be tuned for different energy physics.

wdconinc: No idea... @ajentsch How much will proton beam energy affect (far) forward beam magnet settings?

ajentsch: So, the magnet settings mostly scale with energy, except the B0pf (pf = proton forward) magnet, which stays constant at all beam settings.

However, this may not always be true long-term (there are changes to the lattice on the horizon for the hadron beam). 

In other words, we should avoid introducing scale factors as this stage of the games, and use the exact values from the tables (e.g. separate XML for various beam settings) to avoid possible pitfalls down the road.

If the concern is figuring out how to change the fields at runtime for the user, that's something we need to think about.

simonge: The outgoing electron magnets don't currently have different proposed fields dependant on the proton energy from some second order focussing? For the time being we could keep the forward and backwards settings in separate files rather than making one for each combination?

ajentsch: @simonge No, what you have is the current official set of tables, and what everyone should be using until there is an official new release.

wdconinc: So you don't think we need something like this, for future proofness (also for far_forward)?
```suggestion
    <include ref="${DETECTOR_PATH}/compact/far_backward/fields_{{ ebeam | default("18", true) }}x{{ pbeam | default("275", true) }}.xml" />
```

simonge: They might then be better consolidated into one file in the compact/fields folder. It is also quite likely in the future the simple dipole and quadrupole fields should be replaced with maps. I don't think either implementation at this point would cause any significant difference in the work needed to make changes down the line.

dhevang:
I see that the far_backward extensions are no longer in there by default.  @simonge , you shifted this extension to far_backward_extended.xml, but this file is not loaded anywhere.  Why is this?

dhevang (changes requested):
I see that the far_backward extensions are no longer in there by default. @simonge , you shifted this extension to far_backward_extended.xml, but this file is not loaded anywhere. Why is this?

dhevang:
@wdconinc ,  a while ago you asked me to tell you when a certain issue came up.  It just came up when I switched out of this PR branch back to main and tried to compile epic.  I got errors related to lingering configuration files:
Traceback (most recent call last):
  File "/home/dhevan/eic/epic/bin/make_detector_configuration", line 26, in <module>
    with open(args.config) as f:
FileNotFoundError: [Errno 2] No such file or directory: '/home/dhevan/eic/epic/configurations/5x100.yml'

Of course I could just delete everything in the donfigurations dir and proceed, but I think some neater solution should be implemented.

simonge:
I moved the extensions out from the default configuration as they add an overhead when loading the geometry. Also for electrons carrying on down the beamline lots of secondary particles are produced in collision with the beam pipe which slows down the average event time. We could set the beampipe and magnet material to vacuum to remove these but the volumes are there as a design guide (possibly useful for second order background studies later?) in the long term they aren't a necessary part of the base simulation.

I have added the main configuration file epic_ip6_extended.xml which should add everything in as it was before. If you've a preferred way of separating this out I'm happy for suggestions.

dhevang (approved):

configurations/10x100.yml
dhevang, line 1: With this file and the other similar ones, you introduce a E_e x E_p set of yml files.  But for the fields, e.g. fields_5.xml, (and as discussed with Alex and Wouter), you have only electron beam values. 
In templates/epic.xml.jinja2, you load electron and hadron fields separately.  Would it be simple enough to replace those separated files with the "correlated" set of files like the others suggested: fields_10x100.xml?

dhevang (changes requested):

compact/far_backward/definitions.xml
```diff
@@ -1,18 +1,16 @@
 <!-- SPDX-License-Identifier: LGPL-3.0-or-later -->
 <!-- Copyright (C) 2022 Simon Gardner, Jaroslav Adam, Wouter Deconinck -->
 
-<lccdd>
```
wdconinc, line 4: You have to stay in the lccdd namespace. Why did you remove this?

compact/far_backward/definitions.xml
```diff
@@ -298,5 +294,3 @@
     <constant name="LumiDirect_DZ"        value="0.2*m"/>
 
   </define>
-
-</lccdd>
```
wdconinc, line 302: id.

compact/far_backward_extended.xml
```diff
@@ -0,0 +1,16 @@
+<!-- SPDX-License-Identifier: LGPL-3.0-or-later -->
+<!-- Copyright (C) 2022 Wouter Deconinck, Simon Gardner -->
```
wdconinc, line 2: 2023

compact/fields/beamline_10x100.xml
```diff
@@ -1,12 +1,31 @@
-<!-- SPDX-License-Identifier: LGPL-3.0-or-later -->
```
wdconinc, line 1: Don't remove copyright headers.

compact/fields/beamline_10x275.xml
```diff
@@ -1,12 +1,31 @@
-<!-- SPDX-License-Identifier: LGPL-3.0-or-later -->
```
wdconinc, line 1: id.

compact/fields/beamline_18x275.xml
```diff
@@ -0,0 +1,53 @@
+<define>
```
wdconinc, line 1: Add copyright header.

wdconinc: Also, add lccdd tag.

compact/fields/beamline_5x100.xml
```diff
@@ -0,0 +1,53 @@
+<define>
```
wdconinc, line 1: id.

compact/fields/beamline_5x41.xml
```diff
@@ -0,0 +1,53 @@
+<define>
```
wdconinc, line 1: id.

wdconinc (changes requested):

compact/far_backward/definitions.xml
simonge: It was removed to mirror the situation in the main and far_forward definitions.xml files. It might be nicer and more consistent to remove the surrounding define block from the top xml file and add lccdd back into these, allowing flexibility to nest the define files.

compact/far_backward/definitions.xml
aniakowalewska: Hello why you would like to remove the line connected to Lumi direct photon detector from definitions.xml file?

simonge: The LumiDirect_DZ constant is untouched, it is just the lccdd tag which is under discussion.

aniakowalewska: Ok, thank you for explanation.

compact/far_backward/definitions.xml
simonge: do the content of define blocks take precedence over the include blocks file content? Moving the main definitions file out of the define and adding lccdd tags to it means world_dz is not yet defined when the world block is processed.

wdconinc: Ok, let's leave as is.

compact/far_backward/definitions.xml
wdconinc: Let's leave as is.

wdconinc (approved):
Ready for merging.

simonge:
> Ready for merging.

Sorry, I had apparently broken it in the meantime in other ways. Please can you check and re-approve.

wdconinc (approved):
Looks good.

PR #395, opened 2023-03-10
ci: only run dawn slices views on main

wdconinc:
We are not able to keep up with workflows at times. While it is useful to make sure all jobs run, even on pull requests, it is more useful to make sure the visualization jobs run on main. This restricts prim file generation and dawn views to main only.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [x] Other: __

No.

No.


veprbl (approved):
I agree

c-dilks:
I like the speedup, but having `view1_top` is very useful when comparing a geometry-changing dev branch to `main` to see how things move, especially when `definitions.xml` is modified.

c-dilks:
Can we keep `view1` for PRs? I also don't mind commenting out these `if: github.ref` lines temporarily the next time someone asks me to touch `definitions.xml`.

wdconinc:
I don't find a reference for jobs.<job>.strategy.parallel.if support. We can use a quick job to create a list of views and tweak that based on PR vs main.

.github/workflows/linux-eic-shell.yml
```diff
@@ -508,6 +508,7 @@ jobs:
   generate-prim-file:
     runs-on: ubuntu-latest
     needs: build
+    if: github.ref == 'refs/heads/main'
```
wdconinc, line 511:
```suggestion
```

.github/workflows/linux-eic-shell.yml
```diff
@@ -536,6 +537,7 @@ jobs:
   dawn-view:
     runs-on: ubuntu-latest
     needs: generate-prim-file
+    if: github.ref == 'refs/heads/main'
```
wdconinc, line 540:
```suggestion
```

wdconinc (commented):
How about we just skip the dawn slices jobs? Those are largely responsible too. It would be a start and we can continue to evaluate.

c-dilks (approved):

PR #354, opened 2023-01-23
fix(ci): more robust comment of fiber/lens details

wdconinc:
### Briefly, what does this PR introduce?
Instead of a fixed number of lines drop, this does a replace of the relevant open/close tags to surround with a `comment` tag.

### What kind of change does this PR introduce?
- [x] Bug fix (issue: adding fields to fiber causes failures in prim jobs)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [x] Changes have been communicated to collaborators: @mariakzurek 

### Does this PR introduce breaking changes? What changes might users need to make to their code?
No.

### Does this PR change default behavior?
No.

PR #396, opened 2023-03-11
ci: fix commenting of cladded imaging fibers

wdconinc:
This was accidentally changed with the fiber cladding: fibers were never commented for fast overlap checks.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


wdconinc:
This hasn't led to CI errors because `sed` without arguments reads stdin and GitHub Actions does something like the following (see also `apt install` without the `-y` fails to install anything but doesn't hang):
```console
true | command 2>&1 | cat
```
This means that the following both succeed in the has-exit-code-zero sense, but don't do anything:
```console
true | sed -z 's/a/b/' | cat
true | apt install jq | cat
```

wdconinc:
@veprbl Do you have any ideas on how to fail (and whether we even want to, in general) when a read from stdin fails due to not having a pseudo-tty?

veprbl:
Are you not just missing a `-i` ?

wdconinc:
Addressed by #354.

PR #397, opened 2023-03-11
feat: update pfRICH `gdml` to `v03`

c-dilks:
Update version `v03` of the pfRICH from @alexander-kiselev

#### TODO
- [ ] verify there is no overlap with the tracker support (or anything else)

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

no

no


c-dilks:
There is still a slight overlap with the barrel TOF. If I push the pfRICH toward -z, then it bumps into the ECal, so the pfRICH z-length needs to be reduced.

The menagerie says the max length can be 54.1 cm.

c-dilks:
> Anyway, pfRICH container volume is 541.5mm . Are you saying we should fit into 541mm, and then everything will be fine?

That's what I thought, and yes 541.0 mm should be good. Could you make this change and email me the new GDML, or just push it to this branch?

> It can perfectly be that other subsystems (as described in dd4hep) do not really follow the details of their respective CAD models.

This is a general problem resulting from 3 (4?) different places with geometry specification, each claiming to be "official." We need open access to everything, including the CAD models, improved communcation, and automation to ensure consistency. The production of GDML files on the DD4hep side is fully automated and the latest ones are always available at https://eic.github.io/epic/artifacts

wdconinc:
Merged main into branch, fixed conflicts due to removal of `no_hcal` geometries.

c-dilks:
Closing in favor of https://github.com/eic/epic/pull/512.

PR #406, opened 2023-04-06
Changes for LFHCal

FriederikeBock:
- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

This PR introduces the proper implementation of the forward hadronic calorimeter (LFHCal). Accordingly the default configuration for arches... need to be updated. 

no changes to the default behavior have been introduced.


compact/display_geoviewer.xml
```diff
@@ -82,6 +82,9 @@
     <vis name="HcalBarrelStaveVis"    ref="AnlBlue"   showDaughters="false" visible="true"/>
     <vis name="HcalBarrelLayerVis"    ref="AnlBlue"   showDaughters="true"  visible="true"/>
     <vis name="HcalEndcapVis"         ref="AnlBlue"   showDaughters="false" visible="true"/>
+    <vis name="LFHCALVis"             ref="AnlRed"   showDaughters="true" visible="true"/>
+    <vis name="LFHCALLayerSteelVis"   ref="AnlBlue"   showDaughters="false" visible="true"/>
+    <vis name="LFHCALLayerTungstenVis" ref="AnlDarkBlue"   showDaughters="false" visible="true"/>
```
wdconinc, line 87: I'm surprised this is shorter. Doesn't this lead to missing visualization attribute definitions when running with this set of display definitions? In any case, unless you are using the difference in visualization attributes between geoviewer vs regular, this should go into the detector's xml file.

compact/hcal/lfhcal.xml
```diff
@@ -0,0 +1,204 @@
...
+        notchWidthAbsA="FourM_NotchWidthAbsA"
+        notchWidthAbsC="FourM_NotchWidthAbsC"
+        notchWidthScin="FourM_NotchWidthScin"
+        notchDepth="Mod_NotchDepth"
+        sepDepth="Mod_SepWidth"
+        />
+        <position x="50*cm" y="60*cm" z="LFHCAL_length - LFHCALElectronicsThickness" />
+      </fourmodule>
+
+      <layer repeat="1" vis="InvisibleWithDaughters" readoutlayer = "0">
```
wdconinc, line 127:
```suggestion
      <layer repeat="1" vis="InvisibleWithDaughters" readoutlayer="0">
```

compact/hcal/lfhcal.xml
```diff
@@ -0,0 +1,204 @@
...
+        <slice material="Tyvek" thickness="LFHCALTyvekThickness" vis="LFHCALLayerTyvekVis"
+               type="2" offset="LFHCALAirOffset"/>
+        <slice material="Polystyrene" thickness="LFHCALPolystyreneThickness" vis="LFHCALLayerScintVis" sensitive=" yes"   limits="cal_limits" type="3" />
+      </layer>
+      <layer repeat="LFHCALLayer_NTungstenRepeat" vis="InvisibleWithDaughters" readoutlayer = "0" >
```
wdconinc, line 132:
```suggestion
      <layer repeat="LFHCALLayer_NTungstenRepeat" vis="InvisibleWithDaughters" readoutlayer="0" >
```

compact/hcal/lfhcal.xml
```diff
@@ -0,0 +1,204 @@
...
+        <slice material="Steel235" thickness="LFHCALAbsorberThickness" vis="LFHCALLayerSteelVis" type="1"/>
+<!--         <slice material="Tungsten" thickness="LFHCALAbsorberThickness" vis="LFHCALLayerSteelVis" type="1"/>  -->
+        <slice material="Tyvek" thickness="LFHCALTyvekThickness" vis="LFHCALLayerTyvekVis" type="2" offset="LFHCALAirOffset"/>
+        <slice material="Polystyrene" thickness="LFHCALPolystyreneThickness" vis="LFHCALLayerScintVis" sensitive=" yes" limits="cal_limits" type="3" />
+      </layer>
+      <layer repeat="LFHCALLayer_NSteelRepeatS" vis="InvisibleWithDaughters" readoutlayer = "1">
```
wdconinc, line 138:
```suggestion
      <layer repeat="LFHCALLayer_NSteelRepeatS" vis="InvisibleWithDaughters" readoutlayer="1">
```

compact/hcal/lfhcal.xml
```diff
@@ -0,0 +1,204 @@
...
+        <slice material="Steel235" thickness="LFHCALAbsorberThickness" vis="LFHCALLayerSteelVis" type="1"/>
+        <slice material="Tyvek" thickness="LFHCALTyvekThickness" vis="LFHCALLayerTyvekVis" type="2" offset="LFHCALAirOffset"/>
+        <slice material="Polystyrene" thickness="LFHCALPolystyreneThickness" vis="LFHCALLayerScintVis" sensitive=" yes"   limits="cal_limits" type="3" />
+      </layer>
+      <layer repeat="LFHCALLayer_NSteelRepeatS" vis="InvisibleWithDaughters" readoutlayer = "2">
```
wdconinc, line 143:
```suggestion
      <layer repeat="LFHCALLayer_NSteelRepeatS" vis="InvisibleWithDaughters" readoutlayer="2">
```

compact/hcal/lfhcal.xml
```diff
@@ -0,0 +1,204 @@
...
+        notchDepth="Mod_NotchDepth"
+        sepDepth="Mod_SepWidth"
+        />
+        <position x="50*cm" y="60*cm" z="LFHCAL_length - LFHCALElectronicsThickness" />
+      </fourmodule>
+
+      <layer repeat="1" vis="InvisibleWithDaughters" readoutlayer = "0">
+        <slice material="Tyvek" thickness="LFHCALTyvekThickness" vis="LFHCALLayerTyvekVis"
+               type="2" offset="LFHCALAirOffset"/>
+        <slice material="Polystyrene" thickness="LFHCALPolystyreneThickness" vis="LFHCALLayerScintVis" sensitive=" yes"   limits="cal_limits" type="3" />
```
wdconinc, line 130:
```suggestion
        <slice material="Polystyrene" thickness="LFHCALPolystyreneThickness" vis="LFHCALLayerScintVis" sensitive="yes"   limits="cal_limits" type="3" />
```

compact/hcal/lfhcal.xml
```diff
@@ -0,0 +1,204 @@
...
+      </layer>
+      <layer repeat="LFHCALLayer_NTungstenRepeat" vis="InvisibleWithDaughters" readoutlayer = "0" >
+        <slice material="Steel235" thickness="LFHCALAbsorberThickness" vis="LFHCALLayerSteelVis" type="1"/>
+<!--         <slice material="Tungsten" thickness="LFHCALAbsorberThickness" vis="LFHCALLayerSteelVis" type="1"/>  -->
+        <slice material="Tyvek" thickness="LFHCALTyvekThickness" vis="LFHCALLayerTyvekVis" type="2" offset="LFHCALAirOffset"/>
+        <slice material="Polystyrene" thickness="LFHCALPolystyreneThickness" vis="LFHCALLayerScintVis" sensitive=" yes" limits="cal_limits" type="3" />
```
wdconinc, line 136:
```suggestion
        <slice material="Polystyrene" thickness="LFHCALPolystyreneThickness" vis="LFHCALLayerScintVis" sensitive="yes" limits="cal_limits" type="3" />
```

compact/hcal/lfhcal.xml
```diff
@@ -0,0 +1,204 @@
...
+      </layer>
+      <layer repeat="LFHCALLayer_NSteelRepeatS" vis="InvisibleWithDaughters" readoutlayer = "1">
+        <slice material="Steel235" thickness="LFHCALAbsorberThickness" vis="LFHCALLayerSteelVis" type="1"/>
+        <slice material="Tyvek" thickness="LFHCALTyvekThickness" vis="LFHCALLayerTyvekVis" type="2" offset="LFHCALAirOffset"/>
+        <slice material="Polystyrene" thickness="LFHCALPolystyreneThickness" vis="LFHCALLayerScintVis" sensitive=" yes"   limits="cal_limits" type="3" />
```
wdconinc, line 141:
```suggestion
        <slice material="Polystyrene" thickness="LFHCALPolystyreneThickness" vis="LFHCALLayerScintVis" sensitive="yes"   limits="cal_limits" type="3" />
```

compact/hcal/lfhcal.xml
```diff
@@ -0,0 +1,204 @@
...
+      </layer>
+      <layer repeat="LFHCALLayer_NSteelRepeatS" vis="InvisibleWithDaughters" readoutlayer = "2">
+        <slice material="Steel235" thickness="LFHCALAbsorberThickness" vis="LFHCALLayerSteelVis" type="1"/>
+        <slice material="Tyvek" thickness="LFHCALTyvekThickness" vis="LFHCALLayerTyvekVis" type="2" offset="LFHCALAirOffset"/>
+        <slice material="Polystyrene" thickness="LFHCALPolystyreneThickness" vis="LFHCALLayerScintVis" sensitive=" yes"   limits="cal_limits" type="3" />
```
wdconinc, line 146:
```suggestion
        <slice material="Polystyrene" thickness="LFHCALPolystyreneThickness" vis="LFHCALLayerScintVis" sensitive="yes"   limits="cal_limits" type="3" />
```

compact/hcal/lfhcal.xml
```diff
@@ -0,0 +1,204 @@
...
+      </layer>
+      <layer repeat="LFHCALLayer_NSteelRepeatS" vis="InvisibleWithDaughters" readoutlayer = "3">
```
wdconinc, line 148:
```suggestion
      <layer repeat="LFHCALLayer_NSteelRepeatS" vis="InvisibleWithDaughters" readoutlayer="3">
```

compact/hcal/lfhcal.xml
```diff
@@ -0,0 +1,204 @@
...
+        <slice material="Steel235" thickness="LFHCALAbsorberThickness" vis="LFHCALLayerSteelVis" type="1"/>
+        <slice material="Tyvek" thickness="LFHCALTyvekThickness" vis="LFHCALLayerTyvekVis" type="2" offset="LFHCALAirOffset"/>
+        <slice material="Polystyrene" thickness="LFHCALPolystyreneThickness" vis="LFHCALLayerScintVis" sensitive=" yes"   limits="cal_limits" type="3" />
```
wdconinc, line 151:
```suggestion
        <slice material="Polystyrene" thickness="LFHCALPolystyreneThickness" vis="LFHCALLayerScintVis" sensitive="yes"   limits="cal_limits" type="3" />
```

compact/hcal/lfhcal.xml
```diff
@@ -0,0 +1,204 @@
...
+      </layer>
+      <layer repeat="LFHCALLayer_NSteelRepeatS" vis="InvisibleWithDaughters" readoutlayer = "4">
```
wdconinc, line 153:
```suggestion
      <layer repeat="LFHCALLayer_NSteelRepeatS" vis="InvisibleWithDaughters" readoutlayer="4">
```

compact/hcal/lfhcal.xml
```diff
@@ -0,0 +1,204 @@
...
+        <slice material="Steel235" thickness="LFHCALAbsorberThickness" vis="LFHCALLayerSteelVis" type="1"/>
+        <slice material="Tyvek" thickness="LFHCALTyvekThickness" vis="LFHCALLayerTyvekVis" type="2" offset="LFHCALAirOffset"/>
+        <slice material="Polystyrene" thickness="LFHCALPolystyreneThickness" vis="LFHCALLayerScintVis" sensitive=" yes"   limits="cal_limits" type="3" />
```
wdconinc, line 156:
```suggestion
        <slice material="Polystyrene" thickness="LFHCALPolystyreneThickness" vis="LFHCALLayerScintVis" sensitive="yes"   limits="cal_limits" type="3" />
```

compact/hcal/lfhcal.xml
```diff
@@ -0,0 +1,204 @@
...
+      </layer>
+      <layer repeat="LFHCALLayer_NSteelRepeatS" vis="InvisibleWithDaughters" readoutlayer = "5">
```
wdconinc, line 158:
```suggestion
      <layer repeat="LFHCALLayer_NSteelRepeatS" vis="InvisibleWithDaughters" readoutlayer="5">
```

compact/hcal/lfhcal.xml
```diff
@@ -0,0 +1,204 @@
...
+        <slice material="Steel235" thickness="LFHCALAbsorberThickness" vis="LFHCALLayerSteelVis" type="1"/>
+        <slice material="Tyvek" thickness="LFHCALTyvekThickness" vis="LFHCALLayerTyvekVis" type="2" offset="LFHCALAirOffset"/>
+        <slice material="Polystyrene" thickness="LFHCALPolystyreneThickness" vis="LFHCALLayerScintVis" sensitive=" yes"   limits="cal_limits" type="3" />
```
wdconinc, line 161:
```suggestion
        <slice material="Polystyrene" thickness="LFHCALPolystyreneThickness" vis="LFHCALLayerScintVis" sensitive="yes"   limits="cal_limits" type="3" />
```

compact/hcal/lfhcal.xml
```diff
@@ -0,0 +1,204 @@
...
+      </layer>
+      <layer repeat="LFHCALLayer_NSteelRepeatS" vis="InvisibleWithDaughters" readoutlayer = "6">
```
wdconinc, line 163:
```suggestion
      <layer repeat="LFHCALLayer_NSteelRepeatS" vis="InvisibleWithDaughters" readoutlayer="6">
```

compact/hcal/lfhcal.xml
```diff
@@ -0,0 +1,204 @@
...
+        <slice material="Steel235" thickness="LFHCALAbsorberThickness" vis="LFHCALLayerSteelVis" type="1"/>
+        <slice material="Tyvek" thickness="LFHCALTyvekThickness" vis="LFHCALLayerTyvekVis" type="2" offset="LFHCALAirOffset"/>
+        <slice material="Polystyrene" thickness="LFHCALPolystyreneThickness" vis="LFHCALLayerScintVis" sensitive=" yes"   limits="cal_limits" type="3" />
```
wdconinc, line 166:
```suggestion
        <slice material="Polystyrene" thickness="LFHCALPolystyreneThickness" vis="LFHCALLayerScintVis" sensitive="yes"   limits="cal_limits" type="3" />
```

compact/hcal/lfhcal.xml
```diff
@@ -0,0 +1,204 @@
...
+      <fourmodulepositions name="xpos" unit="cm" values="-270 -250 -270 -260 -270 -250 -270 -260 -270 -270 240 260 -10 0 250 260 260 240 260 250 260 260 -10 0 -10 0 -10 0 -10 0 -10 0 -10 0 -10 0 -10 0 -10 0 -10 0 -10 0 -10 0 -10 0 -10 0 -10 0 -10 0 -10 0 -10 0 -10 0 -10 0 -10 0 -10 0 -10 0 30 0 30 -10 -20 -20"/>
+      <fourmodulepositions name="ypos" unit="cm" values="-55 -115 45 -85 -35 105 25 75 -15 5 -115 -55 -265 -265 -85 45 -35 105 25 75 -15 5 255 255 -245 -245 235 235 -225 -225 215 215 -205 -205 195 195 -185 -185 175 175 -165 -165 155 155 -145 -145 135 135 -125 -125 115 115 -105 -105 95 95 -85 -85 75 75 -65 -65 55 55 -45 -45 35 35 -15 -25 5 -25 -15 5"/>
+      <fourmodulepositions name="zpos" unit="cm" values="0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0"/>
```
wdconinc, line 191: Positions should not be specified as independent x and y lists, and units should be attached to all values not provided as a single field. This should be in the format e.g.:
```
<fourmodulepositions z="0*cm">
  ...
  <position x="-20*cm" y="5*cm"/>
</fourmodulepositions>
```
See mRICH, negative endcap ecal, etc.

compact/hcal/lfhcal.xml
```diff
@@ -0,0 +1,204 @@
...
+    </detector>
+
+  </detectors>
+
+  <!--  Definition of the readout segmentation/definition  -->
+  <readouts>
+    <readout name="LFHCALHits">
+       <segmentation type="NoSegmentation"/> -->
```
wdconinc, line 199:
```suggestion
       <segmentation type="NoSegmentation"/>
```

compact/hcal/lfhcal_insertGeom.xml
```diff
@@ -0,0 +1,206 @@
+<lccdd>
+
+  <info name="lfhcal.xml"
```
wdconinc, line 3: The insert geometry should not duplicate the geometry of the geometry without insert. This makes code management difficult. If possible, the insert geometry should be separate, with module geometry definitions in a separate file that can be read first. The insert geometry should only define what is different/additional for the insert geometry.

(I won't review the rest of this file since the same comments as for lfhcal.xml apply.)

configurations/forwardInsert.yml
```diff
@@ -0,0 +1,3 @@
+features:
```
wdconinc, line 1: Filename should explicitly include `hcal`.

configurations/forwardNew.yml
```diff
@@ -0,0 +1,19 @@
+features:
```
wdconinc, line 1: Filename should explicitly include `hcal` and not refer to "new" (this is going to be the default when we merge this, and there's nothing new about it).

src/InsertCalorimeter_geo.cpp
```diff
@@ -89,7 +89,7 @@ static Ref_t createDetector(Detector& desc, xml_h handle, SensitiveDetector sens
 
   // Assembly that will contain all the layers
   Assembly assembly(detName);
-
+  assembly.setVisAttributes(desc.visAttributes("InvisibleWithDaughters"));
```
wdconinc, line 92: This plugin is used for ecal insert too, I think. Making it invisible in all cases will cause confusion about whether it is enabled or not.
```suggestion

```

configurations/lfhcal_withInlay.yml
```diff
@@ -0,0 +1,4 @@
+features:
+  hcal:
+    lfhcal_insertGeom:
+    forward_insert:
```
wdconinc, line 4: This will cause overlaps since the forward_insert is the same as the insert.

src/LFHCAL_geo.cpp
```diff
@@ -0,0 +1,1125 @@
...
+  double      mod_height;
+  double      mod_notchWidthAbsA;
+  double      mod_notchWidthAbsB;
+  double      mod_notchWidthAbsC;
+  double      mod_notchWidthScin;
+  double      mod_notchDepth;
+  double      mod_sepDepth;
+  std::string mod_visStr;
+  std::string mod_regStr;
+  std::string mod_limStr;
```
wdconinc, line 62: For clarity and to make all fields are always initialized, can we add the default values with `{}` initializes here and get rid of the default constructor?

wdconinc (commented):
Review part 1.

There is a convention for naming detectors in the geometry, and for this detector that is `HcalEndcapP`. It'd be nice if we could stick to that in all top level contexts (configurations etc).

wdconinc:
Until the lfhcal is included in arches and brycecanyon, it doesn't mean much that the checks are passing. There is no real overlap check done with other detector systems.

src/LFHCAL_geo.cpp
FriederikeBock: prefer to keep as is.

configurations/lfhcal_withInlay.yml
FriederikeBock: I don't see which "insert" you mean. "lfhcal_insertGeom.xml" is a file with the module definitions such that those which would overlap with the  insert are taken out of the geometry. Not the insert itself.  So there should not be a problem at all.

configurations/forwardNew.yml
FriederikeBock: no this is the whole forward setup, not just the hcal

configurations/lfhcal_withInlay.yml
wdconinc: Unclear name then.

compact/display_detailed.xml
```diff
@@ -87,6 +87,15 @@
...
     <vis name="HcalEndcapInsertVis"   ref="AnlGray"   showDaughters="true"  visible="true"/>
+    <vis name="LFHCAL8MModVis"        ref="AnlBlue" showDaughters="true" visible="true"/>
+    <vis name="LFHCAL4MModVis"        ref="AnlViolet" showDaughters="true" visible="true"/>
+    <vis name="LFHCAL8MModOuter"      ref="AnlBlue" showDaughters="false" visible="false"/>
+    <vis name="LFHCALLayerScintVis"   ref="AnlLight_Gray"   showDaughters="false" visible="false"/>
+    <vis name="LFHCALLayerSteelVis"   ref="AnlRed"   showDaughters="false" visible="false"/>
+    <vis name="LFHCALLayerTungstenVis" ref="AnlViolet"   showDaughters="true" visible="false"/>
+    <vis name="LFHCALLayerTyvekVis"   ref="AnlGold"   showDaughters="false" visible="false"/>
+    <vis name="LFHCALLayerTiOVis"     ref="AnlDarkGreen"   showDaughters="false" visible="false"/>
+    <vis name="GFHCALVis"             ref="AnlBlue"   showDaughters="true" visible="true"/>
```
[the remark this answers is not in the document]
FriederikeBock: done

src/InsertCalorimeter_geo.cpp
FriederikeBock: Otherwise the visualization of the insert is broken (regardless of which insert) we probably need it on

compact/hcal/lfhcal.xml
FriederikeBock: why would it make a difference? This is a pure formatting choice and would just blow up the file size even further.

compact/hcal/lfhcal_insertGeom.xml
FriederikeBock: this is not the insert Geometry but the lfhcal geometry in case there is an insert, which means I need the positions to be different as I need to take out modules which otherwise would collide with the actual insert.

compact/hcal/lfhcal.xml
FriederikeBock: Modified handing over of positions according to suggestion by Wouter.

src/LFHCAL_geo.cpp
FriederikeBock: I now initialized also all fields, but I would like to keep the default constructor, which I think is good practise.

configurations/lfhcal_withInlay.yml
FriederikeBock: Per suggestion of @veprbl I now renamed the file to: `lfhcal_with_space_for_insert.xml`, hope this helps making things clearer.

src/InsertCalorimeter_geo.cpp
veprbl:
```suggestion
  // Workaround for https://github.com/eic/epic/issues/411
  assembly.setVisAttributes(desc.visAttributes("InvisibleWithDaughters"));
```

src/LFHCAL_geo.cpp
wdconinc: Sure, that's fine. You probably know this, but https://en.cppreference.com/w/cpp/language/rule_of_three (and rules of zero and five included therein; naturally rule of zero is least work).

wdconinc:
My only remaining concern is that we don't include this new detector in any comprehensive geometries where it is used in a simulation or an overlap check. There are also not even internal overlap checks done on the lfhcal geometry. That should be done before we can merge, by creating a configuration like arches or brycecanyon that includes everything with the lfhcal. That's why I was suggesting that the lfhcal should be included instead of the regular positive endcap hadronic calorimeter.

FriederikeBock:
>

FriederikeBock:
@wdconinc I deliverately didn't include them in the arches and brycecanyon configs, as I wasn't sure whether they will be the once which we will be running in the future. I of course ran the overlap checks locally with them and there shouldn't be any conflicts. If you tell me that we won't be creating a 3rd global config starting with "C". I can of course include the LFHCal in brycecanyon and arches.

wdconinc:
Maybe I misunderstood the lfhcal's purpose then; I thought it was supposed to be the reference design for the forward endcap hcal and that the current implementation by Ryan was a conceptual approximation. Is it intended as an alternate technology choice that you are evaluating for a future upgrade?

FriederikeBock:
You are correct in your assumption, that this should be the default forward HCal, I am just deferring to our software coordinators (including you) to decide when you want to include in the central productions. But if you tell me go ahead I'll prepare the respective change.

FriederikeBock:
@wdconinc what are you checking in the benchmarks for the reco? I see those failing but I am not sure why.

src/LFHCAL_geo.cpp
```diff
@@ -0,0 +1,1084 @@
...
+  }
+
+  // create 8M modules
+  Volume  eightMassembly = createEightMModule ( desc, eightM_params, slice_Params, length, sens, renderComponents, allSensitive);
+  if (xpos8M.size() != ypos8M.size() || xpos8M.size() != zpos8M.size()){
+    std::cout << xpos8M.size() << "\t" <<ypos8M.size() <<  "\t" << zpos8M.size() << std::endl;
+    std::cout <<  "idiot you can't count" << std::endl;
+  } else {
+    for (int e = 0; e < (int)xpos8M.size(); e++){
+      if(e%20 == 0 ) std::cout <<  "LFHCAL placing 8M module: " << e << "/"<< (int)xpos8M.size() << "\t" << xpos8M[e] << "\t" << ypos8M[e] << "\t" << zpos8M[e]<< std::endl;
```
veprbl, line 1026:
```suggestion
```

src/LFHCAL_geo.cpp
```diff
@@ -0,0 +1,1084 @@
...
+  }
+
+  // create 4M modules
+  Volume  fourMassembly = createFourMModule ( desc, fourM_params, slice_Params,  length, sens, renderComponents, allSensitive);
+  if (xpos4M.size() != ypos4M.size() || xpos4M.size() != zpos4M.size()){
+    std::cout << xpos4M.size() << "\t" <<ypos4M.size() <<  "\t" << zpos4M.size() << std::endl;
+    std::cout <<  "idiot you can't count" << std::endl;
+  } else {
+    for (int f = 0; f < (int)xpos4M.size(); f++){
+      if(f%20 == 0 )std::cout <<  "LFHCAL placing 4M module: " << f << "/"<< (int)xpos4M.size() << "\t" << xpos4M[f] << "\t" << ypos4M[f] << "\t" << zpos4M[f]<< std::endl;
```
veprbl, line 1061:
```suggestion
```

veprbl (commented):
Let's avoid unnecessary output

src/LFHCAL_geo.cpp
```diff
@@ -0,0 +1,1084 @@
...
+                              SensitiveDetector sens,
+                              bool renderComp,
+                              bool allSen
+){
+  std::string baseName = "LFHCAL_8M";
+
+  // assembly definition
+  Box         modBox( mod_params.mod_width / 2., mod_params.mod_height / 2., length / 2.);
+  Volume  vol_mod(baseName,modBox,desc.material("Air"));
+  std::cout << "visualization string module: " << mod_params.mod_visStr.data() << std::endl;
```
veprbl, line 635: This is also not needed
```suggestion
```

veprbl:
I think the failure is irrelevant to this PR. Might be that we've dropped EventStore patch for DD4hep and Juggler doesn't support Frames.

wdconinc:
> I see those failing but I am not sure why.

They are failing right now because of the transitions in dd4hep/podio/eicrecon/juggler over the past weekend. No need to worry about them in the short term. I am waiting for a key4hep PR that they promised would come in the next two days. Once that's available, we can fix juggler and those checks will pass.

The fact htat hte physics_benchmarks are passing is what I'm mainly looking at.

compact/hcal/lfhcal_with_space_for_insert.xml
```diff
@@ -0,0 +1,1351 @@
...
+
+    </detector>
+
+  </detectors>
+
+  <!--  Definition of the readout segmentation/definition  -->
+  <readouts>
+    <readout name="LFHCALHits">
+      <segmentation type="NoSegmentation"/>
+      <id>system:8,moduleIDx:6,moduleIDy:6,moduletype:1,passive:1,towerx:2,towery:1,rlayerz:4,layerz:4</id>     </readout>
```
wdconinc, line 1348:
```suggestion
      <id>system:8,moduleIDx:6,moduleIDy:6,moduletype:1,passive:1,towerx:2,towery:1,rlayerz:4,layerz:4</id>
    </readout>
```

configurations/lfhcal_only.yml
```diff
@@ -0,0 +1,4 @@
+features:
+#  beampipe:
```
wdconinc, line 2:
```suggestion
```

src/InsertCalorimeter_geo.cpp
```diff
@@ -89,7 +89,8 @@ static Ref_t createDetector(Detector& desc, xml_h handle, SensitiveDetector sens
 
   // Assembly that will contain all the layers
   Assembly assembly(detName);
-
+  // Workaround for https://github.com/eic/epic/issues/411
```
wdconinc, line 92:
```suggestion
  // FIXME Workaround for https://github.com/eic/epic/issues/411
```

src/LFHCAL_geo.cpp
wdconinc: Replaced all `std::cout` instances with `printout(DEBUG)` calls. Users can run with `ddsim -v DEBUG` to see them.

wdconinc:
Two more things here (which you might have noticed I'm putting some cycles towards to):
- avoiding the calculation of module x and y id from the position (risk of incorrect values due to floating point vagaries when assigning floats to int), but specifying it in the position lines,
- avoiding the duplication of the module definitions in both lfhcal files (and a lot of positions that are in both).

FriederikeBock:
> Two more things here (which you might have noticed I'm putting some cycles towards to):
> 
> * avoiding the calculation of module x and y id from the position (risk of incorrect values due to floating point vagaries when assigning floats to int), but specifying it in the position lines,
> * avoiding the duplication of the module definitions in both lfhcal files (and a lot of positions that are in both).

According to what did you now order the module lines?

wdconinc:
> According to what did you now order the module lines?

Just alphabetically (because easiest and still deterministic). If you want, we can order by x^2+y^2 too, which would put the relevant differences up front.

FriederikeBock:
> > According to what did you now order the module lines?
> 
> Just alphabetically (because easiest and still deterministic). If you want, we can order by x^2+y^2 too, which would put the relevant differences up front.

I would prefer ordering by R (x^2+y^2), which at least the default LFHCal one should have been ordered by (just in decreasing order). Thank you.

wdconinc (approved):

PR #410, opened 2023-04-14
feat(dRICH): add a constant to provide the number of sensors

c-dilks:
Add constant `DRICH_num_sensors` to provide easy access to the number of SiPM sensors per sector. Warn if this number is different in any sectors.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators @DelloStritto

no

no

### DIFF
--- a/src/DRICH_geo.cpp
+++ b/src/DRICH_geo.cpp
@@ -552,7 +552,6 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
     else if (imod != desc.constantAsLong("DRICH_num_sensors"))
       printout(WARNING, "DRICH_geo", "number of sensors is not the same for each sector");
 
-
   } // END SECTOR LOOP //////////////////////////
 
   return det;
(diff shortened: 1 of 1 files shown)


src/DRICH_geo.cpp
```diff
@@ -546,6 +546,12 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
 
     // END SENSOR MODULE LOOP ------------------------
 
+    // add constant for access to the number of modules per sector
+    if (isec == 0)
+      desc.add(Constant("DRICH_num_sensors", std::to_string(imod)));
+    else if (imod != desc.constantAsLong("DRICH_num_sensors"))
```
wdconinc, line 552: Is there no deterministic variable that can be used? Does this need to be the final value of a loop counter?

c-dilks: Not really, this comes from the spherical tiling algorithm above, with constraints set by https://github.com/eic/epic/blob/6446592846efd24d7d4c902b0f7b1e5e30970314/compact/pid/drich.xml#L197-L222
We have a few other derived 'constants', such as the mirror center and radius; the compact file specifies the desired focal region (offsets) and the plugin uses spherical optics equations to determine the mirror parameters from this specification.

wdconinc (approved):
Thanks for the response. I figured it was something like that.

PR #414, opened 2023-04-20
feat: epic full with pfrich, imaging

wdconinc:
The default epic.xml geometry, constructed from the full.yml configuration, is updated to use the pfRICH and imaging barrel calorimeter.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

Yes, different detector is default.

### DIFF
--- a/compact/tracking/definitions.xml
+++ b/compact/tracking/definitions.xml
@@ -131,12 +131,6 @@
       vis="TrackerSubAssemblyVis">
       <composite name="BarrelTOF"/>
     </detector>
-    <detector id="TrackerSubAssembly_5_ID"
-        name="OuterMPGDSubAssembly"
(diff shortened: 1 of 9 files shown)


sly2j:
This also needs to pull in the correct tracking configuration (currently in `compact/tracking/definitions_brycecanyon.xml`. Ideally, I would do
```bash
git mv compact/tracking/definitions.xml compact/tracking/definitions_arches.xml
git mv compact/tracking/definitions_brycecanyon.xml compact/tracking/definitions.xml
```

I *think* that should take care of everything.

wdconinc:
> I *think* that should take care of everything.

Good catch! Also needed some changes to configurations.

sly2j (approved):

PR #437, opened 2023-06-01
fix: set default CMAKE_INSTALL_PREFIX

wdconinc:
If CMAKE_INSTALL_PREFIX is not defined, then it takes on an incorrect default value that is written into the `setup.sh` script (which leads to a `setup.sh` installed by default in the source dir with a DETECTOR_PATH set to a /usr/local dir).

This commit explicitly defines the CMAKE_INSTALL_PREFIX to the source directory if it is not explicitly set.

- [x] Bug fix (issue: incorrect DETECTOR_PATH in setup.sh)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,11 @@ endif()
 # Export compile commands as json for run-clang-tidy
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
+# Install to the top directory by default
+if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
+    set(CMAKE_INSTALL_PREFIX ${CMAKE_SOURCE_DIR} CACHE PATH "Install in top directory by default" FORCE)
+endif()
+
 # Install into GNU standard directories
 include(GNUInstallDirs)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,7 +33,7 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 # Install to the top directory by default
 if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
-    set(CMAKE_INSTALL_PREFIX ${CMAKE_SOURCE_DIR} CACHE PATH "Install in top directory by default" FORCE)
+    set(CMAKE_INSTALL_PREFIX ${CMAKE_SOURCE_DIR}/prefix CACHE PATH "Install in $src/prefix/ directory by default" FORCE)
 endif()
 
 # Install into GNU standard directories


wdconinc:
> Sounds like a separate issue. 

It is in a sense caused by how an unset `CMAKE_INSTALL_PREFIX` is used by the `GNUInstallDirs` module versus the regular `install` command.
- If CMAKE_INSTALL_PREFIX is not set, then `GNUInstallDirs` defaults to `/usr/local`, which is not documented as far as I could find. Then inside the `setup.sh.in` template the `${CMAKE_INSTALL_FULL_DATADIR}` ends up in `/usr/local/share`.
- However, the `install( .. DESTINATION ${CMAKE_INSTALL_PREFIX})` just becomes DESTINATION without any value, which just becomes `./`.

So, that's why they end up in different places.

This PR in essence avoids the undocumented `GNUInstallDirs` behavior by making sure that `CMAKE_INSTALL_PREFIX` is always defined.

CMakeLists.txt
```diff
@@ -31,6 +31,11 @@ endif()
 # Export compile commands as json for run-clang-tidy
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
+# Install to the top directory by default
+if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
+    set(CMAKE_INSTALL_PREFIX ${CMAKE_SOURCE_DIR} CACHE PATH "Install in top directory by default" FORCE)
```
veprbl, line 36: Installing to the source tree is a bit dodgy, to my taste. Is there any opposition to going for additional `./prefix` ?
```suggestion
    set(CMAKE_INSTALL_PREFIX ${CMAKE_SOURCE_DIR}/prefix CACHE PATH "Install in $src/prefix/ directory by default" FORCE)
```

wdconinc: No opposition. I was just using what EICrecon does, but agree it is ugly.

veprbl:
> * If CMAKE_INSTALL_PREFIX is not set, then `GNUInstallDirs` defaults to `/usr/local`, which is not documented as far as I could find. Then inside the `setup.sh.in` template the `${CMAKE_INSTALL_FULL_DATADIR}` ends up in `/usr/local/share`.

I always believed that /usr/local was the CMake default, not specific to GNUInstallDirs.
https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX.html#variable:CMAKE_INSTALL_PREFIX

> * However, the `install( .. DESTINATION ${CMAKE_INSTALL_PREFIX})` just becomes DESTINATION without any value, which just becomes `./`.

According to docs https://cmake.org/cmake/help/latest/command/install.html, DESTINATION is already relative to CMAKE_INSTALL_PREFIX, so it should have been (not tested):
```
install( .. DESTINATION ./)
```

veprbl (approved):

PR #442, opened 2023-06-03
fix: disable optical photon generation in hpDIRC

wdconinc:
The hpDIRC is very inefficient right, causing 20x slowdowns for DIS datasets. Since there is no reconstruction yet for the hpDIRC, there is no reason to simulate all these photons.

This PR reverts the optical materials to their regular counterparts.

@niwgit This follows in part from today's discussion. There is no digitization or reconstruction of the hpDIRC, so in the short term there is no real use for these optical photons. Before re-enabling the optical photons by default we would probably need digitization and reconstruction, and we need a better understanding of the impact on running times.

Cc: @mdiefent @rahmans1 @T-Britton 

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

  - We need a benchmark to catch regressions in running time like this... @veprbl @torrijeske 

Yes, reverts to not having optical photons in the hpDIRC.

Yes, no more optical photons in the hpDIRC.

### DIFF
--- a/compact/materials.xml
+++ b/compact/materials.xml
@@ -35,6 +35,17 @@
     <composite n="1" ref="Si"/>
     <composite n="2" ref="O"/>
   </material>
+  <material name="Epotek">
+    <D type="density" value="1.2" unit="g/cm3"/>
+    <composite n="5" ref="H"/>
+    <composite n="3" ref="C"/>
+    <composite n="2" ref="O"/>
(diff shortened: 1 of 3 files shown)


T-Britton (approved):
These changes appear to get rid of optical photons as desired.

PR #449, opened 2023-06-11
fix: dd4hep::CartesianField renamed type to field_type

wdconinc:
In the next DD4hep version, the field type will be stored in a variable `field_type` not `type`. This will cause epic to fail to compile. Of course, until then this version will fail to compile. And all older geometries will fail to compile with the newer DD4hep as well. So that's kinda annoying and something I don't have a good solution to...

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,10 +43,6 @@ include(GNUInstallDirs)
 find_package(DD4hep 1.21 REQUIRED COMPONENTS DDCore DDRec)
 find_package(fmt REQUIRED)
 
-# ePIC compile definitions
-option(EPIC_COMPILE_DEFINITIONS "Additional compile definitions" "")
-add_compile_definitions(${EPIC_COMPILE_DEFINITIONS})
-
 #-----------------------------------------------------------------------------------
 set(a_lib_name ${PROJECT_NAME})
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,6 +43,10 @@ include(GNUInstallDirs)
 find_package(DD4hep 1.21 REQUIRED COMPONENTS DDCore DDRec)
 find_package(fmt REQUIRED)
 
+# ePIC compile definitions
+option(EPIC_COMPILE_DEFINITIONS "Additional compile definitions" "")
+add_compile_definitions(${EPIC_COMPILE_DEFINITIONS})
+
 #-----------------------------------------------------------------------------------
 set(a_lib_name ${PROJECT_NAME})
 
--- a/src/FieldMapBrBz.cpp
+++ b/src/FieldMapBrBz.cpp
@@ -23,7 +23,7 @@ using namespace dd4hep;
 // implementation of the field map
 class FieldMapBrBz : public dd4hep::CartesianField::Object {
 public:
-  FieldMapBrBz(const std::string& field_type = "magnetic");
+  FieldMapBrBz(const std::string& field_type_str = "magnetic");
   void Configure(double rmin, double rmax, double rstep, double zmin, double zmax, double zstep);
   void LoadMap(const std::string& map_file, double scale);
   void GetIndices(double r, double z, int& ir, int& iz, double& dr, double& dz);
@@ -42,9 +42,9 @@ private:
 };
 
 // constructor
-FieldMapBrBz::FieldMapBrBz(const std::string& field_type)
+FieldMapBrBz::FieldMapBrBz(const std::string& field_type_str)
 {
-  std::string ftype = field_type;
+  std::string ftype = field_type_str;
   for (auto& c : ftype) {
     c = tolower(c);
   }
@@ -56,7 +56,7 @@ FieldMapBrBz::FieldMapBrBz(const std::string& field_type)
     type = CartesianField::ELECTRIC;
   } else {
     type = CartesianField::UNKNOWN;
-    std::cout << "FieldMapBrBz Warning: Unknown field type " << field_type << "!" << std::endl;
+    std::cout << "FieldMapBrBz Warning: Unknown field type " << field_type_str << "!" << std::endl;
   }
 }
 
--- a/src/FieldMapBrBz.cpp
+++ b/src/FieldMapBrBz.cpp
@@ -20,6 +20,13 @@ namespace fs = std::filesystem;
 
 using namespace dd4hep;
 
+// In DD4hep 1.26, the name for the `field_type` enum changed (from `type`).
+#if DD4HEP_VERSION_GE(1, 26)
+#define DD4HEP_FIELD_TYPE field_type
+#else
+#define DD4HEP_FIELD_TYPE type
+#endif
+
 // implementation of the field map
 class FieldMapBrBz : public dd4hep::CartesianField::Object {
 public:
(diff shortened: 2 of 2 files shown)


wdconinc:
Interested in ideas of how we can keep this working for newer versions of DD4hep with older version of the geometry (6 months at least).

veprbl:
Can't reproduce a compilation failure with https://github.com/AIDASoft/DD4hep/commit/3347996361696054fc32b1515bc277be731c724f and epic 23.06.1

wdconinc:
Yeah, not sure why this was failing before.

wdconinc:
Here's what the failure looks like: https://eicweb.phy.anl.gov/containers/eic_container/-/jobs/1697713

That's a dd4hep PR 1080 merged into 1.25.1, and then fails to compile epic.

veprbl:
`field_type` should not conflict as it's in the scope `TypedObject`
https://github.com/MarkusFrankATcernch/DD4hep/blob/51d7ec194fb5c3e445139d2843a9d9bdab6363c5/DDCore/include/DD4hep/Fields.h#L56
Need to test with `-Wshadow=error`.

wdconinc:
I wonder if something was added in a later PR that makes it work again?

veprbl:
I thought so too, but the line number 56 is matching the final result. Would be a big coincidence.

wdconinc:
So, this is definitely something we need to fix. But it depends on a few things. Summary of facts.
- DD4hep 1.26 will change the name of `dd4hep::CartesianField::type` to `dd4hep::CartesianField::field_type`, see https://github.com/AIDASoft/DD4hep/pull/1080/files#diff-03c4f002cae351121043bc742e94e1aa888c78fc38af4da96a844d5a9dfb71c7R56 (this is already in master as part of https://github.com/AIDASoft/DD4hep/pull/1080)
  - Opinion: This is an annoying public interface change, but it's for good reason since `dd4hep::CartesianField` inherits from `TypedObject` which has a `type` field as well, so it avoids overriding a member variable that is needed for handles to get type strings.
- Depending on compiler defaults, a shadowing warning (or error) may be emitted because we use `field_type` as a local variable in the `FieldMapBrBz` constructor, https://github.com/eic/epic/blob/main/src/FieldMapBrBz.cpp#L26.
- Even if no shadowing warning is emitted, the code in `FieldMapBrBz` now sets the string `type` to the enum dd4hep::CartesianField::FieldType`. With DD4hep 1.26 (or master) this compiles but doesn't work. It throws an error when running because the field type is unset.
  - Opinion: The assignment likely works due to various implicity upconversions from enum to int to string.

wdconinc:
With the latest pushes, this now should work for released DD4hep 1.25 versions and released 1.26 versions (when released). It uses preprocessor directives wit the `DD4HEP_VERSION_GE(1,26)` macro.

It also allows for overriding the automatic test because the current master (or a backport of PR 1080 into 1.25.1 in order to fix the far forward magnetic field rotations) requires that we override the automatic field type variable name determination.

### Testing
This was tested by adding the following field printing line in `FieldMapBrBz::fieldComponents`
```c++
  std::cout << r << "," << z << "," << phi << ": " << field[0] / tesla << "," << field[1] / tesla << ","
            << field[2] / tesla << std::endl;
```
and executing the following:
- in `eic-shell` (DD4hep 1.25.1):
```console
cmake -Bbuild-1.25.1 -S. -DCMAKE_INSTALL_PREFIX=install-1.25.1
make -Cbuild-1.25.1 -j8 install
source install-1.25.1/setup.sh
ddsim --compactFile ${DETECTOR_PATH}/epic_tracking_only.xml -G --gun.distribution "uniform" --random.seed 1 -N 1
```

- in docker with `eicweb/jug_dev:unstable-mr-290-585` (DD4hep 1.25.1 with PR 1080 backport):
```console
docker run --rm -v ~/git/epic:/mnt -it eicweb/jug_dev:unstable-mr-290-585
cmake -Bbuild-1.26 -S. -DCMAKE_INSTALL_PREFIX=install-1.26 -DEPIC_COMPILE_DEFINITIONS="DD4HEP_FIELD_TYPE_OVERRIDE=field_type"
make -Cbuild-1.26 -j8 install 
source install-1.26/setup.sh 
ddsim --compactFile ${DETECTOR_PATH}/epic_tracking_only.xml -G --gun.distribution "uniform" --random.seed 1 -N 1
```

In both cases the printed field values were identical and of the right magnitude (-1.7 T).

### What now?
- For the next campaign we will backport PR 1080 into DD4hep 1.25.1 (by rebasing the eic-spack dd4hep-pr1080 branch against the current develop and releases/v0.20), and use the override mechanism as above to install the correct version.
- This also means that all previous tagged geometries will be unable to be compiled 1.26, or with the backported DD4hep 1.25.1 in the production environment. We will patch these geometry versions 'by hand' upon installation in the eic-shell container.

wdconinc:
I'm also hoping we can get rid of the ugly override hack as soon as we don't need it anymore for backwards compatibility

src/FieldMapBrBz.cpp
```diff
@@ -20,10 +20,29 @@ namespace fs = std::filesystem;
...
+#define DD4HEP_FIELD_TYPE field_type
+#else
+#define DD4HEP_FIELD_TYPE type
+#endif
+
+// This allows us to specify the name of the variable by hand, when patching
+// the previous versions, by setting `DD4HEP_FIELD_TYPE_OVERRIDE`.
+// TODO remove after DD4hep 1.26 release
+#define XSTR(x) STR(x)
+#define STR(x) #x
```
sly2j, line 34: Does this section, in principle, need some #ifdef XSTR #undef XSTR ... etc., to minimize side effects?

wdconinc: And push pop the previous value? https://gcc.gnu.org/onlinedocs/gcc/Push_002fPop-Macro-Pragmas.html

veprbl: Not needed in cpp file

CMakeLists.txt
```diff
@@ -43,6 +43,10 @@ include(GNUInstallDirs)
 find_package(DD4hep 1.21 REQUIRED COMPONENTS DDCore DDRec)
 find_package(fmt REQUIRED)
 
+# ePIC compile definitions
+option(EPIC_COMPILE_DEFINITIONS "Additional compile definitions" "")
+add_compile_definitions(${EPIC_COMPILE_DEFINITIONS})
```
veprbl, line 48: cmake already respects CXXFLAGS from the environment variable. Any reason to not use that?

wdconinc: No real functional reason other than a deeply engrained reflex towards compiler abstraction :-)

`CXXFLAGS` is passed to the compiler unchanged, while `add_compile_definitions()` does 'the right thing' for whichever compiler is active (not as relevant here since I suspect that LLVM uses the same flag for this).

wdconinc: I'll make sure this works with `CXXFLAGS` and revert to using that.

wdconinc:
Downstream ugly hack at https://eicweb.phy.anl.gov/containers/eic_container/-/merge_requests/585 (@sly2j you'll love this)

CMakeLists.txt
wdconinc: Reverted that commit since we can use both env var `CXXFLAGS` and cmake var `CMAKE_CXX_FLAGS`.

veprbl: Reminder: `-DCMAKE_CXX_FLAGS` is a bit clunky as it does not influence `CMAKE_CXX_FLAGS_RELEASE` and such.

veprbl (approved):
I wasn't able to reproduce the warning, but I can confirm that this still compiles with DD4hep master.

PR #451, opened 2023-06-21
SciGlassCalorimeter_geo: support automatic calculation of flaring angles

veprbl:

This extends geometry generation capabilities to easily produce alternative geometries with some of the flaring angles calculated automatically. The reference ePIC geometry should not be affected by this change, as it still uses cells in the shape of truncated pyramids (now specified explicitly) and hardcoded flaring-at-face angles.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No


veprbl:
Example custom geometry with both features enabled:
image

src/BarrelCalorimeterSciGlass_geo.cpp
```diff
@@ -209,25 +209,55 @@ static Ref_t create_detector(Detector& lcdd, xml_h handle, SensitiveDetector sen
...
 
       const double z  = length / 2;
       // Face parameters (see doc/sciglass_tower_front_view.svg for definitions)
       const double y1 = family_dim_handle.y1();
       const double y2 = y1 + length * tan(flare_angle_polar);
       double       x1 = family_dim_handle.x1();
+
+      const auto   flare_angle_at_face_s = family_dim_handle.attr<string>(_Unicode(flare_angle_at_face));
+      double flare_angle_at_face;
+      if (flare_angle_at_face_s == "auto") {
```
wdconinc, line 225: Is `auto` the only option, or can `auto` be more descriptive? E.g. how you use `pyramidal` instead of `auto` for that value.

wdconinc (approved):

src/BarrelCalorimeterSciGlass_geo.cpp
veprbl: I was asking myself this question. In principle, one could call it `fit`, although that's still vague about all known and unknown subtle details. I can rename to that, or we could always deprecate the `auto` value if some alternative naming naturally emerges.

wdconinc: Sounds good to me. I approved in any case, so feel free to merge.

veprbl: This is good enough for now. Let's merge

PR #503, opened 2023-08-21
configurations/ip6_*.yml: add marco field

veprbl:
This would enable the barrel magnet field for FF studies by default.


ajentsch:
> This would enable the barrel magnet field for FF studies by default.

No no no, please DO NOT make it default to have the MARCO field in the FF + FB ONLY simulations. Having a field present without the geometry will potentially induce confusion, since all of the stand-alone studies DO NOT use the MARCO field (or any other solenoid).

PR #506, opened 2023-08-26
Add a G4 material scan script

Chao1009:
Add a material scan script that utilizes the GEANT4 material scan tool.
It scans over given eta and phi ranges (take the average material thickness over phi).

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,3 +115,8 @@ configure_file(templates/setup.sh.in ${CMAKE_CURRENT_BINARY_DIR}/setup.sh @ONLY)
 install(FILES ${CMAKE_CURRENT_BINARY_DIR}/setup.sh
     DESTINATION ${CMAKE_INSTALL_PREFIX}
 )
+
+# install programs
+install(PROGRAMS bin/g4mat_scan_to_csv
+  DESTINATION ${CMAKE_INSTALL_BINDIR}
+)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,6 +117,6 @@ install(FILES ${CMAKE_CURRENT_BINARY_DIR}/setup.sh
 )
 
 # install programs
-install(PROGRAMS bin/g4mat_scan_to_csv
+install(PROGRAMS bin/g4MaterialScan_to_csv
   DESTINATION ${CMAKE_INSTALL_BINDIR}
 )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,8 +115,3 @@ configure_file(templates/setup.sh.in ${CMAKE_CURRENT_BINARY_DIR}/setup.sh @ONLY)
 install(FILES ${CMAKE_CURRENT_BINARY_DIR}/setup.sh
     DESTINATION ${CMAKE_INSTALL_PREFIX}
 )
-
-# copying test scripts
-install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/scripts/subdetector_tests
-  DESTINATION ${CMAKE_INSTALL_PREFIX}/scripts
- )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,3 +115,8 @@ configure_file(templates/setup.sh.in ${CMAKE_CURRENT_BINARY_DIR}/setup.sh @ONLY)
 install(FILES ${CMAKE_CURRENT_BINARY_DIR}/setup.sh
     DESTINATION ${CMAKE_INSTALL_PREFIX}
 )
+
+# copying test scripts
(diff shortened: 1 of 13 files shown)


Chao1009:
running a test with 81 x 5 = 405 scans.

wdconinc (changes requested):
This doesn't install anywhere. In eic-shell we don't copy the repository but install to a prefix, so this won't be accessible. If we want this user-facing, then it should install. Can you add this to a CMakeLists.txt target?

CMakeLists.txt
```diff
@@ -115,3 +115,8 @@ configure_file(templates/setup.sh.in ${CMAKE_CURRENT_BINARY_DIR}/setup.sh @ONLY)
 install(FILES ${CMAKE_CURRENT_BINARY_DIR}/setup.sh
     DESTINATION ${CMAKE_INSTALL_PREFIX}
 )
+
+# copying test scripts
+install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/scripts/subdetector_tests
+  DESTINATION ${CMAKE_INSTALL_PREFIX}/scripts
```
veprbl, line 121:
```suggestion
install(PROGRAMS scripts/subdetector_tests/g4_material_scan.py
  DESTINATION ${CMAKE_INSTALL_BINDIR}
```

scripts/subdetector_tests/g4_material_scan.py
```diff
@@ -0,0 +1,148 @@
+# SPDX-License-Identifier: LGPL-3.0-or-later
```
veprbl, line 1:
```suggestion
#!/usr/bin/env python3

# SPDX-License-Identifier: LGPL-3.0-or-later
```

CMakeLists.txt
Chao1009: I think we could put this script in `bin` instead of `scripts/subdetector_tests` instead of installing it as a program

wdconinc:
I would like to avoid confusion between `materialScan`, `g4MaterialScan`, and this script if it is installed in ther user's PATH. It should be clear(er) from the name what it does in addition to these two existing utilities.

Chao1009:
> I would like to avoid confusion between `materialScan`, `g4MaterialScan`, and this script if it is installed in ther user's PATH. It should be clear(er) from the name what it does in addition to these two existing utilities.

Yes, I renamed it to `g4mat_scan_to_csv`.

bin/g4mat_scan_to_csv
```diff
@@ -0,0 +1,150 @@
...
+        # print(eta_scan.sum(axis=1))
+        # replace nan value with 0 and then calculate average values
+        for mat, xval in eta_scan.fillna(0.).mean(axis=1).items():
+            if mat not in mats_indices:
+                mats_indices[mat] = len(mats_indices)
+            j = mats_indices.get(mat)
+            data[i, j] = xval
+
+    result = pd.DataFrame(columns=mats_indices.keys(), index=etas, data=data[:, :len(mats_indices)])
+    result.to_csv(args.output)
```
veprbl, line 152:
```suggestion
    result.to_csv(args.output, sep="\t", quoting=csv.QUOTE_NONE)
```
Also, we could switch from CSV to TSV, which is a more common format in HEP

veprbl:
```suggestion
    result.to_csv(args.output, sep="\t")
```
`quoting` not supported, it looks like

Chao1009:
The first five materials from the scan
`g4mat_scan_to_csv epic_central.xml --phi 0 --eta 1:2:1 `

          Air  AirOptical  Aluminum  Aluminum5083        Ar
1.0  0.145034    0.000106  0.207398      1.417401  0.000065
2.0  0.329851    0.000000  0.055385      0.000000  0.000000

Chao1009:
Removed averaging in the output, now it has a multi-index: (eta, phi).
One can average the result over phi during the analysis.

Chao1009:
@wdconinc requested changes are addressed.

Chao1009:
> Looks good to me. The only change I'd like you to consider is whether we can rename the script to `g4MaterialScan_to_csv` since it builds on `g4MaterialScan`. That will allow users to find it when they are currently using `g4MaterialScan` by just typing tab. It also makes it clearer that it really is just using `g4MaterialScan` under the hood, so the users will have an accurate understanding of what it does and how it works.
> 
> Similarly, it would be good to use arguments that are similar to `g4MaterialScan`, so a named `--compact` or `-c` flag for the xml file instead of a positional argument. I think this consistency will reduce the possibility of errors.

These are both good suggestions, I've implemented them in the latest commit.

wdconinc (approved):

PR #507, opened 2023-08-27
labeler.yml: assign files for "topic: magnets"

veprbl:


.github/labeler.yml
```diff
@@ -49,6 +49,11 @@
   - "templates/setup.sh.in"
   - "views/**/*"
 
+"topic: magnets":
+  - "compact/fields/**/*"
+  - "compact/fields.xml"
+  - "compact/solenoid.xml"
```
wdconinc, line 55: This may be fragile since far backward and far forward might change their xml interfaces, but should we add this?
```suggestion
  - "compact/solenoid.xml"
  - "compact/far_backward/magnets.xml"
  - "compact/far_forward/magnets.xml"
```

veprbl: We do

wdconinc (approved):

PR #508, opened 2023-08-31
Add hadron magnetic fields for nuclei

ajentsch:

Added a few extra magnetic field configurations to make user's life easier. There is a caveat for light-ions, however. We do not have real lattice files, so properly steering the beams for the far-forward detectors will mean that the B0 field is scaled down, which will worsen the resolution for B0 tracking + reco. You will still see hits, etc., but you will see pT and p resolution much worse than for a standard field file. Until we have new lattice files, users will need to run simulations twice if using both B0 + RP/OMD information for various events.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No, just adds a few more configuration files.

No.


wdconinc:
Nothing major to comment on. If we use he3, he4, can we use h2, pb, and au as well? And while we're at it can we use capitalization like Au, Pb, He, H?

ajentsch:
Sure. I'm taking a day off tomorrow so I'll fix it after the holiday
Monday, so Tuesday.

wdconinc:
In that case I can do it. Should make it easier as we want to tag on Tuesday.

ajentsch:
Okay, thank you.

wdconinc (approved):

PR #524, opened 2023-09-13
Update ff magnet vacuum

ajentsch:

This PR updates the magnet FF vacuum to remove gaps and overlaps and properly use CutTubes to connect the inner pieces together. Additionally, it has been modified to take the magnet locations as its primary input, and calculate everything relative to that, which should make maintenance easier in the long-run.

It should be noted that this volume is ONLY for the region between the end of the central beam pipe and the end of the b1apf magnet. The region occupied by the RP and OMD is handled separately.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No, it should further improve basic functionality.

Yes, it make the vacuum volume in the FF region more-continuous.

### DIFF
--- a/compact/far_forward/vacuum.xml
+++ b/compact/far_forward/vacuum.xml
@@ -10,12 +10,58 @@
 
   <detectors>
 
-    <detector id="VacuumMagnetElement_1_ID" name="VacuumMagnetElement" type="magnetElementInnerVacuum" vis="AnlGold">
-      <position x="0.0" y="0.0" z="0.0" />
-      <rotation x="0*rad" y="0*rad" z="0*rad" />
-    </detector>
-
+      <detector id="VacuumMagnetElement_1_ID" name="VacuumMagnetElement" type="magnetElementInnerVacuum" vis="AnlGold">
+        <element item="0">
+		<placement x="B0PF_XPosition" y="0*m" z="B0PF_CenterPosition" theta="B0PF_RotationAngle" />
+        	<dimensions x="B0PF_InnerRadius*4" y="B0PF_InnerRadius*4" z="B0PF_Length" r="B0PF_InnerRadius*2.0" />
+        	<apperture x="B0PF_InnerRadius" y="B0PF_InnerRadius" r="B0PF_InnerRadius" />
+	</element>
+      
+        <element item="1">
+		<placement x="B0APF_XPosition" y="0*m" z="B0APF_CenterPosition" theta="B0APF_RotationAngle" />
+        	<dimensions x="B0APF_InnerRadius*4" y="B0APF_InnerRadius*4" z="B0APF_Length" r="B0APF_InnerRadius*2.0" />
+        	<apperture x="B0APF_InnerRadius" y="B0APF_InnerRadius" r="B0APF_InnerRadius" />
+	</element>
+      
+        <element item="2">
+	        <placement x="Q1APF_XPosition" y="0*m" z="Q1APF_CenterPosition" theta="Q1APF_RotationAngle" />
+	        <dimensions x="Q1APF_InnerRadius*4" y="Q1APF_InnerRadius*4" z="Q1APF_Length" r="2.0*Q1APF_InnerRadius"/>
+	        <apperture x="Q1APF_InnerRadius*2" y="Q1APF_InnerRadius*2"  r="Q1APF_InnerRadius"/>
+        </element>
+	
+        <element item="3">
+	        <placement x="Q1BPF_XPosition" y="0*m" z="Q1BPF_CenterPosition" theta="Q1BPF_RotationAngle" />
+	        <dimensions x="Q1BPF_InnerRadius*4" y="Q1BPF_InnerRadius*4" z="Q1BPF_Length"  r="2.0*Q1BPF_InnerRadius"/>
+	        <apperture x="Q1BPF_InnerRadius*2" y="Q1BPF_InnerRadius*2"  r="Q1BPF_InnerRadius"/>
+        </element>
+      
+        <element item="4">
+	        <placement x="Q2PF_XPosition" y="0*m" z="Q2PF_CenterPosition" theta="Q2PF_RotationAngle" />
+	        <dimensions x="Q2PF_InnerRadius*4" y="Q2PF_InnerRadius*4" z="Q2PF_Length" r="2.0*Q2PF_InnerRadius" />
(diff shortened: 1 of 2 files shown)


compact/far_forward/vacuum.xml
```diff
@@ -1,5 +1,5 @@
 <!-- SPDX-License-Identifier: LGPL-3.0-or-later -->
-<!-- Copyright (C) 2022 Whitney Armstrong, Alex Jentsch -->
+<!-- Copyright (C) 2023 Alex Jentsch -->
```
wdconinc, line 2: Keep previous years and names (unless you think the previous names were a mistake or this is a full rewrite).

src/magnetVacuumFF.cpp
```diff
@@ -1,14 +1,7 @@
...
-
-//==========================================================================
-//
-//      <detector name ="DetName" type="Beampipe" >
-//      <layer id="#(int)" inner_r="#(double)" outer_z="#(double)" >
-//      <slice material="string" thickness="#(double)" >
-//      </layer>
-//      </detector>
-//==========================================================================
+// Copyright (C) 2023 Alex Jentsch
```
wdconinc, line 2: Same comment as above.

compact/far_forward/vacuum.xml
```diff
@@ -10,12 +10,50 @@
...
-  </detectors>
+        <element item="1">
+                <placement x="B0APF_XPosition" y="0*m" z="B0APF_CenterPosition" theta="B0APF_RotationAngle" />
+                <dimensions x="B0APF_InnerRadius*4" y="B0APF_InnerRadius*4" z="B0APF_Length" r="B0APF_InnerRadius*2.0" />
+                <apperture x="B0APF_InnerRadius" y="B0APF_InnerRadius" r="B0APF_InnerRadius" />
+        </element>
+
+        <element item="2">
+                <placement x="Q1APF_XPosition" y="0*m" z="Q1APF_CenterPosition" theta="Q1APF_RotationAngle" />
+                <dimensions x="Q1APF_InnerRadius*4" y="Q1APF_InnerRadius*4" z="Q1APF_Length" r="2.0*Q1APF_InnerRadius"/>
```
wdconinc, line 28: Is this correct: `r="2.0*..._Radius`?

wdconinc: Ok, I think I see what's happening here.

src/magnetVacuumFF.cpp
```diff
@@ -31,155 +24,256 @@ using namespace dd4hep;
...
+        PlacedVolume pv_assembly;
+
+        //----------------------------------------------
+        // Starting point is only the magnet centers,
+        // lengths, rotations, and radii -->
+        // everything else calculated internally to
+        // make it easier to update later.
+        //----------------------------------------------
+
+        const int numGaps = 6; //number of gaps between magnets (excluding the IP to B0pf transition -- special case)
```
wdconinc, line 51: Instead of using a prefilled dimension, I'd suggest a dynamically expanding vector of struct with radius etc. Then you can fill this in the loop, and avoid the index idx.

wdconinc (commented):
Probably more comments later but some of the code not formatting very readably on mobile.

compact/far_forward/vacuum.xml
ajentsch: It was essentially a full re-write in both cases - and for the .cpp file, I had originally written it from scratch, but copy and pasted the skeleton from a template during the ATHENA times. I just want to make sure the relevant person is contacted in the event of a question.

src/magnetVacuumFF.cpp
ajentsch: Same as above - this has basically been completely re-written from scratch.

compact/far_forward/vacuum.xml
ajentsch: I removed the unused numbers entirely.

src/magnetVacuumFF.cpp
ajentsch: This is fixed - everything is std::vectors now.

src/magnetVacuumFF.cpp
```diff
@@ -31,155 +24,256 @@ using namespace dd4hep;
...
+        pv_assembly = det.pickMotherVolume(sdet).placeVolume(assembly);
+        pv_assembly.addPhysVolID("system", x_det.id()).addPhysVolID("barrel", 1);
+        sdet.setPlacement(pv_assembly);
+        assembly->GetShape()->ComputeBBox();
+        return sdet;
+}
+
+double getRotatedZ(double z, double x, double angle){
+
+        return z*TMath::Cos(angle) - x*TMath::Sin(angle);
```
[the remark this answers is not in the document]
ajentsch: Done - I also changed TMath::Sqrt() and TMath::Power to sqrt and pow, and change TMath::TwoPi() to 2*M_PI.

src/magnetVacuumFF.cpp
```diff
@@ -31,155 +23,264 @@ using namespace dd4hep;
  * \endcode
  *
  */
+
+double getRotatedZ(double z, double x, double angle);
+double getRotatedX(double z, double x, double angle);
```
wdconinc, line 28: These should be static to have static linkage, i.e. this function with this name is only valid within this file. Otherwise someone else could use getRotatedZ and pick up this one, or worse (for your case) you could end up using someone else's getRotatedZ whenever they implement one.

```suggestion
static double getRotatedZ(double z, double x, double angle);
static double getRotatedX(double z, double x, double angle);
```
and I would put the actual function implementation that you have below here as well.

src/magnetVacuumFF.cpp
```diff
@@ -31,155 +23,264 @@ using namespace dd4hep;
...
-  DetElement tube_de_7(sdet, "sector7_de", 1);
-  tube_de_7.setPlacement(pv_b1apf_vacuum);
-
-
-  pv_assembly = det.pickMotherVolume(sdet).placeVolume(assembly); //, posAndRot);
-  pv_assembly.addPhysVolID("system", x_det.id()).addPhysVolID("barrel", 1);
-  sdet.setPlacement(pv_assembly);
-  assembly->GetShape()->ComputeBBox();
-  return sdet;
+        //using namespace ROOT::Math;
```
wdconinc, line 33:
```suggestion
```

wdconinc (approved):

wdconinc (approved):

PR #527, opened 2023-09-18
Implement lumi dipole B mapping

dhevang:

Implement lumi dipole B mapping and modify source code to treat R-Z and X-Y-Z coordinate systems.

- [ ] Bug fix (issue #__)
- [ ] Documentation update
- [ ] Other: __


dhevang:
@wdconinc , in your code, FieldMapBrBz.cpp, which I used as the basis for this extension, you had a coordinate transformation here: https://github.com/eic/epic/blob/19ec3f38d53ff21b4ebb0646670a015a878927ca/src/FieldMapBrBz.cpp#L152
and then a B field transformation (trans) here:
https://github.com/eic/epic/blob/19ec3f38d53ff21b4ebb0646670a015a878927ca/src/FieldMapBrBz.cpp#L182
I'm not sure I understand the latter.  To me, only the first one was needed.  Anyway, the translation and rotation for the MARCO field were zero, so this had no effect.  However, for the lumi dipole field, I have the translation active.

wdconinc:
Can you [rebase](https://docs.github.com/en/get-started/using-git/about-git-rebase) so this lists only the changes relevant to this PR under [files](https://github.com/eic/epic/pull/527/files)? Probably just a matter of running `git rebase origin/main` in your branch, then `git push --force` (but check first with `git log` that this has all relevant commits).

dhevang:
Do you mean to undo the little changes to the files compact/fields/beamline_*.xml?  
I meant those to part of this commit but I can undo them if you want.  I meant for 19 files to be changed in this PR.  All but a few are trivial.

wdconinc:
Please keep the changes to code and changes to quantities separate. Makes it easier to determine if there are any changes introduced by the change to the code that are not due to the changes to the quantities.

dhevang:
Ok, done.

dhevang:
@ajentsch, would this implementation of the BxByBz field map work for your purposes in the B0?

ajentsch:
At first glance, it should. I will be able to test it tomorrow with my
field map and give you feedback. Sorry for the delay.

dhevang:
Thanks.  you can use the dumpBfield program to print the field in your location of interest, and confirm that the mapping implementation worked.
My PR to extend the functionality of this in DD4hep was merged into their main branch today.  Not sure when that propagates to what's in the eic-shell though...

ajentsch (approved):
I was able to upload the B0pf field map and see DD4HEP successfully build it. I will do further tests for the B0 map in a separate PR, but the updates to allow bx, by, bz field maps seems to work nicely.

dhevang:
@wdconinc , how often are DD4hep updates propagated to the eic-shell?  The new dumpBfield is in their master branch as of yesterday.

wdconinc:
> how often are DD4hep updates propagated to the eic-shell?

Normally we update when a new release comes out. I can (and often do) backport patches into eic-shell too (I backported https://github.com/AIDASoft/DD4hep/pull/1161 just yesterday). That often requires that the PR has been merged (so it is stable). Is this about https://github.com/AIDASoft/DD4hep/pull/1170? I can backport that too.

src/FieldMapB.cpp
```diff
@@ -0,0 +1,365 @@
...
+void FieldMapB::fieldComponents(const double* pos, double* field)
+{
+  // coordinate conversion
+  auto p = trans_inv * ROOT::Math::XYZPoint(pos[0], pos[1], pos[2]);
+
+  std::vector<int>  indices = {};
+  std::vector<double> deltas = {};
+  ROOT::Math::XYZPoint B = ROOT::Math::XYZPoint(0,0,0);
+
+  if( coord_type.compare("BrBz") == 0 ) {
```
wdconinc, line 217: This is the single most time-critical function in this entire repository: it is evaluated for every step of every track in every event (the rest of this repository typically just runs on geometry initialization). Therefore I'm going to apply a lot more scrutiny to this, and I apologize in advance. Magnetic field lookups are often among the most CPU intensive aspects of our simulations.

- We should avoid rerunning this string comparison for every step: it should be done once, and we should use a boolean that is a trivial lookup.
- We should not create dynamically allocated `indices` and `deltas`, and `getIndices` should be on statically allocated variables too.
- There should be no need for a std::count over `indices`.
- Avoid the loop over `comp` in the 3D case.

Normally I'd encourage the use of transitions to std::vector. Just not here...

dhevang:
> > how often are DD4hep updates propagated to the eic-shell?
> 
> Normally we update when a new release comes out. I can (and often do) backport patches into eic-shell too (I backported [AIDASoft/DD4hep#1161](https://github.com/AIDASoft/DD4hep/pull/1161) just yesterday). That often requires that the PR has been merged (so it is stable). Is this about [AIDASoft/DD4hep#1170](https://github.com/AIDASoft/DD4hep/pull/1170)? I can backport that too.

Yes

veprbl:
I've tried benchmarking simulation performance for this. I get:

before
```
        Command being timed: "ddsim --runType batch --filter.tracker edep0 -v WARNING --numberOfEvents 100000 --compactFile ...
        User time (seconds): 5558.81
        System time (seconds): 1.14
        Percent of CPU this job got: 99%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 1:32:41
```

after
```
        Command being timed: "ddsim --runType batch --filter.tracker edep0 -v WARNING --numberOfEvents 100000 --compactFile ...
        User time (seconds): 8117.60
        System time (seconds): 1.02
        Percent of CPU this job got: 99%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 2:15:19
```

This is not great, I suppose.

src/FieldMapB.cpp
veprbl: I've tried benchmarking simulation performance for this. I get:

before
```
        Command being timed: "ddsim --runType batch --filter.tracker edep0 -v WARNING --numberOfEvents 100000 --compactFile ...
        User time (seconds): 5558.81
        System time (seconds): 1.14
        Percent of CPU this job got: 99%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 1:32:41
```

after
```
        Command being timed: "ddsim --runType batch --filter.tracker edep0 -v WARNING --numberOfEvents 100000 --compactFile ...
        User time (seconds): 8117.60
        System time (seconds): 1.02
        Percent of CPU this job got: 99%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 2:15:19
```

This is not great, I suppose.

dhevang: Ok, let me see what I can do to streamline the fieldComponents() function.

dhevang: I'm trying to understand which part of FieldMapB.cpp is causing the main slow down.
Most particles don't go all the way down to the FB area where the new lumi dipole field map exists.
Will a particle a mid rapidity call on the fieldComponents() for the FB area or just in the area that it traverses?

wdconinc:
DumpBField changes backported and merged into eic-shell. Should get to cvmfs in the next 12 hours.

src/FieldMapB.cpp
dhevang: @veprbl , thanks for checking the simulation performance time.  Can you check it again after my latest commit.  Hopefully it has been improved.

compact/far_backward/lumi/lumi_magnets.xml
```diff
@@ -77,20 +77,54 @@ Construct the sweeper and analyzer dipole magnets for the luminosity subsystem.
   Detailed field map to be inserted later.
   </comment>
 
+
   <fields>
-    <field name="LumiSweeperField" type="MultipoleMagnet">
+
+    <comment> Ideal fields </comment>
+    <!--<field name="LumiSweeperField" type="MultipoleMagnet">
```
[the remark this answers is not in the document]
dhevang: done.

src/FieldMapB.cpp
veprbl: Seems like we are back to normal:
```
        Command being timed: "ddsim --runType batch --filter.tracker edep0 -v WARNING --numberOfEvents 100000 --compactFile ...
        User time (seconds): 5715.81
        System time (seconds): 1.32
        Percent of CPU this job got: 89%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 1:46:54
```
Note there was some extra load on the machine, so look at the user time

src/FieldMapB.cpp
```diff
@@ -0,0 +1,377 @@
...
+  void SetTransform(const Transform3D& tr)
+  {
+    trans     = tr;
+    trans_inv = tr.Inverse();
+  }
+
+  virtual void fieldComponents(const double* pos, double* field);
+
+private:
+  bool                                                          case_BrBz; // is it BrBz or not (BxByBz)
```
wdconinc, line 70: Can we use `enum` and switch case (the "case" hints at that too)?

src/FieldMapB.cpp
```diff
@@ -0,0 +1,377 @@
...
+  }
+
+  // set type
+  if (ftype == "magnetic") {
+    DD4HEP_FIELD_TYPE = CartesianField::MAGNETIC;
+  } else if (ftype == "electric") {
+    DD4HEP_FIELD_TYPE = CartesianField::ELECTRIC;
+  } else {
+    DD4HEP_FIELD_TYPE = CartesianField::UNKNOWN;
+    std::cout << "FieldMapB Warning: Unknown field type " << field_type_str << "!" << std::endl;
```
wdconinc, line 96: Use dd4hep `printout` logger instead of `std::cout` (I know that I am to blame for this one, but we might as well fix it now). Applies throughout.

src/FieldMapB.cpp
```diff
@@ -0,0 +1,377 @@
...
+    Bvals_RZ.resize(nr);
+    for (auto& B2 : Bvals_RZ) {
+      B2.resize(nz);
+      for (auto& B : B2) {
+        B.resize(2, 0.); // Br, Bz
+      }
+    }
+  }
+  else {
+    int nx = int((maxs[0] - mins[0]) / steps[0]) + 2;
```
wdconinc, line 130: Is this not wrong or at least a bit fragile since it depends on truncation of potentially incorrect floating point representation? If `maxs = mins + steps` then we naively get `int(1) + 2 = 3` when we should get a size of 2 for the 2 grid points. Overcount is better than undercount when prepping for a resize, but can we fix this with roundf, remquof, or use a modf with iptr arg?

src/FieldMapB.cpp
```diff
@@ -0,0 +1,377 @@
...
+// get RZ cell indices corresponding to point of interest
+bool FieldMapB::GetIndices(double R, double Z, int *idxR, int *idxZ, double *deltaR, double *deltaZ)
+{
+  // boundary check
+  if( R > maxs[0] || R < mins[0] || Z > maxs[1] || Z < mins[1] ) {
+    return false;
+  }
+
+  // get indices
+  *deltaR = std::modf( (R - mins[0]) / steps[0], &ir_d );
```
wdconinc, line 156: Could also use remquof here.

src/FieldMapB.cpp
```diff
@@ -0,0 +1,377 @@
...
+  }
+}
+
+// get field components
+void FieldMapB::fieldComponents(const double* pos, double* field)
+{
+  // coordinate conversion
+  auto p = trans_inv * ROOT::Math::XYZPoint(pos[0], pos[1], pos[2]);
+
+  if( case_BrBz ) {
```
wdconinc, line 239: Use switch case on enum here for scalability to future BrBzBphi description of MARCO I imagine.

src/FieldMapB.cpp
```diff
@@ -0,0 +1,377 @@
...
+    // coordinates conversion
+    const double r   = sqrt(p.x() * p.x() + p.y() * p.y());
+    const double z   = p.z();
+    const double phi = atan2(p.y(), p.x());
```
wdconinc, line 243: Only calculate phi when we know we need it, after we have gotten indices.

src/FieldMapB.cpp
```diff
@@ -0,0 +1,377 @@
...
+      double b0 = b00 * (1 - dy) + b10 * dy;
+      double b1 = b01 * (1 - dy) + b11 * dy;
+      // Finally along Z
+      b[comp] = b0 * (1 - dz) + b1 * dz;
+    }
+
+    B_interpolated = ROOT::Math::XYZPoint( b[0], b[1], b[2] );
+  }
+
+  field[0] += B_interpolated.x() * tesla;
```
wdconinc, line 297: This unit conversion shouldn't be necessary to do on the interpolated value and should already have been done on reading in and storing values in the grid.

src/FieldMapB.cpp
```diff
@@ -0,0 +1,377 @@
...
+      coord.resize(2);
+      Bcomp.resize(2);
+      iss >> coord[0] >> coord[1] >> Bcomp[0] >> Bcomp[1];
+
+      if( ! GetIndices( coord[0], coord[1], &ir, &iz, &dr, &dz) ) {
+        std::cout << "FieldMapB Warning: coordinates out of range, skipped it."<< std::endl;
+      }
+      else {
+        Bvals_RZ[ ir ][ iz ] = { Bcomp[0] * scale, Bcomp[1] * scale };
+        //ROOT::Math::XYZPoint p(coord[0], 0, coord[1]);
```
wdconinc, line 213: Remove commented lines or turn into debug printout.

src/FieldMapB.cpp
```diff
@@ -0,0 +1,377 @@
...
+  virtual void fieldComponents(const double* pos, double* field);
+
+private:
+  bool                                                          case_BrBz; // is it BrBz or not (BxByBz)
+  Transform3D                                                   trans, trans_inv; // transformation matrices
+  std::vector<double>                                           steps, mins, maxs; // B map cell info
+  int                                                           ir, ix, iy, iz; // lookup indices
+  double                                                        ir_d, ix_d, iy_d, iz_d; // transient indices for std::modf
+  double                                                        dr, dx, dy, dz; // deltas for interpolation
+  std::vector<std::vector<std::vector<double>>>                 Bvals_RZ; // B map values:  {R}, {Z}, {Br,Bz}
```
wdconinc, line 76: It may make sense to evaluate changing the innermost vector to a std::array of 2 (or 3). Those are smaller and faster, and we'll always pick up the two (three) values with a single read, and likely the next grid point too. With the overhead of the vector that might not be true.

It may also make sense to use float instead of double. The precision here is almost certainly not at the double precision level.

src/FieldMapB.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+    // p1    p3
+    //    p
+    // p0    p2
+    auto& p0 = Bvals_RZ[ ir     ][ iz    ];
+    auto& p1 = Bvals_RZ[ ir     ][ iz + 1];
+    auto& p2 = Bvals_RZ[ ir + 1 ][ iz    ];
+    auto& p3 = Bvals_RZ[ ir + 1 ][ iz + 1];
+
+    // Bilinear interpolation
```
wdconinc, line 258: As an optional exercise in a future pr, it may be fun to see how much of this can be 1) modified to use fma, fused multiply-add (single CPU op) or how much the compiler already figures that out, 2) use veccore or std:: experimental::simd types to parallelize the two or three vector components that need the same operation anyway. Or, of course, one would start with figuring out how much it actually matters...

src/FieldMapB.cpp
```diff
@@ -0,0 +1,377 @@
...
+      double b00 = Bvals_XYZ[ ix     ][ iy     ][ iz     ][comp] * (1 - dx)
+                 + Bvals_XYZ[ ix  + 1][ iy     ][ iz     ][comp] * dx;
+      double b01 = Bvals_XYZ[ ix     ][ iy     ][ iz  + 1][comp] * (1 - dx)
+                 + Bvals_XYZ[ ix  + 1][ iy     ][ iz  + 1][comp] * dx;
+      double b10 = Bvals_XYZ[ ix     ][ iy  + 1][ iz     ][comp] * (1 - dx)
+                 + Bvals_XYZ[ ix  + 1][ iy  + 1][ iz     ][comp] * dx;
+      double b11 = Bvals_XYZ[ ix     ][ iy  + 1][ iz  + 1][comp] * (1 - dx)
+                 + Bvals_XYZ[ ix  + 1][ iy  + 1][ iz  + 1][comp] * dx;
+      // Next along Y, along 2 lines
+      double b0 = b00 * (1 - dy) + b10 * dy;
```
wdconinc, line 288: Related to the fma comment above, this is really identical to `(b10 - b00) * dy + b00` so some fast linerp codes don't store the corner grid points in memory, but convert upon reading the grid points to these coefficients on a cell basis, i.e. c0 = b10 - b00, c1 = b00. Then it becomes a fma(dy, c0, c1). With a 4 float simd register which overloads fma (and simd instructions likely contain an overloaded fma CPU op) this would turn into a single CPU op for the full linerp.

Again, this is not within scope here, but I warned you this is a critical function ;-)

src/FieldMapB.cpp
```diff
@@ -0,0 +1,377 @@
...
+      }
+    }
+  }
+}
+
+// get field components
+void FieldMapB::fieldComponents(const double* pos, double* field)
+{
+  // coordinate conversion
+  auto p = trans_inv * ROOT::Math::XYZPoint(pos[0], pos[1], pos[2]);
```
wdconinc, line 237: I think trans and trans_inv can be done on reading in the map, right? We'd just apply it to the relevant mins, maxs for shifts.

wdconinc (commented):
Some comments, several out of scope of the PR but some relevant.

src/FieldMapB.cpp
wdconinc: (this might actually be optimized correctly by the compiler)

src/FieldMapB.cpp
dhevang: For translation transformations only, yes.  Including rotations, which is an option in the steering xml, I'm not sure.
That brings me to another question I had earlier about this line:
https://github.com/eic/epic/blob/c16103002ee38090a6ab033234b2bd544ed75349/src/FieldMapB.cpp#L267
Isn't this wrong?  We use trans_inv on the coordinates, then we use trans on the field.
Shouldn't we only have the rotation part of the transformation applied here?  Otherwise it's like translating back and forth.  In fact, it gave the wrong field values for the new BxByBz case, where I have a non-zero transformation (unlike the MARCO field).  That's why I didn't use the transformation on the BxByBz field.

src/FieldMapB.cpp
dhevang: Good idea.  done.

src/FieldMapB.cpp
dhevang: I now use remquof and with it, I removed the overcounting: +2 --> +1.

src/FieldMapB.cpp
dhevang: I've never looked this much into optimization, although I'm curious how it could be done.
So, is the point that a long operation, i.e. a*b + c*d + ..., won't fit into a single CPU op, which means it takes longer? 
What kinds of operations fit into one operation?

src/FieldMapB.cpp
dhevang: Same here, not sure how to optimize based on fma.  Can you point me to some examples online?

src/FieldMapB.cpp
veprbl: I think, it's a bug. Magnetic field vector should be only rotated. Perhaps, it was only working as long as translation was:
https://github.com/eic/epic/blob/1202217cde2bae1a0cfe521c65883bff1d035da3/compact/fields/marco.xml#L14

src/FieldMapB.cpp
veprbl: I'm also having difficulty understanding this comment. I believe @wdconinc 's suggestion is not about trying explicitly coding FMA with intrinsics, but instead to reshuffle the memory access to allow the compiler to achieve certain optimization. If that's a correct interpretation, then this line is not the subject, but rather one need to look to achieve FMA for calculations of bXX themselves by precomputing differences in Bvals along x direction.

In any case, we'd want to ensure the correctness of this code first, and do fine optimizations as needed, while looking at performance benchmarks and the output assembly.

src/FieldMapB.cpp
veprbl: Yeah, it's such a basic optimization. I'm sure it's happening. Let's mark resolved.

src/FieldMapB.cpp
veprbl: Yes, we should multiply by tesla right when we read the file.

veprbl: In `LoadMap`

src/FieldMapB.cpp
wdconinc: Yeah, all this is just me geeking on the ability for low-level optimizations by preparing the problem to allow it. This is out of scope here and should only be done if it matters.

src/FieldMapB.cpp
dhevang: Done.

src/FieldMapB.cpp
dhevang: One more item to possibly optimize:
https://github.com/eic/epic/blob/9f53900013851663002bd88f096107e58508748a/src/FieldMapB.cpp#L231
It was mentioned that the field map could be transformed once during the load, and then remove this from critical fieldComponents() function.  If we only consider translations of the field map only, then it's simple.  However, things can be tricky with rotation transformations.  What is to be rotated?  The 3D position coordinates, or the field, or both?

src/FieldMapB.cpp
veprbl: That is coming from
https://github.com/eic/epic/blob/19ec3f38d53ff21b4ebb0646670a015a878927ca/src/FieldMapBrBz.cpp#L182
?

If there is a mistake, we should be seeing some azimuthal angle modulations in tracking performance.

veprbl: Removing transformation is not entirely correct. You are still supposed to apply the rotation.

dhevang: Yes, it's coming from that line.  
Br is very small in the central region, so perhaps the error on this went undetected.
Yes, we still should allow for a rotation of the field but I am unclear on how we want to do that.  From the other unresolved conversation above, do we want to rotate the field vector alone?  If so, we would have translations to move around our field region, and rotations to rotate the field vector within that translated region.

veprbl: > Br is very small in the central region, so perhaps the error on this went undetected.

Would you mind submitting a fix separately against main, so we can review and assess impact separately?

> Yes, we still should allow for a rotation of the field but I am unclear on how we want to do that. From the other unresolved conversation above, do we want to rotate the field vector alone? If so, we would have translations to move around our field region, and rotations to rotate the field vector within that translated region.

Inverse translation and inverse rotation are applied to the coordinate. Direct rotation is applied to the magnetic field. We could name those `coord_trans` and `vec_trans` respectively.

src/FieldMapB.cpp
dhevang: I don't see a simple way to move the field transformation to the load stage while still keeping the same number of calculations in GetIndices().
I suppose that this (3x3) x (3x1) multiplication is optimized automatically by the compiler.  So maybe it's ok to leave this coordinate transformation here as it is?

src/FieldMapB.cpp
dhevang: I now have a separate translation and rotation operation in place.
I'll issue a separate fix for the potential X and Y swapping.

dhevang:
Are there any other optimizations needed for this PR? @ajentsch did some checks using some new field maps in the B0 region, using the new BxByBz option.  Things looked good from his perspective.

veprbl (approved):
ddsim performance for single electron

`main`:
```
        User time (seconds): 8500.01
        System time (seconds): 4.50
        Percent of CPU this job got: 99%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 2:21:45
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 1963884
```
this branch:
```
        User time (seconds): 8381.21
        System time (seconds): 4.38
        Percent of CPU this job got: 99%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 2:19:47
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 1952888
```

capybara report for reco outputs: https://veprbl.github.io/capybara-reports/f9794437ad70f939c89b642899035e99/

src/FieldMapB.cpp
dhevang: @veprbl , @wdconinc , can we resolve this discussion of performance for this PR?  From the recent benchmark test that Dima just ran (thanks), it appears that there is no slow down with the changes.

veprbl:
@kkauder feel free to remove from queue if you are going to look into this

PR #528, opened 2023-09-18
FileLoader improvements

veprbl:
Fixes some edge cases, improves usability.

FileLoader: improve usage
---
Remove line starting with "name:" to not suggest that it's a parseable parameter.
Print which entry was not recognized during parsing.
Use cerr.

FileLoaderHelper: fix for when cache is a relative path
---
Previously

    mkdir ./cache/sub/dir -p
    echo q > cache/sub/dir/33b980fae79f456c
    geoPluginRun -compact $DETECTOR_PATH/$DETECTOR_CONFIG.xml -plugin epic_FileLoader file:bar/baz url:https://example.com/ cache:./cache

Would result in

    FileLoader       INFO  checking ./cache/./cache/sub
    FileLoader       INFO  checking ./cache/./cache/sub/dir

Now ./cache produces a symlink in bar:

    33b980fae79f456c -> ../cache/sub/dir/33b980fae79f456c

but $PWD/cache produces a symlink

    33b980fae79f456c -> /path/to/cache/sub/dir/33b980fae79f456c

FileLoaderHelper: remove invalid hashed symlinks
---
If an entry from the cache directory disappears, the loader will fail to recreate a symlink like so:

    FileLoader       INFO  file bar/baz with hash 33b980fae79f456c found in /home/veprbl/epic/cache/sub/dir/33b980fae79f456c
    FileLoader       ERROR unable to link from bar/33b980fae79f456c to /home/veprbl/epic/cache/sub/dir/33b980fae79f456c
    FileLoader       ERROR hint: this may be resolved by removing directory bar
    FileLoader       ERROR hint: or in that directory removing the file or link /home/veprbl/epic/cache/sub/dir/33b980fae79f456c

or fail to download like so:

    FileLoader       INFO  downloading bar/baz as hash 33b980fae79f456c with curl --retry 5 -f https://example.com/ -o bar/33b980fae79f456c
      % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                     Dload  Upload   Total   Spent    Left  Speed
      0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0Warning: Failed to open the file bar/33b980fae79f456c: No such file or
    Warning: directory
    100  1256  100  1256    0     0  21830      0 --:--:-- --:--:-- --:--:-- 22035
    curl: (23) Failure writing output to destination
    FileLoader       ERROR unable to run the download command curl --retry 5 -f https://example.com/ -o bar/33b980fae79f456c
    FileLoader       ERROR the return value was
    FileLoader       ERROR hint: check the command and try running manually
    FileLoader       ERROR hint: allow insecure connections on some systems with the flag -k

FileLoaderHelper: allow finding hash directly in top-level cache directory
---
Previously it would not work to do `echo q > cache/33b980fae79f456c`,
because the recursive_directory_iterator starts at depth=1. An
explicit check on the top-level is needed.

### DIFF
--- a/src/FileLoaderHelper.h
+++ b/src/FileLoaderHelper.h
@@ -92,12 +92,12 @@ inline void EnsureFileFromURLExists(std::string url, std::string file, std::stri
             // symlink hash to cache/.../hash
             printout(INFO, "FileLoader",
                      "file " + file + " with hash " + hash + " found in " + cache_hash_path.string());
-	    fs::path link_target;
-	    if (cache_hash_path.is_absolute()) {
-	      link_target = cache_hash_path;
-	    } else {
-	      link_target = fs::proximate(cache_hash_path, parent_path);
-	    }
+            fs::path link_target;
+            if (cache_hash_path.is_absolute()) {
+              link_target = cache_hash_path;
+            } else {
+              link_target = fs::proximate(cache_hash_path, parent_path);
+            }
             try {
               fs::create_symlink(link_target, hash_path);
               success = true;
@@ -108,9 +108,9 @@ inline void EnsureFileFromURLExists(std::string url, std::string file, std::stri
             }
             return true;
           }
-	  return false;
+          return false;
         };
-	if (!check_path(cache_path)) {
+        if (!check_path(cache_path)) {
           for (auto const& dir_entry : fs::recursive_directory_iterator(cache_path)) {
             if (!dir_entry.is_directory())
               continue;
--- a/src/FileLoaderHelper.h
+++ b/src/FileLoaderHelper.h
@@ -72,6 +72,11 @@ inline void EnsureFileFromURLExists(std::string url, std::string file, std::stri
     return;
   }
 
+  if (fs::exists(fs::symlink_status(hash_path)) && !fs::exists(fs::status(hash_path))) {
+    printout(INFO, "FileLoader", "removing broken \"" + hash_path.string() + "\" symlink");
+    remove(hash_path);
+  }
+
   // if hash does not exist, we try to retrieve file from cache
   if (!fs::exists(hash_path)) {
     // recursive loop into cache directories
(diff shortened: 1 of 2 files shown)


wdconinc:
Can we make url a comma separated list? I am really dying for that option just about now...

I hadn't added a sha checksum because of the additional dependency it would require. Even a secondary url without checksumming would help us now.

wdconinc:
Then again, we can't roll out a new epic feature in the containers until we have a working eicweb, so moot point.

veprbl:
>I hadn't added a sha checksum because of the additional dependency it would require. Even a secondary url without checksumming would help us now.

There is an implementation of CRC in boost, but I'd really prefer a cryptographic hash. OpenSSL should be universally available, although the dependency would be a bit weird. There is sha1sum which is part of coreutils, but on macOS we'd have to use shasum.

wdconinc:
This is getting out of scope for this PR, but we could use libcurl instead of a system call (which requires openssl libcrypto anyway). Surprisingly there doesn't seem to be any support in libcurl for determining the cryptographic hash while downloading a file...

src/FileLoaderHelper.h
```diff
@@ -72,6 +72,11 @@ inline void EnsureFileFromURLExists(std::string url, std::string file, std::stri
     return;
   }
 
+  if (fs::exists(fs::symlink_status(hash_path)) && !fs::exists(fs::status(hash_path))) {
+    printout(INFO, "FileLoader", "removing broken \"" + hash_path.string() + "\" symlink");
+    remove(hash_path);
+  }
```
wdconinc, line 78: This will make a number of people happy and they won't even know it. @ajentsch

src/FileLoaderHelper.h
```diff
@@ -80,27 +85,38 @@ inline void EnsureFileFromURLExists(std::string url, std::string file, std::stri
...
-            break;
+            return true;
+          }
+          return false;
+        };
+        if (!check_path(cache_path)) {
+          for (auto const& dir_entry : fs::recursive_directory_iterator(cache_path)) {
+            if (!dir_entry.is_directory())
+              continue;
+            if (check_path(dir_entry.path())) {
```
wdconinc, line 117: Why don't we need `fs::path cache_dir_path = cache_path / dir_entry;` anymore in the call to check_path?

wdconinc (approved):

ajentsch:
Oh, I will know it. (Thanks, Wouter)

wdconinc:
> Oh, I will know it. (Thanks, Wouter)

Thanks goes to @veprbl

ajentsch:
Thanks, @veprbl!

src/FileLoaderHelper.h
veprbl: Because, previously it was not quite effective: `fs::path("path") / fs::path("/absolute/path") == fs::path("/absolute/path")`

PR #529, opened 2023-09-23
Update solenoid material budget

johnlajoie:

This PR updates the ePIC Solenoid (MARCO) to the material specifications provided by the design team located at:

https://wiki.bnl.gov/EPIC/index.php?title=Experimental_Solenoid

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [X] Other: Update solenoid material map

No breaking changes 

No, it just updates the material budget for the MARCO solenoid.

This PR is a response to a material map run for me by Nicholas Schmidt at ORNL - you can see below that the solenoid is almost an interaction length, when it should be closer to 0.5 interaction lengths according to the designers:

image

The new material map reduces the material budget for the solenoid to be a bit closer to what we would expect:

image

image

Fully updated based on the material map from Renuka and Valerio dated 17 Sept 2023. The endcaps got a bit thicker, but most of the barrel is essentially the same. I removed the He layers that were in the original based on a comment from Valerio:

"The total volume of helium circulating inside the pipes is just 6000 mm3, corresponding to an uniform layer of Helium of 0.05 mm around the magnet. We neglected it for this reason, but if you think it is important we can add it."

[Material_in_the_magnet_Marco_2T_September.17.2023 (1).xlsx](https://github.com/eic/epic/files/12705405/Material_in_the_magnet_Marco_2T_September.17.2023.1.xlsx)

### DIFF
--- a/compact/materials.xml
+++ b/compact/materials.xml
@@ -478,18 +478,18 @@
     <D type="density" value="8.41" unit="g / cm3"/>
     <fraction n="0.70" ref="Cu"/>
     <fraction n="0.30" ref="Zn"/>
-  </material> 
+  </material>
   <material name="NbTi">
     <D type="density" value="6.54" unit="g / cm3"/>
     <fraction n="0.50" ref="Nb"/>
     <fraction n="0.50" ref="Ti"/>
-  </material> 
+  </material>
   <material name="Solder">
     <D type="density" value="11.35" unit="g / cm3"/>
     <fraction n="0.975" ref="Pb"/>
     <fraction n="0.015" ref="Sn"/>
     <fraction n="0.010" ref="Ag"/>
-  </material> 
+  </material>
  <material name="SolenoidCoil">
     <D type="density" value="8.73" unit="g / cm3"/>
     <fraction n="0.9201" ref="Cu"/>
--- a/compact/materials.xml
+++ b/compact/materials.xml
@@ -485,10 +485,9 @@
     <fraction n="0.50" ref="Ti"/>
   </material>
   <material name="Solder">
-    <D type="density" value="11.35" unit="g / cm3"/>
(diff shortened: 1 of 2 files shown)


compact/materials.xml
```diff
@@ -474,4 +474,26 @@
...
+    <fraction n="0.30" ref="Zn"/>
+  </material>
+  <material name="NbTi">
+    <D type="density" value="6.54" unit="g / cm3"/>
+    <fraction n="0.50" ref="Nb"/>
+    <fraction n="0.50" ref="Ti"/>
+  </material>
+  <material name="Solder">
+    <D type="density" value="11.35" unit="g / cm3"/>
+    <fraction n="0.975" ref="Pb"/>
```
wdconinc, line 489: This seems very lead-heavy for electronic solder. Is this standard solder or should we give it a more specific name so no one uses this in some maps tracker or so, thinking it's mostly tin.

wdconinc (commented):
Looks good to me.

I am a bit confused by the absence of helium. We didn't include it because of the weight or material but because of the space it takes up. All layers are defined as in contact, so where is the space where the helium would go? If we don't have this space, then we'll end up with incorrect radii (I think the thickness was 2 x 2 cm, so without this we'd have an outer radius 4 cm smaller than it should be).

compact/materials.xml
johnlajoie: Good catch, this should be standard 60/40 electronics solder and I had specialty values. The solder definition has been updated. 

As I understood Valerio's comment the He layer is only 50um thick surrounding the coil so it was neglected in their Excel table (see the comment at the end of the first post).  The Excel table explicitly lists the inner and outer radii for each layer so we do go from 2.84-3.54m.

Chao1009 (approved):

PR #533, opened 2023-09-24
linux-eic-shell.yml: remove workflow_dispatch functionality

veprbl:
Should be unused, since we don't have reverse dependencies (like ip6 repo) on GitHub anymore.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -8,6 +8,7 @@ on:
       - '*'
   pull_request:
   merge_group:
+  workflow_dispatch: # allow manual triggering
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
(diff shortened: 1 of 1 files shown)


.github/workflows/linux-eic-shell.yml
```diff
@@ -8,18 +8,6 @@ on:
       - '*'
   pull_request:
   merge_group:
-  workflow_dispatch:
```
wdconinc, line 11: I'd keep this one line, and remove the dependency repository info. Just having a workflow dispatch trigger gives us a run button at https://github.com/eic/epic/actions/workflows/linux-eic-shell.yml which has been useful at times.

wdconinc (approved):

.github/workflows/linux-eic-shell.yml
veprbl: Sure, added with an inline comment.

wdconinc (approved):

PR #536, opened 2023-09-25
changed vector causing mem leak to vector of DetElements, rather than pointers

ajentsch:

Fixes memory leak caused be erroneous missing "deletes" for vector of DetElement*. DetElement changed to actual object instead of pointer.

- [x] Bug fix (issue # 535)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No - fixed a bug.

No, just fixes the bug. Geometry renders the way it should.

### DIFF
--- a/src/magnetVacuumFF.cpp
+++ b/src/magnetVacuumFF.cpp
@@ -30,209 +30,209 @@ static double getRotatedX(double z, double x, double angle);
 static Ref_t create_detector(Detector& det, xml_h e, SensitiveDetector /* sens */)
 {
 
-        xml_det_t x_det    = e;
-        string    det_name = x_det.nameStr();
-        DetElement sdet(det_name, x_det.id());
-        Assembly   assembly(det_name + "_assembly");
-        Material   m_Vac    = det.material("Vacuum");
-        string     vis_name = x_det.visStr();
+    xml_det_t x_det    = e;
+    string    det_name = x_det.nameStr();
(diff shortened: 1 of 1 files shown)


src/magnetVacuumFF.cpp
```diff
@@ -242,34 +242,37 @@ static Ref_t create_detector(Detector& det, xml_h e, SensitiveDetector /* sens *
...
+                auto pv = assembly.placeVolume(specialGap_v, Transform3D(RotationY(crossingAngle), Position(specialGap_x, 0.0, specialGap_z)));
+                pv.addPhysVolID("sector", 1);
 
-                        DetElement * tmp = new DetElement(sdet, Form("sector%d_de", numGaps + numMagnets), 1);
-                        detectorElement.push_back(tmp);
+                DetElement tmp(sdet, Form("sector%d_de", numGaps + numMagnets), 1);
+                detectorElement.push_back(tmp);
 
-                        detectorElement[numGaps + numMagnets]->setPlacement(pv);
+                detectorElement[numGaps + numMagnets].setPlacement(pv);
```
Chao1009, line 262: The code always accesses the last element of the detector, so the vector itself seems not necessary.
The following code should've done the same job.
`DetElement tmp(sdet, Form("sector%d_de", numGaps + numMagnets), 1);`
'tmp.setPlacement(pv);`

wdconinc (approved):

PR #538, opened 2023-09-30
feat(ci): valgrind check for memory leaks

wdconinc:
This adds a new ci check that runs valgrind on the main geometries, and fails out if non-suppressed errors are found. It also prints out the additions to the suppressions file that would be required if it turns out that the error is a false positive.

This branch is intentionally on a non-updated main so it should be able to fail due to a known memory leak that was since fixed.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


wdconinc:
> @wdconinc Is this going in for February?

This is a CI feature. Irrelevant for production and won't go in soon due to not completed.

Issue #546, opened 2023-10-05
HCal forward insert - increase number of W/Sc layers to 10


rymilton:
**Is your feature request related to a problem? Please describe.**
The current implementation of the Hcal insert has 4 layers of W/Sc, similar to the LFHCAL. The actual design should have 10 layers of W/Sc.

**Describe the solution you'd like**
Increase the number of W/Sc layers to 10 and decrease the number of Steel/Sc layers from 60 to 54 (remove 6 layers of steel, add 6 layers of tungsten).

**Describe alternatives you've considered**
None.

**Additional context**

Chao1009:
Closes with #547

PR #547, opened 2023-10-05
feat - Increase number of W/Sc layers in forward HCal insert from 4 to 10.

rymilton:
There are currently 4 layers of W/Sc in the forward HCal insert to match the layering of the surrounding LFHCAL. The correct layer structure for the HCal insert should be 1 front layer of Steel/Sc, 10 layers of W/Sc, 54 layers of Steel/Sc, 1 back layer of steel.

- [ ] Bug fix (issue #__)
- [X] New feature (issue #546 )
- [ ] Documentation update
- [ ] Other: __

None.
Only the changes to the HCal insert listed above.


Chao1009 (approved):
Looks good.

PR #559, opened 2023-10-11
fix: BackwardServiceGap_length:=124cm, HcalEndcapN_length:=45cm

wdconinc:
And we start to address the mismatches in https://github.com/eic/epic/pull/553. This one should be mostly isolate from other parts.

- [x] Bug fix (issue: BackwardServiceGap_length is larger)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators @lkosarz 

No.

No.

### DIFF
--- a/compact/definitions.xml
+++ b/compact/definitions.xml
@@ -500,8 +500,8 @@ Service gaps in FW direction (before endcapP ECAL) and BW direction (before endc
     <constant name="ForwardServiceGap_rmin"       value="ForwardPIDRegion_rmax"/>
     <constant name="ForwardServiceGap_rmax"       value="ForwardServiceGap_rmin + ForwardServiceGap_thickness"/>
 
-    <constant name="BackwardServiceGap_length"    value="10.0 * cm"/>
-    <constant name="BackwardServiceGap_zmin"      value="320.0 * cm"/> <!-- FIXME hardcoded -->
+    <constant name="BackwardServiceGap_length"    value="124.0 * cm"/>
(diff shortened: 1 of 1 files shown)


wdconinc:
Ref: image

wdconinc:
All benchmarks passed, no overlaps. Should be ready for merging.

Chao1009 (approved):

wdconinc:
@Chao1009 That doesn't seem to have made sense:
image

wdconinc:
This might have been another mistake in the detector parameter table.

Chao1009:
I also noticed the `Length` and `Physical Start` seem not right in #561 , they were much smaller before the recent PRs.
It is caused by the increase of `BackwardServiceGap_length` from 10.0 cm to 124 cm, maybe this is a typo (12.4 cm) in the table?

wdconinc:
Hmm that could be it indeed.

lkosarz:
Sorry for late response. This doesn't make sense indeed. First of all, the position should be negative - backward. I initially was confused, because with the values in the table below it would actually contain the backward HCal. My understanding is that the service gap is supposed to be in front to lead the cables out. We should figure out or correct the parameter table in the first place.

lkosarz:
Also, the 12.4 cm length of the gap surprisingly resembles the value for inner radius of the HCal.

Chao1009:
@wdconinc @lkosarz 
Tanja has fixed the gap service length in the latest version of the table
https://eic.jlab.org/Geometry/Detector/Detector-20231020152234.html
it is now 15 cm

lkosarz:
Good, thanks!

Issue #561, opened 2023-10-12
Detector Parameters Update: Barrel Imaging Calorimeter


Chao1009:
A series of smaller issues from #552 
`det`: new value from 2023/09/27 parameter table
`sim`: current simulation value as of 2023/10/11
`template_var`: from which we extract the sim value in compact files. `Empty` means we do not know.
`stat`: `Correct`, `Missing`, or `Mismatched`

Please implement the new detector parameters, and try to fill in the `template_var` for the `Missing` ones if you think it's important for future simulation/design comparison.

Here is the detailed report for this detector subsystem:

```
CENTRAL DETECTOR, Barrel EMCal, : 
                            det    sim                  template_var        stat
Length (cm)               470.0  635.0         {{EcalBarrel_length}}  Mismatched
Inner Radius (cm)          81.0   78.5           {{EcalBarrel_rmin}}  Mismatched
Outer Radius (cm)         116.0  140.0           {{EcalBarrel_rmax}}  Mismatched
Offset from Center (cm)  -38.75    NaN                         Empty     Missing
Physical Start (cm)     -273.75 -440.0  -{{EcalBarrelBackward_zmax}}  Mismatched
Physical End (cm)        196.25  195.0    {{EcalBarrelForward_zmax}}  Mismatched

CENTRAL DETECTOR, Barrel EMCal, Exterior Cover: 
                            det  sim template_var     stat
Length (cm)               460.0  NaN        Empty  Missing
Inner Radius (cm)         114.0  NaN        Empty  Missing
Outer Radius (cm)         116.0  NaN        Empty  Missing
Offset from Center (cm)  -38.75  NaN        Empty  Missing
Physical Start (cm)     -268.75  NaN        Empty  Missing
Physical End (cm)        191.25  NaN        Empty  Missing

CENTRAL DETECTOR, Barrel EMCal, HD Readout Electronics: 
                            det  sim template_var     stat
Length (cm)                15.0  NaN        Empty  Missing
Inner Radius (cm)          81.0  NaN        Empty  Missing
Outer Radius (cm)         116.0  NaN        Empty  Missing
Offset from Center (cm)  181.25  NaN        Empty  Missing
Physical Start (cm)      181.25  NaN        Empty  Missing
Physical End (cm)        196.25  NaN        Empty  Missing

CENTRAL DETECTOR, Barrel EMCal, Imaging Part: 
                            det  sim template_var     stat
Length (cm)               440.0  NaN        Empty  Missing
[...]

PR #577, opened 2023-10-12
fix: add EcalBarrel vars to template; EcalBarrel_rmin:=81cm; EcalBarrel_thickness:=35cm

wdconinc:
This adds a number of EcalBarrel variables to the detector parameter table template.

- [x] Bug fix (issue #561)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/compact/definitions.xml
+++ b/compact/definitions.xml
@@ -542,8 +542,8 @@ Service gaps in FW direction (before endcapP ECAL) and BW direction (before endc
     <constant name="EcalEndcapN_rmin"               value="9.*cm"/>    <!-- Currently fix value hardcoded -->
     <constant name="EcalEndcapN_rmax"               value="63.*cm"/>   <!-- Currently fix value hardcoded -->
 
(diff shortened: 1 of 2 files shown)


wdconinc:
@Chao1009 Increasing the radius doesn't cause any issues with overlaps in my local tests (tgeo only). The tracking region parametrization will likely need to change, so I use an rmin = max(81cm, determined radius) to make sure we are at 81cm but if the tracking region gets larger than that we still expand outwards.

Chao1009 (approved):
Looks good to me.
Note the failed overlap check seems to be caused by `BarrelCalorimeterSciGlass`. It is an issue with outdated configurations, not the parameters, but we have to fix that before this merge, otherwise, it will fail all the PR checks.

wdconinc:
> we have to fix that (https://github.com/eic/epic/pull/579) before this merge

Rebased after that PR was merged.

Issue #562, opened 2023-10-12
Detector Parameters Update: Central Barrel Hcal


Chao1009:
A series of smaller issues from #552 
`det`: new value from 2023/09/27 parameter table
`sim`: current simulation value as of 2023/10/11
`template_var`: from which we extract the sim value in compact files. `Empty` means we do not know.
`stat`: `Correct`, `Missing`, or `Mismatched`

Please implement the new detector parameters, and try to fill in the `template_var` for the `Missing` ones if you think it's important for future simulation/design comparison.

Here is the detailed report for this detector subsystem:
```
CENTRAL DETECTOR, Barrel Hadron Calorimeter, : 
                           det    sim                  template_var        stat
Length (cm)              570.0  635.6         {{HcalBarrel_length}}  Mismatched
Inner Radius (cm)        180.0  177.0           {{HcalBarrel_rmin}}  Mismatched
Outer Radius (cm)        267.0  270.0           {{HcalBarrel_rmax}}  Mismatched
Offset from Center (cm)    0.0    NaN                         Empty     Missing
Physical Start (cm)     -285.0 -316.0  -{{HcalBarrelBackward_zmax}}  Mismatched
Physical End (cm)        285.0  319.6    {{HcalBarrelForward_zmax}}  Mismatched

CENTRAL DETECTOR, Barrel Hadron Calorimeter, Central Section: 
                           det  sim template_var     stat
Length (cm)              300.0  NaN        Empty  Missing
Inner Radius (cm)        180.0  NaN        Empty  Missing
Outer Radius (cm)        267.0  NaN        Empty  Missing
Offset from Center (cm)  -35.0  NaN        Empty  Missing
Physical Start (cm)     -185.0  NaN        Empty  Missing
Physical End (cm)        115.0  NaN        Empty  Missing

CENTRAL DETECTOR, Barrel Hadron Calorimeter, HD Section: 
                           det  sim template_var     stat
Length (cm)              170.0  NaN        Empty  Missing
Inner Radius (cm)        194.0  NaN        Empty  Missing
Outer Radius (cm)        267.0  NaN        Empty  Missing
Offset from Center (cm)  200.0  NaN        Empty  Missing
Physical Start (cm)      115.0  NaN        Empty  Missing
Physical End (cm)        285.0  NaN        Empty  Missing

CENTRAL DETECTOR, Barrel Hadron Calorimeter, LD Section: 
                           det  sim template_var     stat
Length (cm)              100.0  NaN        Empty  Missing
[...]

PR #588, opened 2023-10-26
Updated HCAL parameters and template for comparison with detector table

johnlajoie:
- [X] Bug fix (issue #562 )
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

None. The changes here are contained to the barrel HCAL geometry alone, and then only the size of the outer HCAL mother volume. The only actual effect is on the outer HCAL mother volume, which was reduced in size. 

No. The actual outer HCAL detector geometry is unchanged. 

NOTE: This review found that some of the parameters in the Detector Table are incorrect. These changes will allow the proper values to be calculated from the simulation input, but they won't match the Detector Table until that table is updated.


compact/hcal/barrel_gdml.xml
```diff
@@ -15,11 +15,11 @@
...
-          rmin1="HcalBarrel_rmin"
-          rmin2="ForwardServiceGap_rmax - 5.0"
-          rmax="HcalBarrel_rmax"
-          z1="SolenoidBarrel_length"
-          z2="HcalBarrel_length"
+          rmin1="HcalBarrel_rmin1 - 2.0"
+          rmin2="HcalBarrel_rmin2 - 2.0"
+          rmax="HcalBarrel_rmax + 2.0"
+          z1="HcalBarrel_length1 + 0.01"
+          z2="HcalBarrel_length2 + 0.01"
```
wdconinc, line 22: Are these 'fudge factors' necessary? If so, does that mean that the gdml objects are extending outside the envelope?

johnlajoie: The lengths in the detector table are exactly the lengths of the sectors. Without the fudge factors I had a few tens of micron overlaps between the sector components and the mother volume.  The radial dimensions I didn't look very carefully, that could likely be tighter.

johnlajoie: Latest commit removes length fudge factors after a first pass on updating the parameter tables with Tanja and Walt. I have tightened the fudge factors for the radius but there is still a disagreement for the inner radii and the length of the central section (the definition of which is ambiguous).  Changing this to a draft for now until I can resolve remaining issues with Tanja and Walt.

Chao1009: Can we merge the parameters that have no issues and leave the inner radii and length of the central section to another PR? In this case we can break the big issue (parameter update for HCal) into smaller ones.

johnlajoie: I think we have this close to resolved. I sent Tanja and Walt a set of updated numbers taken directly from the 3D model just yesterday. The problem was that they are working ONLY from a 2D model that has an incorrect model of the barrel HCAL.  You can tell just by looking at it because the plates show flat in the cross section and they are actually tilted.  

I just need to find some time to cross-check what is in the simulations, which I can hopefully find time to do this afternoon. 

[HCAL_Geometry_Summary_from_3D.pdf
[EPIC Envelope - 09-14-2023.pdf](https://github.com/eic/epic/files/13500724/EPIC.Envelope.-.09-14-2023.pdf)
](https://github.com/eic/epic/files/13500723/HCAL_Geometry_Summary_from_3D.pdf)

Chao1009 (approved):

wdconinc (changes requested):
We agreed that this needs changes, in particular moving the large gdml file into the epic-data repository, along with the original STEP file so it can be regenerated later, potentially as a STL or other more generic mesh format. Examples in e.g. https://github.com/AIDASoft/DD4hep/blob/master/examples/DDCAD/compact/Check_Shape_PLY_Wuson.xml (for PLY file).

johnlajoie:
Yes - Wouter and I spoke about this at the RRB Meeting. To my mind, what I would like to do is stop storing the mesh in the XML files and just store the filename of a gdml file.  The gdml file would be stored in some standard location accessible by the sims.  The shape could then be created from the gdml in a few simple lines. 

For now we could control things administratively, so we could require that the CAD file and the GDML file be submitted and committed to the repository as a pair, perhaps along with a text file that has some metadata about the conversion. 

Finally, I believe that I can further simplify the mesh for the full tools using the new facet combination tools in gmsh, which would simplify things further. 

I am slammed this week, but the wife is gone next week, so the plan would be to make all these updates then.  Ideally this would be an example for future imports of this type.

johnlajoie:
Latest updates move the HCAL steel tessellated description into a separate GDML file and out of the xml file. Right not the GDML file is in compact/hcal with the XML file, but if it was desirable to move this out of github (large file) that could be done.

johnlajoie:
Completed - the absorber steel mesh is now in a separate gdml file, and that file is part of epic-data and is pulled from epic-data on-demand, the same way the filed map files are pulled in. 

I think this completes the work to reconcile the detector geometry with the detector table (the original intent) as well as fixes up how the large absorber mesh is dealt with.

wdconinc:
Geoviewer error is due to a branch mismatch error in our CI scripts, and rebasing or updating should fix it.

(We create a list of detector configs that should be converted into geoviewer link on GitHub based on what is in configurations. Then we pass that list to eicweb to create the actual geoviewer link. Creating the list is done on the branch after it is test-merged into the target, while eicweb only has access to the branch itself. So, if a new configuration is added in the target branch but not in this branch, then we will get this error.)

compact/definitions.xml
```diff
@@ -601,15 +601,20 @@ Service gaps in FW direction (before endcapP ECAL) and BW direction (before endc
...
     <constant name="HcalEndcapN_rmin"     value="max((HcalEndcapN_zmin + HcalEndcapN_length) * tan(abs(CrossingAngle)) + 22.25 * mm, 14.0 * cm)"/>
 
-    <constant name="HcalBarrel_thickness" value="93.0*cm"/> <!-- ref: as-built CAD, plus a little room for the longer half-plates -->
-    <constant name="HcalBarrel_rmin"      value="Solenoid_rmax"/>
-    <constant name="HcalBarrel_rmax"      value="HcalBarrel_rmin + HcalBarrel_thickness"/>
-    <constant name="HcalBarrelForward_zmax"  value="ForwardServiceGap_zmin"/>
-    <constant name="HcalBarrelBackward_zmax" value="BackwardServiceGap_zmin"/>
-    <constant name="HcalBarrel_length"    value="HcalBarrelForward_zmax + HcalBarrelBackward_zmax"/>
-    <constant name="HcalBarrel_offset"    value="(HcalBarrelForward_zmax - HcalBarrelBackward_zmax)/2"/>
+    <constant name="HcalBarrel_thickness"     value="86.18*cm"/>
```
wdconinc, line 604: Can we add a reference to where these numbers are coming from? A reference to an internal document number is fine too, not suggesting we need to put this on a public document database.

compact/hcal/barrel_gdml.xml
```diff
@@ -15,17 +16,17 @@
...
+          z1="HcalBarrel_length1"
+          z2="HcalBarrel_length2"
           calorimeterType="HAD_BARREL">
 
   <define>
 
-    <matrix name="xposOuter" unit="mm" coldim="1" values="23.6005"/>
-    <matrix name="yposOuter" unit="mm" coldim="1" values="1802.84"/>
+    <matrix name="xposOuter" unit="mm" coldim="1" values="18.400271"/>
+    <matrix name="yposOuter" unit="mm" coldim="1" values="1833.4834"/>
```
wdconinc, line 30: Same here, do we have a reference for these numbers?

compact/hcal/barrel_gdml.xml
```diff
@@ -41,1067 +42,854 @@
...
 
-    <matrix name="plates_x" unit="mm" coldim="1" values="161.935 118.175 -741.606 -771.303"/>
-    <matrix name="plates_y" unit="mm" coldim="1" values="2295.048 2267.198 2160.230 2119.176"/>
-    <matrix name="plates_z" unit="mm" coldim="1" values="235.587 253.54 0.0 0.0"/>
-
-    <!--These offsets are used to tweak the tower and sector positions by ~100um to avoid numerical issues when looking for overlaps-->
-    <matrix name="tweak_tiles" unit="rad" values="0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0"/>
-    <matrix name="tweak_chimney_tiles" unit="rad" values="0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0"/>
-    <matrix name="tweak_sectors" unit="rad" values="0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 -0.00008 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 -0.000008"/>
+    <constant name="gdmlfile" file="gdml/barrel_hcal_steel_full_nocombs.gdml" url="https://github.com/eic/epic-data/raw/main/barrel_HCAL_gdml/barrel_hcal_steel_full_nocombs.gdml" material="Steel1020"/>
```
wdconinc, line 46: I'm surprised this works. I don't think it's compliant with the schema for `constant`, which should only have `name`, `unit` and `value` attributes, and this may stop working on dd4hep upgrades. It would be safer to define a custom xml type that we control. Before `define` it would be e.g. 
```xml
      <detector
          id="HCalBarrel_ID"
          name="HcalBarrel"
          type="epic_HcalBarrelGDML"
          readout="HcalBarrelHits"
          vis="HcalBarrelVis"
          env_vis="HcalBarrelEnvelopeVis"
          rmin1="HcalBarrel_rmin1"
          rmin2="HcalBarrel_rmin2"
          rmax="HcalBarrel_rmax"
          z1="HcalBarrel_length1"
          z2="HcalBarrel_length2"
          calorimeterType="HAD_BARREL">
        <gdmlfile file="gdml/barrel_hcal_steel_full_nocombs.gdml" url="https://github.com/eic/epic-data/raw/main/barrel_HCAL_gdml/barrel_hcal_steel_full_nocombs.gdml" material="Steel1020"/>
```
and then
```cpp
  // Sector steel tessellated shape gdml file info
  std::string gdml_file = getAttrOrDefault<std::string>(x_det, _Unicode(file), " ");
  std::string gdml_material = getAttrOrDefault<std::string>(x_det, _Unicode(material), " ");
  std::string gdml_url = getAttrOrDefault<std::string>(x_det, _Unicode(url), " ");
  std::string gdml_cache = getAttrOrDefault<std::string>(x_det, _Unicode(cache), " ");
```

We also should use a `cache` attribute here (or at least in the plugin) because we don't want to force a download on every run. This should be cached inside the containers.

compact/hcal/barrel_gdml.xml
johnlajoie: These are essentially meaningless.  They are a reference location based on how the tiles are oriented in the STEP file. I can add a comment to that effect.

compact/hcal/barrel_gdml.xml
johnlajoie: The plugin does cache the file - it's only downloaded again if it doesn't exist or has a different hash.  Is this what you mean?

wdconinc: When we run in the container for production campaigns, we want the cache to succeed even the first time it is run, regardless where we run this from, so we pass `cache="$DETECTOR_PATH:/opt/detector"` to these plugins and they go looking in the geometry directories where we have staged the files during the container build. I think we should just do the same here, and it would go look for the gdml file in those directories too.

wdconinc (dismissed):
Per discussion on mattermost, approving this and referring to https://github.com/eic/epic/issues/614 for future implementation.

wdconinc:
I think only change needed is moving from `sector` to `tower` in https://eicweb.phy.anl.gov/EIC/benchmarks/reconstruction_benchmarks/-/blob/master/benchmarks/clustering/options/full_cal_reco.py#L342. Not sure what that impacts downstream.

The juggler reconstruction is only a canary here, and the same will be needed in EICrecon at https://github.com/eic/EICrecon/blob/main/src/detectors/BHCAL/BHCAL.cc#L62.

@ruse-traveler Do you have a way to test in EICrecon what the settings should be changed to there? Can you link the PR here?

ruse-traveler:
> I think only change needed is moving from `sector` to `tower` in https://eicweb.phy.anl.gov/EIC/benchmarks/reconstruction_benchmarks/-/blob/master/benchmarks/clustering/options/full_cal_reco.py#L342. Not sure what that impacts downstream.
> 
> The juggler reconstruction is only a canary here, and the same will be needed in EICrecon at https://github.com/eic/EICrecon/blob/main/src/detectors/BHCAL/BHCAL.cc#L62.
> 
> @ruse-traveler Do you have a way to test in EICrecon what the settings should be changed to there? Can you link the PR here?

I've been working on testing this out: the only things that should need to be updated are the `layerField` and `sectorField` for the `CalorimeterHitReco` and then the corresponding fields in the adjacency matrix for the `CalorimeterIslandCluster`.

I'll open a PR to track the changes in EICrecon.

veprbl:
Not a blocker for this PR, but we should really cut down on the amount of custom geometry code. The pieces of geometry can be loaded from GDML the same way as the steel done here.

veprbl:
With this change, the check-overlap-geant4 takes almost 6 hours instead of 45 min.

johnlajoie:
When we started down this road we had issues with the VecGeom library giving false overlaps. If I recall correctly from my discussions with Wouter we didn't have the compiler flags set for architectures where VecGeom would make a difference so it was falling back on software, and these spurious overlaps were made to go away by disabling VecGeom. 

Is VecGeom still out of the build? And should we try configuring the simulations to use it properly?  I realize this will involve setting some sort of a floor for the hardware required to run the simulations, but the simulations are only going to get more complex with time. If we are not taking advantage of VecGeom now we will likely need to in the future.

wdconinc:
It's unlikely that vecgeom (even with full avx512 on x86_64_v4) will generate factors of speed improvements. I think we can maybe expect, say, 20% speed improvement (in the tesselated volume overlap checks only) since a lot of the code is still not going to be simd-vectorizable (ref Amdahl's law).

The tradeoff is then whether we 1) take on additional risk due to vecgeom having in the past been a source of spurious errors, and 2) kill off some compatibility with older infrastructure. I have a branch of the container that gets us up to v3, not yet v4, and we could expriment with that. We'd lose about 10% of OSG nodes last time @rahmans1 and @T-Britton ran the numbers. (v4 cuts the nodes to only 20% of total, so that's a non-starter right now.)

johnlajoie:
It might be worth some simple experimenting just so we know what the tradeoffs are, and we never determined if the vecgeom issues persisted if we actually tried a path that used the hardware. Of course, this is all easy for me to make work for other people. 

It might also be worth some benchmarking of times to process min-bias DIS events. A large factor time increase in the overlap checks doesn't translate to the same factor for processing events as most particles don't make it to the HCAL.

wdconinc:
> With this change, the check-overlap-geant4 takes almost 6 hours instead of 45 min.

For total DIS 18x275 Q2>1000 physics_benchmarks dis:simulation job:
- https://eicweb.phy.anl.gov/EIC/benchmarks/physics_benchmarks/-/jobs/2416566 (this branch) runs 41 minutes
- https://eicweb.phy.anl.gov/EIC/benchmarks/physics_benchmarks/-/jobs/2451116 (main branch) run 33 minutes

So, not too different. The overlap check is definitely an operational concern for the CI pipelines here. There is [precedent](https://github.com/eic/epic/blob/main/.github/workflows/linux-eic-shell.yml#L101-L105) for taking expensive geometry that is known not to overlap out of the test geometry. That does require that this is easy to do in xml. Haven't evaluated yet here.

rahmans1:
@johnlajoie We are tagging a new release in preparation for February campaign soon. If there is nothing else outstanding, could you please resolve the conflicts and then we can merge it.

johnlajoie:
Thanks, Wouter for the interesting benchmark results!

I believe that I have addressed the merging issues from the TGI Fridays at the Knoxville airport.....

wdconinc:
Is the EICrecon support for this ready now?

ruse-traveler:
> Is the EICrecon support for this ready now?

Yep! eic/EICrecon#1240 is ready to go!

johnlajoie:
I had some time so I completed the transition to the gdml files, the custom geometry code has been removed and the tile solids are no longer stored in the xml file. Instead they are loaded from the gdml files in epic-data, just like the steel sector. 

And f**k me, it looks like prior to this we picked up an overlap with the backwards HCAL.  I'll have to look into that but I think there were some recent updates for the backwards HCAL that got pulled in. Since the barrel HCAL *defines* constraints for many things, because it *exists* it would be good to get this merged once the current conflict is resolved so we are not playing whack-a-mole all the time.

johnlajoie:
Found that I had screwed up an entry in definitions.xml when I had to manually fix a merge last week - fixed!

veprbl:
The 6 hour geant overlap check would not work for the current CI. We could try to split/parallelize this, but before that, I'm seeing a 6x slowdown in single pion simulation:

```bash
ddsim --runType batch --filter.tracker edep0 -v WARNING --numberOfEvents 1000 --compactFile $DETECTOR_PATH/epic_craterlake.xml --outputFile sim.edm4hep.root --enableGun --gun.momentumMin 10*GeV --gun.momentumMax 10.0*GeV --gun.etaMin -1.0 --gun.etaMax 1.0 --gun.particle pi- --gun.distribution uniform
```
*main*
```
	User time (seconds): 241.56
	System time (seconds): 0.53
	Percent of CPU this job got: 99%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 4:02.38
```
*ohcalgeo_reconcile*
```
	User time (seconds): 1561.86
	System time (seconds): 0.54
	Percent of CPU this job got: 99%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 26:05.86
```

This is a comparison for the raw simulation files:
https://veprbl.github.io/capybara-reports/53982ec70d0b5353b70c8447e73c7d3a/#HcalBarrelHits
It would seem like we are loosing information about z position for sensitive volumes.

johnlajoie:
Hi @veprbl - I think there are a few important points to make here with respect to slowdown with this improved geometry: 

- First, a slowdown of a factor of 6 for HCAL showers doesn't translate to a factor of 6 in overall event time, as Wouter has shown. Only a fraction of events put significant energy into the barrel HCAL.  So, the effect on the overall time to run the production campaign will be much less than a factor of six. 

- There is much more material in the outer HCAL with the revised geometry. The previous geometry was just the horizontal plates, this new geometry includes the endplates and "dogbones" on the ends which are a significant amount of additional steel. This material is important to be able to adequately estimate the hadronic energy resolution in the transition region between the barrel HCAL and the LFHCAL. Showering in this material will certainly increase the average event time. 

- We can't just reject changes because the event time increases - improving the realistic description of the detector will increase the event time, so we have to find ways to deal with this, there needs to be a balance. 

- That being said, I do agree that a factor of six is significant and we should dig into it as much as we can. I have some thoughts on this, in that by importing the steel as a while we have one large tessellated object that needs to be searched any time the particle is within the HCAL mother volume. This was done because of the sheer complexity of the structure and the difficulty to recreate it "by hand" within G4. Now, to avoid numerical issues in overlaps the structure was merged into a single solid in CAD before exporting into GDML. It may be possible to do a better job by importing it as a collection of individual solids, which is *MUCH* more complicated and the easiest way to do this would involve some duplication of the geometry, but it might produce an improvement in speed. 

- Another way to attack this would be experiment with vecgeom optimizations. I know Wouter has expressed the opinion that this is unlikely to help but I really think the SIMD instruction support could well produce an improvement. Yes, it is a bit of brute force to solve the issue of more complicated geometry, but as I have noted things will only get worse with ePIC geometry, not better, so some effort in this direction I think is an investment that will pay off more in the long run. My $0.02. 

- We do need to find a strategy to deal with this, and define what level of performance is acceptable, as we plan to use a similar method to import the services from the project CAD model, and we need to get to that as well. 

I'm willing to try to optimize the GDML from the CAD side, but we have been working on getting this into the ePIC simulations since October, and the geometry improvements are not just gratuitous, they are needed to improve the realism of the simulations.  I can do what I can to shake loose some time to dig into this.

veprbl:
@johnlajoie these are all good points. The 6 hour limit is not negotiable limitation on GitHub free runners, and has to be addressed before this is merged. We don't want to disable overlap checking. A very minimal workaround is to create `craterlake_no_bhcal` configuration with HCal taken out (say, we trust TGeo to check envelope overlaps). You can also add configuration with just `bhcal_only`. Then update this line
https://github.com/eic/epic/blob/c760f33210dad360490f1d078fde057d16a9ec2a/.github/workflows/linux-eic-shell.yml#L404
to include
`[epic_craterlake_no_bhcal, bhcal_only]`

This should split a 6 hour job into 1 and 5 hour jobs.



> Finally, as for loosing the z-information of the hit, weird. I'll go into this with @ruse-traveler. This PR changes the hits structure so you do need a matching EICRecon patch to get the hits properly?

No extra changes required in EICrecon. The issue must be in the epic geometry. ddsim records position of the sensitive volume as the hit coordinate. This should not affect clustering, since you rely on cellIDs, but it does break the cluster shape functionality.

johnlajoie:
I broke out the HCAL steel into components, which should dramatically simplify the geometry search.  Using @veprbl's example to test the speed: 

`time ddsim --runType batch --filter.tracker edep0 -v WARNING --numberOfEvents 1000 --compactFile $DETECTOR_PATH/epic_craterlake.xml --outputFile sim.edm4hep.root --enableGun --gun.momentumMin 10*GeV --gun.momentumMax 10.0*GeV --gun.etaMin -1.0 --gun.etaMax 1.0 --gun.particle pi- --gun.distribution uniform` 

on my laptop I get: 

_main_
`real    5m38.349s
user    5m28.342s
sys     0m8.496s`

_ohcalgeo_reconcile_
`real    7m23.240s
user    7m19.975s
sys     0m4.117s`

so we now take about 40% longer with the gdml geometry, which is a pretty good improvement. 

There was no change to the readout structure, so the changes to EICRecon should still be fine and need to modification. I still don't understand the very strange z values for the hits, but that is the next thing to look at. Not ready to merge until that is completed.....

wdconinc:
> point to main of epic-data for gdml files

We explicitly do NOT want to point to an implicit moving target here. If there is a change needed to the data files, then this repository needs to be changed. Otherwise in a year from now, when epic-data's main branch is pointing to different files, we are not going to be able to make sense of the epic 24.02 release because we will have no idea what was in the epic-data main branch at this time.

johnlajoie:
> > point to main of epic-data for gdml files
> 
> We explicitly do NOT want to point to an implicit moving target here. If there is a change needed to the data files, then this repository needs to be changed. Otherwise in a year from now, when epic-data's main branch is pointing to different files, we are not going to be able to make sense of the epic 24.02 release because we will have no idea what was in the epic-data main branch at this time.

I understand your point. However, this is also a problem for the way the magnetic field map is referenced in the xml files, which is the example you pointed me to and what I used as an example in developing this code.  

Is the solution as simple as also tagging epic-data with every release?  This would allow you to insure a consistent set of files for future releases. You would still need to edit the xml files to pick up a specific release, however, which is not ideal.

johnlajoie:
> For the field maps we use an explicit hash: `https://github.com/eic/epic-data/raw/64b7ca6306b138b7f000e696c82bd8f72db1da56/MARCO_v.6.4.1.1.3_1.7T_Magnetic_Field_Map_2022_11_14_rad_coords_cm_T.txt` in marco.xml.

Ah - I did not fully appreciate the difference. I get it now - thanks, I'll update the xml file.

johnlajoie:
Completed in PR#588

veprbl:
\<off-topic\>
> Is the solution as simple as also tagging epic-data with every release? This would allow you to insure a consistent set of files for future releases. You would still need to edit the xml files to pick up a specific release, however, which is not ideal.

Tagging epic-data would be a good idea, but, maybe, we should stick with git sha1 hashes, still. This is primitive, but the proper solution to be worked on is to have a conditions database.
\</off-topic\>

wdconinc:
> I still don't understand the very strange z values for the hits

What I think is going on here is that the geometry is placed all right, but the DetElements that are reflecting the readout structure are not cloned correctly at the tower level. That results in all those tiles looking exactly the same, and the hits in each tile are just mapped onto the only DetElement that ends up getting defined.

Since thinking in cylinders is not that natural, imagine an endcap calorimeter (blocks stacked in a disk geometry). This would be equivalent as not propagating the y position in the stack correctly, and all blocks thinking they are at y = 0.

To debug, I'd print out the DetElements and their id to see if they make sense (and change `j` into `j_sector`, and `i` into `i_tower` for clarity).

wdconinc:
A few useful diagnostics on cellID could be:
```
events->Scan("HcalBarrelHits.cellID>>8&0x7ff") // prints tower; note that this doesn't work for Draw due to >>
events->SetAlias("HcalBarrelHits_tower", "HcalBarrelHits.cellID>>8&0x7ff")
events->Draw("HcalBarrelHits_tower")
events->Scan("HcalBarrelHits.cellID>>19&0x7") // print tile
```

wdconinc:
@johnlajoie Did b79c663 resolve the |z| < 2 cm issue?

johnlajoie:
Here is what I have when I plot the z position of the hits for the main branch of epic and the ohcalgeo_reconcile branch (in red).  The distributions look very similar, but the first two tiles in the center (surrounding the split in eta) seem to be missing).  Not sure what this means, maybe it does indicate a CellID problem?  That seems odd to me because the Cell ID's are uniformly distributed over what I would expect given the new hits format with only the tower referenced by a unique ID. 

compare

veprbl:
@johnlajoie This is a nice find. However, I believe it's unrelated. Probably, the small shifts in z come from rotations and shifts that you have to perform on tiles.

Returning back to the hit position issue. I'd like to illustrate it with this patch:
```diff
diff --git a/src/BarrelHCalCalorimeter_geo.cpp b/src/BarrelHCalCalorimeter_geo.cpp
index b8f0bee..8f8cf1d 100644
--- a/src/BarrelHCalCalorimeter_geo.cpp
+++ b/src/BarrelHCalCalorimeter_geo.cpp
@@ -17,6 +17,9 @@
 // Specialized generic detector constructor
 //
 //==========================================================================
+#define private public
+#include <TGeoTessellated.h>
+#undef private
 #include "DD4hep/DetFactoryHelper.h"
 #include "DD4hep/Printout.h"
 #include "XML/Layering.h"
@@ -255,6 +258,8 @@ static Ref_t create_detector(Detector& description, xml_h e, SensitiveDetector s
 
   Volume Tile[12];
   Volume ChimneyTile[4];
+  ROOT::Geom::Vertex_t tile_center[11];
+  ROOT::Geom::Vertex_t chimney_tile_center[4];
 
   for(int j=1; j<17; j++){
 
@@ -303,6 +308,16 @@ static Ref_t create_detector(Detector& description, xml_h e, SensitiveDetector s
     solidVolume.setVisAttributes(description, x_det.visStr());
     TessellatedSolid volume_solid = solidVolume.solid();
     volume_solid->CloseShape(true, true, true); // tesselated solid not closed by import!
+    int num_vertices = volume_solid->GetNvertices();
+    // Roughly estimate centre-of-mass for a tile in sPhenix coordinate system
+    ROOT::Geom::Vertex_t accumulator;
+    for (int ix = 0; ix < num_vertices; ix++) {
+      accumulator += volume_solid->GetVertex(ix);
+    }
+    ROOT::Geom::Vertex_t center = accumulator / num_vertices;
+    for (int ix = 0; ix < num_vertices; ix++) {
+      volume_solid->fVertices[ix] -= center;
+    }
     Material tile_material = description.material(tgdml_material.c_str());
     solidVolume.setMaterial(tile_material);
 
@@ -325,11 +340,13 @@ static Ref_t create_detector(Detector& description, xml_h e, SensitiveDetector s
         if (type == "OuterHCalTile") {
 
           Tile[11-tnum] = solidVolume;
+          tile_center[11-tnum] = center;
 
         }
         else if ((tnum > 7) && (type == "OuterHCalChimneyTile")) {
 
           ChimneyTile[11-tnum] = solidVolume;
+          chimney_tile_center[11-tnum] = center;
 
         }
 
@@ -372,7 +389,8 @@ static Ref_t create_detector(Detector& description, xml_h e, SensitiveDetector s
                                                    RotationY(180.0 * dd4hep::deg) *
                                                    Translation3D((-xposTileN[tnum] + (tnum + 1) * tile_tolerance - xposTileS[tnum]) * dd4hep::mm,
                                                                  (yposTileN[tnum] - yposTileS[tnum]) * dd4hep::mm,
-                                                                 (-zposTileN[tnum] - zposTileS[tnum]) * dd4hep::mm) );
+                                                                 (-zposTileN[tnum] - zposTileS[tnum]) * dd4hep::mm)
+                                                   * Translation3D(tile_center[i_eta].x(), tile_center[i_eta].y(), tile_center[i_eta].z()) );
 
         phv1.addPhysVolID("tile", i_phi + (12 + tnum) * 320);
         DetElement sd1 = tile_det.clone(_toString(i_phi + (12 + tnum) * 320, "tile%d"));
@@ -385,7 +403,8 @@ static Ref_t create_detector(Detector& description, xml_h e, SensitiveDetector s
                                                    Transform3D(RotationY(90.0 * dd4hep::deg),
                                                                Translation3D(xposOuter[0] * dd4hep::mm, yposOuter[0] * dd4hep::mm, 0.0 * dd4hep::mm)) *
                                                    RotationX(-tilePlaneRotate * dd4hep::deg) *
-                                                   Translation3D(((tnum + 1) * tile_tolerance) * dd4hep::mm, 0.0 * dd4hep::mm, 0.0 * dd4hep::mm));
+                                                   Translation3D(((tnum + 1) * tile_tolerance) * dd4hep::mm, 0.0 * dd4hep::mm, 0.0 * dd4hep::mm)
+                                                   * Translation3D(tile_center[i_eta].x(), tile_center[i_eta].y(), tile_center[i_eta].z()) );
 
         phv0.addPhysVolID("tile", i_phi + i_eta * 320);
         DetElement sd0 = tile_det.clone(_toString(i_phi + i_eta * 320, "tile%d"));
@@ -414,7 +433,8 @@ static Ref_t create_detector(Detector& description, xml_h e, SensitiveDetector s
                                                      RotationY(180.0 * dd4hep::deg) *
                                                      Translation3D((-xposTileN[tnum] - (tnum + 1) * tile_tolerance - xposTileS[tnum]) * dd4hep::mm,
                                                                    (yposTileN[tnum] - yposTileS[tnum]) * dd4hep::mm,
-                                                                   (-zposTileN[tnum] - zposTileS[tnum])* dd4hep::mm) );
+                                                                   (-zposTileN[tnum] - zposTileS[tnum])* dd4hep::mm)
+                                                   * Translation3D(tile_center[i_eta].x(), tile_center[i_eta].y(), tile_center[i_eta].z()) );
 
           phv1.addPhysVolID("tile", i_phi + i_eta * 320);
           DetElement sd1 = tile_det.clone(_toString(i_phi + i_eta * 320, "tile%d"));
@@ -427,7 +447,8 @@ static Ref_t create_detector(Detector& description, xml_h e, SensitiveDetector s
                                                      Transform3D(RotationY(90.0 * dd4hep::deg),
                                                                  Translation3D(xposOuter[0] * dd4hep::mm, yposOuter[0] * dd4hep::mm, 0.0 * dd4hep::mm)) *
                                                      RotationX(-tilePlaneRotate * dd4hep::deg) *
-                                                     Translation3D((-(tnum + 1) * tile_tolerance) * dd4hep::mm, 0.0 * dd4hep::mm, 0.0 * dd4hep::mm));
+                                                     Translation3D((-(tnum + 1) * tile_tolerance) * dd4hep::mm, 0.0 * dd4hep::mm, 0.0 * dd4hep::mm)
+                                                   * Translation3D(tile_center[i_eta].x(), tile_center[i_eta].y(), tile_center[i_eta].z()) );
           phv0.addPhysVolID("tile", i_phi + (12 + tnum) * 320);
           DetElement sd0 = tile_det.clone(_toString(i_phi + (12 + tnum) * 320, "tile%d"));
           sd0.setPlacement(phv0);
@@ -444,7 +465,8 @@ static Ref_t create_detector(Detector& description, xml_h e, SensitiveDetector s
                                                      Transform3D(RotationY(90.0 * dd4hep::deg),
                                                                  Translation3D(xposOuter[0] * dd4hep::mm, yposOuter[0] * dd4hep::mm, 0.0 * dd4hep::mm)) *
                                                      RotationX(-tilePlaneRotate * dd4hep::deg) *
-                                                     Translation3D(((tnum + 1) * tile_tolerance) * dd4hep::mm, 0.0 * dd4hep::mm, 0.0 * dd4hep::mm));
+                                                     Translation3D(((tnum + 1) * tile_tolerance) * dd4hep::mm, 0.0 * dd4hep::mm, 0.0 * dd4hep::mm) *
+                                                     Translation3D(chimney_tile_center[i_eta].x(), chimney_tile_center[i_eta].y(), chimney_tile_center[i_eta].z()) );
 
           phv1.addPhysVolID("tile", i_phi + i_eta * 320);
           DetElement sd1 = tile_det.clone(_toString(i_phi + i_eta * 320, "tile%d"));
@@ -457,7 +479,8 @@ static Ref_t create_detector(Detector& description, xml_h e, SensitiveDetector s
                                                      Transform3D(RotationY(90.0 * dd4hep::deg),
                                                                  Translation3D(xposOuter[0] * dd4hep::mm, yposOuter[0] * dd4hep::mm, 0.0 * dd4hep::mm)) *
                                                      RotationX(-tilePlaneRotate * dd4hep::deg) *
-                                                     Translation3D((-(tnum + 1) * tile_tolerance) * dd4hep::mm, 0.0 * dd4hep::mm, 0.0 * dd4hep::mm));
+                                                     Translation3D((-(tnum + 1) * tile_tolerance) * dd4hep::mm, 0.0 * dd4hep::mm, 0.0 * dd4hep::mm)
+                                                   * Translation3D(tile_center[i_eta].x(), tile_center[i_eta].y(), tile_center[i_eta].z()) );
           phv0.addPhysVolID("tile", i_phi + (12 + tnum) * 320);
           DetElement sd0 = tile_det.clone(_toString(i_phi + (12 + tnum) * 320, "tile%d"));
           sd0.setPlacement(phv0);
```
I hope this clarifies the issue for you. Unfortunately, modifying immutable tesselated solid in memory is not something we can accept. Probably need to wrap tiles in some sort of assembly, or re-export GDML with meshes centered at the origin.

johnlajoie:
@veprbl  - To be clear, what you are objecting to is storing the GDML tile locations in the xml file and instead suggest collecting them from the solids themselves.  That's a nice solution, and I'm happy to implement it. 

However, I don't see that this returns to the hit position issue at all, unless you are telling me that your patch solves it? The problem is still what you pointed out before - the scale of the z-hits range is -3 to 3 when it should be -3000 to 3000 - note that in the plot I attached  *I multiplied the z-hit in the new code by 1000* to compare it with main.  What this demonstrates is that even in the reduced scale with the new code the hit locations are quantized to the 24 tile positions  just like they are in main. This seems to indicate to me that something is wrong with the *scale* of the z-hits, it's not just wrong in some crazy way because *some information about the tile locations are preserved*.

At an even deeper level, I don't understand why the z-hits are quantized in main either, unless the hit location that is stored is always the center of the sensitive object?

veprbl:
@johnlajoie Yes, the patch works around the issue.
Without the patch:
before
With:
after

My assumption was that the tile offsets in GDML come from sPhenix CAD design. I thought, you eventually could migrate to using ePIC CAD placements, but that requires for design to mature, which, I understand, might not be right now.

I understand your argument about seeing a similar comb after rescaling current positions by factor 1000. You've concluded that the positions of the volumes are correct, but somehow z positions are scaled improperly. I disagree with that idea. My reason is that the positions of the volumes are currently based on values of `{x,y,z}posTile{N,S}` in the XML, and those Z offsets are mostly 0. What remains is an offset by `{x,y}posOuter`, which happens in rotated coordinates, so it probably creates offsets in z that you see.

It's a bit unfortunate that this takes so much of your time. The PR is in good shape and provides a major improvement, the z position was a minor one, it only made sense to rush to debug it in case there is some underlying issue with alignment that you'd discover along the way. Let's stabilize this ASAP together with the EICrecon change, document the TODOs and have it merged.

johnlajoie:
@veprbl many thanks for your help!  I shouldn't complain about the work, you guys work harder than me, 

I think I understand the problem better now - I'll get this cleaned up and submit a PR.

johnlajoie:
Fixed.  GDML files were updated to generate the solids around their center-of-mass, which fixes the problem with the z location of the hits. There was also some cleanup to the tile GDML files as well as some cleanup of the placement code. 
Capybara HCAL
This should finally be ready to merge.  I'll ask @ruse-traveler to double-check the EICRecon changes as well, but nothing there should have changed.

veprbl:
ddsim single pi- on main
```
        User time (seconds): 225.57
        System time (seconds): 0.37
        Percent of CPU this job got: 99%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 3:46.00
```
ddsim single pi- on ohcalgeo_reconcile
```
        User time (seconds): 313.48
        System time (seconds): 0.53
        Percent of CPU this job got: 98%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 5:18.83
```

### Capybara reports
* ddsim https://veprbl.github.io/capybara-reports/89d61a9720240c5b8318bc0e53a91e61/
* EICrecon https://veprbl.github.io/capybara-reports/aa375cd930cd3e30b6e7f2a27a252e23/

I'm not a fan of the new cell IDs, as you removed ability to interpret anything with readout fields, but okay.

veprbl (approved):
LGTM. Glad we've streamlined the gdml export a bit.

johnlajoie:
Hi @veprbl - the only relevant readout element is a tile, we removed the "tower" since it complicated the adjacency matrix for the clustering horribly and just went straight to a tile number.  But I see what you are saying - we could change the readout to a phi and eta index, that would make the values easy to interpret and be very easy for the adjacency matrix. It's easy to do. 

We've come this far, if it takes a few more tweaks to get it right let's do it now rather than come back to it again.

wdconinc:
> rather than come back to it again

"I'm never touching this code again!"

veprbl:
Yes, having an eta and phi indices would be preferable. My thinking is that the complexity in adjacency matrix is not as important, only few people have to look at it, whereas self documenting readout fields are a common good.

johnlajoie:
Let me see what I can do.  This is easy assuming you can associate two properties to a given cell ("ieta" and "iphi" as opposed to just "tile"), which I am assuming is possible? I'll look into the DD4HEP manual.

johnlajoie:
I now remember why we were unable to do this.  A given placed volume needs a unique detector element index. For the chimney sector we have four eta rings where the chimney tiles are different on one end of the detector compared to the other. This means that there will be two different volumes (shapes of tiles) with the same "iphi" index.  As I understand it this is forbidden, so the way around this was to encode ieta and iphi into a single "tile" variable which is unique for each tile.  This is straightforward to decode in the adjacency matrix (already done by @ruse-traveler ) 

So unless there is a better understanding (mine may be flawed) I think what we have now is the best compromise we can make and I would propose to stick with it, unchanged.

veprbl:
Hi @johnlajoie, I'm don't think what we want is forbidden. It is true that one needs to use unique touchable id's (id argument that goes into `placeVolume`), but the requirement is not imposed on readout id's (arguments to `addPhysVolID`). Moreover, if what you say is correct, it would have been impossible to make a detector where two volumes are read out with the same cellID.

johnlajoie:
Can I add more than one PhysVolID?  That would make it trivial to add the two indices for each tile

johnlajoie:
I believe I have it, each tile is now indexed in the readout by ieta, iphi.  Learned something new today....please check my work!

Capbara_ieta_iphi

@ruse-traveler - this will require another update to the EICRecon changes for the adjacency matrix (sorry).

johnlajoie:
Your wish is my command....

johnlajoie:
@ruse-traveler confirms that EICrecon changes produce compatible cluster spectrum. Looks like this PR and the companion EICrecon PR is ready to merge? 

image

veprbl (dismissed):

veprbl (approved):

Issue #587, opened 2023-10-25
inner_detector.yml file uses brycecanyon instead of craterlake


AkshayaVijay:
### Environment: (where does this bug occur, have you tried other environments)
- Which branch (`main` for latest released): 
- Which revision (`HEAD` for the most recent):

### Steps to reproduce: (give a step by step account of how to trigger the bug)
1. https://github.com/eic/epic/blob/main/configurations/inner_detector.yml

### Expected Result: (what do you expect when you execute the steps above)

this should include definitions_craterlake instead of definitions for brycecanyon

Chao1009:
temporarily fixed in #591

PR #586, opened 2023-10-24
update parameters for DIRC 

niwgit:
### Briefly, what does this PR introduce?
This implements following parameters for DIRC.
- Inner radius of 75.5 cm and outer radius of 78.6 cm for the DIRC bar box. The center of bar is at the average of those two values (r = 77.05 cm).
dirc_r

- DIRC readout starts at z = -303 cm. 
- Bar box ends at z = 185 cm. DIRC bar ends at z = 183 cm.
dirc_z

### What kind of change does this PR introduce?
- [x] Bug fix (issue #568)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [ ] Changes have been communicated to collaborators

PR #591, opened 2023-11-07
Update configuration for inner detectors

Chao1009:
Update the configuration so it does not report false overlaps in #586 

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### DIFF
--- a/configurations/inner_detector.yml
+++ b/configurations/inner_detector.yml
@@ -5,9 +5,11 @@ features:
     vertex_barrel:
     silicon_barrel:
     mpgd_barrel:
-    mpgd_dirc:
+    support_service_craterlake:
+    mpgd_outerbarrel:
+    mpgd_forward_endcap:
+    mpgd_backward_endcap:
     silicon_disks:
-    support_service_assembly:
     tof_barrel:
     tof_endcap:
   ecal:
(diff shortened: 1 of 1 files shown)


wdconinc (dismissed):

Chao1009:
```dd4hep: Evaluator::Object : unknown variable : TrackerSupportCyl_rmin1 : value=TrackerSupportCyl_rmin1```
It looks like mismatched ```definitions.xml```
I'm starting to think we need more sophisticated configurations, e.g., some JSON files.

wdconinc:
No, let's not make this any more complicated than it already is. We should remove the older geometries.

wdconinc (approved):
This is fine with me, but I wonder if we cannot just make sure that definitions.xml contains the craterlake definitions. That's the *de facto* standard configuration that everyone should be running in (and I'm tired of typing `export DETECTOR_CONFIG=epic_craterlake`). I kinda like to keep inner_detector.xml to be associate with what is in epic.xml, and would think inner_detector_craterlake.xml is assocated with epic_craterlake.xml. Just a thought.

Chao1009:
It resolves #587

PR #639, opened 2024-02-20
[DONTMERGE] definitions.xml: set EcalEndcapP_rmax to 172 cm

veprbl:
Allegedly, the calorimeter in simulation is larger than in engineering. This number was picked during a meeting from Alexander Bazilevski (of the top of his head).

(172 doesn't divide by 2.5, also it is not supported by the current menagerie https://eic.jlab.org/Geometry/Detector/Detector-20240117135224.html where it's still 195)


wdconinc:
Based on a recent ref, https://indico.bnl.gov/event/22766/, it seems like EcalEndcapP_rmax should stay at 195 cm.

PR #656, opened 2024-03-10
HOTFIX: add back HcalBarrel_rmin

veprbl:
I believe, the #588 breaks the CalorimeterTrackProjections by removing `HcalBarrel_rmin` which is referenced in
https://github.com/eic/EICrecon/blob/fcd8c84b14a9f0e5e486329bd9d7b42938ea4f46/src/global/tracking/tracking.cc#L164

I think, the `rmin1`/`rmin2` naming is a bit confusing, would be nice to know what the `10.4*cm` offset stands for, and get some better names.


wdconinc (approved):
Good catch!

wdconinc:
Backport into 24.03 needed I think.

epic-capybara:
Successfully created backport PR for `24.03`:
- #657

PR #661, opened 2024-03-11
fix: update tracker support and services cones per project CAD info

wdconinc:
### Briefly, what does this PR introduce?
This PR adds the outer support/services cone and places the cones and barrels in their actual intended positions. I'm marking this as a bugfix since the placement was completely wrong (with tracks in some regions going through two support cylinders...).

TODO:
- [x] #663
- [x] #673
  - this is not really a dependency, but if the radius of that system increases, we want to figure it out before putting in the services
- [x] #668 
- [x] #683
- [x] #684

### What kind of change does this PR introduce?
- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [x] Changes have been communicated to collaborators @sly2j 

### Does this PR introduce breaking changes? What changes might users need to make to their code?
No.

### Does this PR change default behavior?
Yes, it adds the project's estimate of the equivalent aluminum thickness of the services.

PR #663, opened 2024-03-12
fix: InnerMPGDBarrel rmin 51.25 -> 55 cm; zmax 135 -> 143 cm

wdconinc:
This PR updates the InnerMPGDBarrel position in accordance with latest drawings from the project engineers.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/compact/tracking/definitions_craterlake.xml
+++ b/compact/tracking/definitions_craterlake.xml
@@ -32,10 +32,10 @@
     <constant name="MPGDOuterBarrelModule_zmin2"     value="174.5*cm"/>
 
     <comment> Inner MPGD unsegmented barrel </comment>
-    <constant name="InnerMPGDBarrel_rmin"            value="51.25*cm"/>
+    <constant name="InnerMPGDBarrel_rmin"            value="55*cm"/>
     <constant name="InnerMPGDBarrel_dz"              value="0"/>
     <constant name="InnerMPGDBarrel_zmin"            value="105*cm"/> <comment> negative z </comment>
-    <constant name="InnerMPGDBarrel_zmax"            value="135*cm"/>    <comment> positive z </comment>
+    <constant name="InnerMPGDBarrel_zmax"            value="143*cm"/> <comment> positive z </comment>
     <constant name="InnerMPGDBarrel_length"          value="InnerMPGDBarrel_zmax + InnerMPGDBarrel_zmin"/>
     <constant name="InnerMPGDBarrel_zoffset"         value="0.5*(InnerMPGDBarrel_zmax - InnerMPGDBarrel_zmin)"/>
 


wdconinc:
References:
- https://indico.bnl.gov/event/22417/, ePIC detector integration, an update, Roland Wimmer, page 4

image

mposik1983:
New MPGD location:
Screen Shot 2024-03-13 at 2 42 14 PM
Screen Shot 2024-03-13 at 2 41 42 PM

radial envelope now extends from 55cm to 60cm

mposik1983:
New envelope has some overlaps with some Tracker support materials
Are these being addressed here or in a separate PR?

Screen Shot 2024-03-13 at 3 04 09 PM

mposik1983 (dismissed):

wdconinc:
> Are these being addressed here or in a separate PR?

You added the overlaps with your changes :-) Now we can't merge this.

compact/tracking/mpgd_barrel.xml
```diff
@@ -20,14 +20,9 @@
...
-    <constant name="InnerMPGDBarrelMod_rmin"                value="InnerMPGDBarrel_rmin"/>
-    <constant name="InnerMPGDBarrelMod_dz"                  value="InnerMPGDBarrel_dz"/>
-
-    <constant name="InnerMPGDBarrelMod_length"              value="InnerMPGDBarrel_length" />
-
-    <constant name="InnerMPGDBarrelLayer_length"            value="InnerMPGDBarrelMod_length + 1*um" />
-    <constant name="InnerMPGDBarrelLayer_thickness"         value="2*cm" />
-    <constant name="InnerMPGDBarrelLayer_rmin"              value="InnerMPGDBarrelMod_rmin - InnerMPGDBarrelLayer_thickness / 2.0"/>
+    <constant name="InnerMPGDBarrelLayer_length"            value="InnerMPGDBarrel_length" />
+    <constant name="InnerMPGDBarrelLayer_thickness"         value="5*cm" />
```
wdconinc, line 24: Probably this is the cause.

compact/tracking/mpgd_barrel.xml
```diff
@@ -63,26 +58,22 @@
       type="epic_TrackerBarrel"
       readout="MPGDBarrelHits"
       insideTrackingVolume="true">
-      <type_flags type="DetType_TRACKER + DetType_BARREL"/>
-      <dimensions
-        rmin="InnerMPGDBarrelLayer_rmin"
-        rmax="InnerMPGDBarrelLayer_rmax"
-        length="InnerMPGDBarrelLayer_length"/>
```
wdconinc, line 70: Maybe this too? Where does the MPGD get its rmin and rmax from now?

mposik1983: Hmmm.. I thought it was the due to the MPGD running into the cable services, since the MPGD envelope has increased in radius and length, unless the cable services were updated when the CF support was added

mposik1983: > Maybe this too? Where does the MPGD get its rmin and rmax from now?

This version of the MPGD is made up of long rectangular staves that are used to approximate a cylinder, whose dimensions are assigned in the module tag(https://github.com/eic/epic/blob/material_binning/compact/tracking/mpgd_barrel.xml#L71).

 rmin and rmax are used to define in the barrel envelop: https://github.com/eic/epic/blob/material_binning/compact/tracking/mpgd_barrel.xml#L89 that is used to place the staves around

wdconinc:
The overlap here is still due to the completely wrong model of services we're using now...
image

mposik1983:
> The overlap here is still due to the completely wrong model of services we're using now... 

yes, these cable services are where I thought the overlap was coming from. Should we also modify their radii here as well?

wdconinc:
> Should we also modify their radii here as well?

I'm just moving the services slightly inward, testing that overlaps are gone, and then we will override that anyway with #661 .

wdconinc:
Moved the outer supports/services cylinder inward by 3cm, now 1cm inward from InnerMPGDBarrel (was defined at InnerMPGDBarrel_rmin + 2 cm, which is indeed not good).

mposik1983 (dismissed):

mposik1983 (approved):

PR #667, opened 2024-03-15
feat: export simulation geometry to CAD models (STEP)

wdconinc:
This PR exports the geometry to STEP files for CAD import and dimensioning. There are no real settings to tweak, but to increase the level at which volumes are exported, hints can be added to the yaml file. Detectors can also be removed (currently only the ACTS subassemblies are removed since they don't do anything).

There are 'issues' that need to be addressed, but they have to be addressed in npsim:
- units are off (npdet_to_step use the cm value from TGeo and writes them as inch values in STEP) https://github.com/eic/npsim/issues/20,
- some shapes cannot be exported; not sure yet which ones and it makes more sense to bypass them in npsim https://github.com/eic/npsim/issues/21.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


wdconinc:
This now works "well enough" for merging: imported y-up and created a back view with left view section drawing.
image
(source: https://cad.onshape.com/documents/9fb38cc05081c441e152b5e1/w/441fcb286b98a55b7de89480/e/fe210e22e03730344768d9d6?renderMode=0&uiState=65f905d605a2a05c1bde1005, which is import of artifact at https://github.com/eic/epic/actions/runs/8336052681)

As I wrote, caveats:
- cm/mm confusion: the STEP file needs scaling by factor 10,
- barrel hcal is bypassed due lack of support for TGeoTessellated,
- too complicated for freecad? fails to load...
- requires "fast" geometry, don't attempt on local full geometry with fibers and all.

wdconinc:
Now with the tracking detectors (except not yet MPGDOuterBarrel), and as a sliced section view:
image

wdconinc:
Now with MPGD properly resolved:
image

.github/workflows/linux-eic-shell.yml
```diff
@@ -249,6 +249,42 @@ jobs:
         path: "*.root"
         if-no-files-found: error
 
+  convert-to-step:
+    runs-on: ubuntu-latest
+    needs:
+      - build
+    strategy:
+      matrix:
+        detector_config: [epic_craterlake_no_bhcal]
```
veprbl, line 258: I think, it would be worthwhile to generate subsets of detectors to see which ones cause issues when loading with FOSS tools like FreeCAD
```suggestion
        detector_config: [epic_craterlake_no_bhcal, epic_craterlake_tracking_only, epic_dirc_only, epic_drich_only, epic_imaging_only, epic_ip6, epic_lfhcal_with_insert]
```

wdconinc: That was the original idea (and implementation). However, this causes poor scaling since each export (even of a subset of detectors) requires the entire geometry to be converted to OCCT internal representation, which is the time-consuming part and the part where failures to convert are apparent. Once it is in OCCT internal representation it always succeeds in converting to STEP (even if not a closed solid as in the case of rmin=0 cylinders), so not much is gained anymore in exporting only part of the geometry. In fact, something is lost since subsets are exported as parts and assemblies with their own origin that is not the interaction point origin, so it's not so useful.

What we can do is export all (or at least more of) the configurations. The export of fast geometries is fast enough to allow this. In many cases (all cases with the central beam line) we would maintain the origin at the interaction point.

veprbl: Sure, I understand, you can toggle the unwanted parts in a more powerful software. If we want to pursue compatibility with FOSS software, it'd be helpful to be able to have those extra configurations available from CI. I'm not sure why the origin would be correct exported only from `epic.xml`, but not from `epic_somesingledetector_only.xml`, how does it even know.

wdconinc: Side question, but did you have trouble loading the step file in freecad? It's produced by OCCT and freecad uses OCCT and I had no issues loading with freecad.

wdconinc: As for origin, subassemblies in CAD are saved in their own coordinate system. The only subassemblies that have a coordinate system that agrees with the world coordinate system are this where we import a whole set of subsystems that includes the beam pipe. Exporting the EcalEndcapN from epic_craterlake puts it in the coordinate system of the ecal. Exporting the entire hypothetical epic_craterlake_ecal_endcap_n_only.xml would end up in the correct coordinate system.

wdconinc: I was going to go with
```yaml
    strategy:
      matrix: ${{ fromJson(needs.list-detector-configs.outputs.configs) }}
```
but that will fail because of the bhcal. So we'd have to amend the explicit list we have now, by hand, with only those configurations that don't have the bhcal.

veprbl: > Side question, but did you have trouble loading the step file in freecad? It's produced by OCCT and freecad uses OCCT and I had no issues loading with freecad.

It was taking 60+ GB of memory on a laptop with 16 Gb RAM. macOS performs quite well under memory pressure, but I didn't have enough patience to let it load.

wdconinc: Did you disable the fibers? This uses the ’fast' geometry because it is otherwise too large and doesn't load anywhere.

wdconinc:
This now exports also things like `epic_drich_only.stp`, but the drawback is that you don't get good info on where the interaction point is in these CAD models. This imports in FreeCAD for those who prefer that for CAD (I can't see how FreeCAD can be told to import models with Y as pointing up, instead of having Z as pointing up, so my rotations are off).
epic_drich_only

veprbl (approved):

PR #669, opened 2024-03-17
fix: tell dd4hep_configure_output which install prefix to use

wdconinc:
This PR prevents [dd4hep_configure_output](https://github.com/AIDASoft/DD4hep/blob/e52af7f770ec3e51460d832428f2cca214678e2a/cmake/DD4hepBuild.cmake#L328-L331) from overriding the `CMAKE_INSTALL_PREFIX` that we just set in case the user doesn't define it.

- [x] Bug fix (issue: setup.sh has path that does not match with actual install location)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

Yes. The default install prefix (if not specified) will change from `${CMAKE_SOURCE_DIR}` to `${CMAKE_SOURCE_DIR}/prefix`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,7 +46,7 @@ find_package(fmt REQUIRED)
 #-----------------------------------------------------------------------------------
 set(a_lib_name ${PROJECT_NAME})
 
-dd4hep_configure_output()
+dd4hep_configure_output(INSTALL ${CMAKE_INSTALL_PREFIX}))
 dd4hep_set_compiler_flags()
 
 file(GLOB sources CONFIGURE_DEPENDS src/*.cpp)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -46,7 +46,7 @@ find_package(fmt REQUIRED)
 #-----------------------------------------------------------------------------------
 set(a_lib_name ${PROJECT_NAME})
 
-dd4hep_configure_output()
+dd4hep_configure_output(INSTALL ${CMAKE_INSTALL_PREFIX}))
```
wdconinc, line 49:
```suggestion
dd4hep_configure_output(INSTALL ${CMAKE_INSTALL_PREFIX})
```

veprbl (approved):
Looks like we want this. I've checked in DD4hep, looks like there is no extra version requirement to use the `INSTALL` argument.

wdconinc:
We might have to announce this since I think we may get questions about where the detector geometry gets installed (or worse we will have people use what used to get installed in the source directory and assume it's reflecting the updated geometry).

Issue #54, opened 2022-08-12
Jinja2 expansion xml files should go to CMAKE_CURRENT_BINARY_DIR


wdconinc:
**Is your feature request related to a problem? Please describe.**
Users may `make` the epic code and end up with an `epic.xml` file in their source dir. They may do something with this file, which refers to environment variables. That may lead to confusing results.

**Describe the solution you'd like**
With produced xml files in the build directory, it is less likely that they would be used inadvertently assuming they work from the source dir top level.

**Describe alternatives you've considered**
Alternative is not changing.

veprbl:
All build products should always go to the build directory. That helps to prevent them from being committed to the repo.

wdconinc:
Totally agree! This was backwards compatibility when we switched from one ecce.xml in the repository to a template with build products. The benchmarks are still using a crazy installation scheme where we are running against a non-installed version in its source directory.

So, yes, and for every rule there's an exception that may or may not have been a good idea at the time ;-)

wdconinc:
Depends on: https://eicweb.phy.anl.gov/EIC/benchmarks/common_bench/-/issues/5

wdconinc:
Fixing that issue required having a way to refer to ip6 without using a symlink. You see, it's all related :-)

c-dilks:
I think this is good idea. Would it have any impact on the typical workflow of changing something in `compact/some_detector.xml`, and then calling `ddsim` (i.e., without running `cmake`)? I suppose if you set `$DETECTOR_PATH` to the directory that has your `compact/some_detector.xml`, then it would be fine.

wdconinc:
Yes, I'd suggest that workflow would be easiest resolved by setting DETECTOR_PATH to a custom location.

PR #672, opened 2024-03-18
fix: mv evaluated templates during build into CMAKE_CURRENT_BINARY_DIR

wdconinc:
The build system writes the evaluated jinja templates (entrypoint xml files) into the source directory instead of a build directory. This PR moves these temporary build artifacts into the build directory instead.

TODO:
- [x] fix common_bench to do a proper install into a prefix instead of a build into the source directory

- [x] Bug fix (issue #54)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

Yes, no more xml files will be written to the source directory; `make install` or bust.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,9 +60,6 @@ target_link_libraries(${a_lib_name}
 
 #-----------------------------------------------------------------------------------
 # Parse jinja templates: once by default, and once for all yml files
-#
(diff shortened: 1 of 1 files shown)


wdconinc:
It seems common_bench is already fixed, https://eicweb.phy.anl.gov/EIC/benchmarks/common_bench/-/commit/39c14811787fdaa5645203817378f292da9f4664

veprbl (approved):

PR #674, opened 2024-03-21
Remove brycecanyon configurations

veprbl:

Removes epic_brycecanyon and upgrades other configurations.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

Not expected

No


simonge:
ip6_arches.yml and ip6_brycecanyon.yml should probably go too as there is no difference in the detector configuration anyway and the beam energy would be better described by a beam energy related suffix.

veprbl:
> ip6_arches.yml and ip6_brycecanyon.yml should probably go too as there is no difference in the detector configuration anyway and the beam energy would be better described by a beam energy related suffix.

Implemented. Thanks!

wdconinc (approved):

wdconinc:
I'm queueing this since #679 depends on this.

PR #675, opened 2024-03-21
test: npsim-dis --enableDetailedShowerMode

wdconinc:
This PR adds the `--enableDetailedShowerMode` to the npsim-dis job to check where the slow late calorimeter sim hits are coming from.

File size of `sim_dis_5x41_minQ2=1_epic_craterlake.edm4hep.root` from 53.7 MB to 89.9 MB.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [x] Other: test

- [x] Changes have been communicated to collaborators @kkauder @jml985

No.

Yes, may be slower.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -662,7 +662,7 @@ jobs:
         setup: install/setup.sh
         run: |
           url=root://dtn-eic.jlab.org//work/eic2/EPIC/EVGEN/DIS/NC/${{matrix.beam}}/minQ2=${{matrix.minq2}}/pythia8NCDIS_${{matrix.beam}}_minQ2=${{matrix.minq2}}_beamEffects_xAngle=-0.025_hiDiv_1.hepmc3.tree.root
-          npsim --compactFile ${DETECTOR_PATH}/${{ matrix.detector_config }}.xml -N 100 --inputFiles ${url} --random.seed 1 --outputFile sim_dis_${{matrix.beam}}_minQ2=${{matrix.minq2}}_${{ matrix.detector_config }}.edm4hep.root -v WARNING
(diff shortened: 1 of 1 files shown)


wdconinc:
`EcalBarrelScFiHitsContributions.time` for all (black), electrons (red), neutrons (neutrons): (FYI @jml985)
EcalBarrelScFiHitsContributions_time_per_PDG

wdconinc:
Ref: https://github.com/eic/epic/actions/runs/8379877674/job/22947919211?pr=675#step:5:362:
```
### Adding tracking cuts for neutron  TimeCut(ns)= 10000  KinEnergyCut(MeV)= 0
```

kkauder:
Excellent test. Also, the file size increase probably doesn't justify enabling this by default.

wdconinc:
> file size increase probably doesn't justify enabling this by default

It is not of much value beyond diagnostics, indeed.

wdconinc:
@kkauder Can you think about the possible ways in which digitization would use the detailed shower mode hit contributions?

Without it we have access to energy and time of all contributions (so this allows waveform determination and filtering out late hits).

With it we have access to pdg and step position as well. That would be required if we want to change the sensitivity to specific particles beyond the differences in energy deposition that geant4 simulates already, or if we have some position dependent efficiency.

About the only use case I can think of is using the position to modify the time in the fibers without needing to use z segmentation as we are doing now.

wdconinc:
@kkauder Any idea about ways in which digitization would use the detailed shower mode hit contributions? If not, then I'll close this and the npsim PR.

(cc @veprbl)

veprbl:
My recollection was that detailed showers produced huge files. I don't think we need it for regular reconstruction, if the detailed time information is available.

wdconinc:
Closing since we won't implement detailed shower mode (not at this point).

PR #679, opened 2024-03-23
feat: improved flexibility to define far forward and far backward components

wdconinc:
Currently, we cannot easily control what is included in the far forward and far backward regions, except by hacking around it like in `configurations/zdc_sipm_on_tile_only.yml`. This is making it less obvious how to add only the required far forward geometry to the Acts material map geometry.

This PR modifies the jinja template to allow more fine-grained far forward and far backward control, similar to the other subsystem types (e.g. ecal, hcal, pid). It maintains the default, so just specifying `features:far_forward:default:` will load the former `far_forward.xml`, but `features:far_forward:[ZDC_Crystal_LYSO:, ZDC_SiPMonTile:]` will load only those two subsystems.

TODO:
- [x] #674 

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators @ShujieL 

No.

No.


templates/epic.xml.jinja2
```diff
@@ -181,18 +181,26 @@
   <documentation level="11">
     ## Far foward detectors
   </documentation>
-  {% for ff in features['far_forward'] -%}
-  <include ref="${DETECTOR_PATH}/compact/{{ ff }}.xml"/>
-  {% endfor -%}
+  {% if features['far_forward'] | length > 1 -%}
```
veprbl, line 184: I don't like the non-uniform approach. Better, just move `compact/far_forward.xml` to `compact/far_forward/full.xml` and keep the explicit `full:` in the configuration yamls.

wdconinc: Yeah, jinja doesn't like it either. `full.xml` in the categories is a good idea.

wdconinc: I'm going to go with `default` instead of `full` since in some cases (far backward here), the default is to have fewer elements than some other 'extended' implementations.

wdconinc: Didn't change the brycecanyon ones removed in #674, but otherwise ready for review again.

veprbl (approved):

PR #681, opened 2024-03-23
feat: allow `source setup.sh epic_craterlake` to set DETECTOR_CONFIG

wdconinc:
This PR adds a quality of life improvement, hopefully, for loading geometry and setting `DETECTOR_CONFIG` in one go.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


veprbl (approved):
I'm not a user of this, but the shell syntax looks correct.

wdconinc:
> I'm not a user of this, but the shell syntax looks correct.

All I was thinking is you'd point out it's a bashism or so ;-)

veprbl:
> > I'm not a user of this, but the shell syntax looks correct.
> 
> All I was thinking is you'd point out it's a bashism or so ;-)

It's a bashism. I assume, in the worst case it will print warnings when running in incompatible shells, and DETECTOR_CONFIG will remain unset.

PR #683, opened 2024-03-24
fix: rotate DRICH by 30 degrees

wdconinc:
This PR rotates the DRICH by 30 degrees. Per Marco's slides at https://indico.bnl.gov/event/22417/, e.g. slides 9 and 10, the DRICH has a readout box at the top, bottom, and at 2 and 4 o'clock, and 8 and 10 o'clock. That's different from where we had it before which was 0 degrees in phi, i.e. X axis, which is 9 o'clock. For an interaction region with a crossing angle, this matters, of course.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

Yes, it rotates the DRICH by 30 degrees.

### DIFF
--- a/src/DRICH_geo.cpp
+++ b/src/DRICH_geo.cpp
@@ -346,7 +346,7 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
       continue;
 
     // sector rotation about z axis
-    RotationZ   sectorRotation(isec * 2 * M_PI / nSectors);
+    RotationZ   sectorRotation((isec + 0.5) * 2 * M_PI / nSectors);
     std::string secName = "sec" + std::to_string(isec);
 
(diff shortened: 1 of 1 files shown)


chchatte92:
I guess this rotation changes the dRICH acceptance, right? Maybe I give a quick look to the acceptances. In one day I will have the figures. Does any other sub detector depend on this change?

wdconinc:
The supports and services don't fit with the dRICH in the wrong orientation.

wdconinc:
> Does any other sub detector depend on this change?

No, this is a change that only affects the dRICH (and the service routing).

chchatte92:
Looks reasonable.

chchatte92 (approved):
Looks good to me. I am approving this one and the following PR on the sensor box.

PR #684, opened 2024-03-25
feat: dRICH sensor boxes in more realistic sector segments

wdconinc:
Until now the dRICH sensor boxes were a 5cm larger outer radius cylinder sticking out of the gas volume. This did not leave room for service routing between the sectors of the dRICH.

This PR modifies the sensor boxes (and envelope volume) to be a cylinder segment of configurable delta phi (set to 48 degrees to fit the sensors). The sections between the segments are conical (polycone), as per both services routing cad and recent TIC presentation.

TODO:
- [x] #683 
- [x] polycone rmin at middle z-plane should be linerp of outer z-plane rmin values, not the current front z-plane rmin (may overlap with beam pipe but is incorrect in any case)

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

Yes, adds azimuthally smaller sensor boxes, with side walls that have locally higher X0 expected

### DIFF
--- a/compact/pid/drich.xml
+++ b/compact/pid/drich.xml
@@ -24,6 +24,7 @@
 <constant name="DRICH_sensorbox_length"   value="50.0*cm"/>  <!-- z-length of the extrusion -->
 <constant name="DRICH_sensorbox_rmin"     value="DRICH_rmax1 + 2*cm"/>  <!-- lower radial limit of the extrusion -->
 <constant name="DRICH_sensorbox_rmax"     value="DRICH_rmax2 + 5*cm"/>  <!-- upper radial limit of the extrusion -->
+<constant name="DRICH_sensorbox_dphi"     value="48*degree"/>  <!-- azimuthal width of the extrusion -->
 <!-- aerogel+filter geometry -->
 <constant name="DRICH_aerogel_thickness"  value="4.0*cm"/>  <!-- aerogel thickness -->
 <constant name="DRICH_airgap_thickness"   value="0.01*mm"/> <!-- air gap between aerogel and filter -->
--- a/src/DRICH_geo.cpp
+++ b/src/DRICH_geo.cpp
@@ -202,7 +202,7 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
 
   // tank solids:
   // - inner: cone along beamline
-  // - outer; cone to back of sensor box, then fixed radius cylinder
+  // - outer: cone to back of sensor box, then fixed radius cylinder
   Polycone vesselTank(0, 2 * M_PI,
            /* rmin */ {vesselSnout.rMin2(), std::lerp(vesselSnout.rMin2(), vesselRmin1, (sensorboxLength - snoutLength) / tankLength), vesselRmin1},
            /* rmax */ {vesselSnout.rMax2(), vesselRmax2, vesselRmax2},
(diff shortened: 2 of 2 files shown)


compact/pid/drich.xml
```diff
@@ -24,6 +24,7 @@
 <constant name="DRICH_sensorbox_length"   value="50.0*cm"/>  <!-- z-length of the extrusion -->
 <constant name="DRICH_sensorbox_rmin"     value="DRICH_rmax1 + 2*cm"/>  <!-- lower radial limit of the extrusion -->
 <constant name="DRICH_sensorbox_rmax"     value="DRICH_rmax2 + 5*cm"/>  <!-- upper radial limit of the extrusion -->
+<constant name="DRICH_sensorbox_dphi"     value="48*degree"/>  <!-- azimuthal width of the extrusion -->
```
wdconinc, line 27: This is the smallest delta phi that still contains the sensors. The project has this as 42 degrees, with the 18 degrees between sensor boxes for service routing. Out of scope for this PR but the sensors placement may need adjustment.

wdconinc:
Looks like this now.
Screenshot_20240324-230020

compact/pid/drich.xml
c-dilks: You can likely reduce `sphericalpatch` `phiw` to decrease the azimuthal extent that sensors are allowed; 30 deg is maximal, meaning _no_ azimuthal cut (letting its `zmin` limit the azimuthal extent)

https://github.com/eic/epic/blob/c3c0277afa1417a77050c15f3707d4789f1a259d/compact/pid/drich.xml#L331

wdconinc: I'm not going to change this within this PR, but @chchatte92 we do need to change this to a smaller value that fits with what the project has in terms of azimuthal width. I don't know what it should be to fit in the 42 degrees azimuthal width, but I'll let you figure that out. For now we can acommodate this by bundling the services up in thicker bundles over a smaller delta phi.

wdconinc:
No more overlaps. Ready for review.

src/DRICH_geo.cpp
```diff
@@ -199,23 +200,43 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
       vesselRmin0 + boreDelta * (snoutLength - windowThickness) / vesselLength + wallThickness,
       vesselRmax1 - wallThickness + windowThickness * (vesselRmax1 - vesselRmax0) / snoutLength);
 
-  // tank solids
-  Cone vesselTank(tankLength / 2.0, vesselSnout.rMin2(), vesselRmax2, vesselRmin1, vesselRmax2);
-  Cone gasvolTank(tankLength / 2.0 - windowThickness, gasvolSnout.rMin2(), vesselRmax2 - wallThickness,
-                  vesselRmin1 + wallThickness, vesselRmax2 - wallThickness);
+  // tank solids:
+  // - inner: cone along beamline
+  // - outer; cone to back of sensor box, then fixed radius cylinder
```
wdconinc, line 205:
```suggestion
  // - outer: cone to back of sensor box, then fixed radius cylinder
```

chchatte92 (approved):
Looks good.

wdconinc:
Note @chchatte92: this includes #683. I've removed it from the merge queue since you'd said you wanted to review the acceptance. If you're ok with #683, then feel free to approve/merge that one and (only then) this one.

chchatte92:
Sure @wdconinc . I will find time tomorrow for the acceptance checks.

Issue #693, opened 2024-04-03
Add barrel flux return backlegs, spacers


wdconinc:
**Is your feature request related to a problem? Please describe.**
Background radiation load calculations in the hall require the barrel flux return backlegs and spacers to be implemented in geometry.

**Describe the solution you'd like**
The CAD model is at https://brookhavenlab.sharepoint.com/sites/EICPublicSharingDocs/Shared%20Documents/Forms/AllItems.aspx?ga=1&id=%2Fsites%2FEICPublicSharingDocs%2FShared%20Documents%2FExperimental%20Program%2FePIC%2FEngineering%2FSTP-Files&viewid=991e5a0b-d053-4acd-8565-2cc07a440143. A relevant drawing is attached. Material is SAE AISI 1008 per https://www.sciencedirect.com/science/article/pii/S0168900202019617, for now assuming same for spacers.

[STAR Flux for EPIC Drawing 1.pdf](https://github.com/eic/epic/files/14857241/STAR.Flux.for.EPIC.Drawing.1.pdf)

![image](https://github.com/eic/epic/assets/4656391/7001f924-a59f-4043-b57f-20d8906d52a2)

PR #694, opened 2024-04-04
feat: add barrel flux return

wdconinc:
This PR adds the barrel flux return backlegs and spacers per the info in #693. Needed for hall background radiation studies. Also adds forward endcap flux return.

image

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators @ajentsch 

No.

Not really for any detector behavior.

### DIFF
--- a/compact/definitions.xml
+++ b/compact/definitions.xml
@@ -632,15 +632,16 @@ Service gaps in FW direction (before endcapP ECAL) and BW direction (before endc
 
     <constant name="LFHCAL_rmax"          value="HcalBarrel_rmax"/>
 
+    <comment> Lepton_Assy_21.stp, rwimmer, 2024-04-03 </comment>
     <constant name="FluxEndcapN_collar_rmax" value="326.2*cm"/>
-    <constant name="FluxEndcapN_collar_rmin" value="269.0*cm"/>
-    <constant name="FluxEndcapN_collar_thickness" value="120.4*cm"/>
-    <constant name="FluxEndcapN_oculus_rmax" value="267.0*cm"/>
-    <constant name="FluxEndcapN_oculus_rmin" value="213.6*cm"/>
+    <constant name="FluxEndcapN_collar_rmin" value="275.0*cm"/>
+    <constant name="FluxEndcapN_collar_thickness" value="120.38*cm"/>
+    <constant name="FluxEndcapN_oculus_rmax" value="FluxEndcapN_collar_rmin"/>
+    <constant name="FluxEndcapN_oculus_rmin" value="221.6*cm"/>
     <constant name="FluxEndcapN_oculus_thickness" value="28.5*cm"/>
-    <constant name="FluxEndcapN_rmax" value="267.0*cm"/>
-    <constant name="FluxEndcapN_rmin" value="14*cm"/>
-    <constant name="FluxEndcapN_thickness" value="10*cm"/>
+    <constant name="FluxEndcapN_exterior_rmax" value="FluxEndcapN_collar_rmin*cm"/>
+    <constant name="FluxEndcapN_exterior_rmin" value="16*cm"/>
+    <constant name="FluxEndcapN_exterior_thickness" value="10*cm"/>
 
     <comment> Hadron End Cap Assembly_North Half.stp, rwimmer, 2024-04-03 </comment>
     <constant name="FluxEndcapP_collar_rmax" value="326.2*cm"/>
--- a/compact/definitions.xml
+++ b/compact/definitions.xml
@@ -630,8 +630,6 @@ Service gaps in FW direction (before endcapP ECAL) and BW direction (before endc
 
     <constant name="HcalEndcapN_rmax"     value="min(HcalBarrel_rmax, 267.0 * cm)"/>
 
-    <constant name="LFHCAL_rmax"          value="HcalBarrel_rmax"/>
-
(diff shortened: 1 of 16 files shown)


wdconinc:
Currently overlaps with LFHCAL since:
```
LFHCAL_rmax                    =      270.030 = HcalBarrel_rmax // should be 267
HcalBarrel_rmax                =      270.030 = HcalBarrel_rmin1 + HcalBarrel_thickness // should be 268.2
HcalBarrel_rmin1               =      183.850 = 183.85*cm // should 180.0
HcalBarrel_rmin2               =      194.250 = HcalBarrel_rmin1 + 10.4*cm // correct
```

@johnlajoie @ruse-traveler This issue seems to start with HcalBarrel_rmin1. What should be changed here to end up with HcalBarrel_rmax = 268.2 cm?

johnlajoie:
As far as I am concerned, the 268.2 cm in the detector table is incorrect, and I went around and around with the project with this for months.  As a reminder, what is in the ePIC geometry for the outer HCAL is the 3D geometry as-built. Rolf explained to me that the detector parameter table is generated from a 2D drawing, and my best guess is that there is some confusion because what they are looking at is 2D and the plate is tilted at an angle. (This can't be the whole problem or it would be a larger discrepancy. It could also have to do with the fact that I think their 2D drawing has a plate and not the completed sector, which can be different, esp. with the end plates.)    Rolf insisted that all they can work from is the 2D drawing. 

In my last request for an update I provided them with a set of dimensions from the 3D model: 

https://eic.jlab.org/Geometry/Detector/local/H/HCAL_Geometry_Summary_from_3D.pdf

which is linked to the Detector Parameter table. I see now they updated some of the parameters in my request, but not others, without comment. 

I can revisit this again with Walt, Tanja and Rolf if necessary

wdconinc:
Ok, so the barrel hcal is fine, and it's just that the positive endcap flux return dimensions are not in agreement with the CAD model. I'll update the positive endcap flux return, and this should resolve the overlaps.

CAD
table

E.g. inner radius of outer collar highlighted: 269 cm in table, 289.56 cm in drawing.

wdconinc:
FluxEndcapN dimensions
CAD
table

wdconinc:
With 7fcecd0 the LFHCAL is now in an envelope volume that is a tube with off-center box removed. This envelop volume is made of Steel235 to account for the "staircase" pattern in https://github.com/eic/epic/pull/694#issuecomment-2037737708.

wdconinc:
Envelope volume of LFHCAL:
image

wdconinc:
This does not add any spurious extra material to the LFHCAL. The output itself is too long, but no changes in sha256 on a non-trivial ray from the origin.

Before:
```
$ materialScan /opt/detector/epic-nightly/share/epic/epic_lfhcal_only.xml 0 0 0 1 200 500 | sha256sum
cfeeaaaf182d53fc67fedc92ad51a28ee61b122b8d7d770d419c3df06b5c6fee  -
```

After:
```
$ materialScan /home/wdconinc/git/epic/.worktree/693-add-barrel-flux-return-backlegs-spacers/prefix/share/epic/epic_lfhcal_only.xml 0 0 0 1 200 500 | sha256sum
cfeeaaaf182d53fc67fedc92ad51a28ee61b122b8d7d770d419c3df06b5c6fee  -
```

compact/definitions.xml
```diff
@@ -630,17 +630,35 @@ Service gaps in FW direction (before endcapP ECAL) and BW direction (before endc
...
+    <constant name="FluxEndcapP_collar_rmin" value="289.56*cm"/>
+    <constant name="FluxEndcapP_collar_thickness" value="167.14*cm"/>
+    <constant name="FluxEndcapP_oculus_rmax" value="FluxEndcapP_collar_rmin"/>
+    <constant name="FluxEndcapP_oculus_rmin" value="210.0*cm"/>
+    <constant name="FluxEndcapP_oculus_thickness" value="22.2*cm"/>
+
+    <comment> LFHCAL includes support, which is part of the flux return </comment>
+    <constant name="LFHCAL_rmax" value="FluxEndcapP_collar_rmin"/>
+
+    <constant name="FluxBarrelForward_zmax"  value="320*cm"/>
```
wdconinc, line 655:
```suggestion
    <comment> STAR Asm for EPIC w cradle.stp, rwimmer, 2024-04-03 </comment>
    <constant name="FluxBarrelForward_zmax"  value="320*cm"/>
```

compact/hcal/lfhcal.xml
```diff
@@ -231,7 +236,7 @@
         <position x="40*cm" y="-35*cm" z="0*cm"/>
         <position x="-30*cm" y="-45*cm" z="0*cm"/>
         <position x="30*cm" y="-45*cm" z="0*cm"/>
-        <position x="0*cm" y="-55*cm" z="0*cm"/>
+        <position x="0* cm" y="-55*cm" z="0*cm"/>
```
wdconinc, line 239:
```suggestion
        <position x="0*cm" y="-55*cm" z="0*cm"/>
```

veprbl:
For some reason endcaps have 50% transparency in teveDisplay. Haven't been able to figure out why.

veprbl (approved):

PR #696, opened 2024-04-04
Added definitions_craterlake to ip6 configurations

simonge:
The ip6 configurations were unable to be used to reconstruct tracks from the B0 surfaces without the DD4hep_SubdetectorAssembly which wraps around it which is defined in definitions_craterlake.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No
No

### DIFF
--- a/configurations/ip6.yml
+++ b/configurations/ip6.yml
@@ -1,5 +1,7 @@
 features:
   beampipe:
+  tracking:
+    definitions_craterlake:
   far_forward:
     default:
   far_backward:
--- a/configurations/ip6_extended.yml
+++ b/configurations/ip6_extended.yml
@@ -1,5 +1,7 @@
 features:
   beampipe:
+  tracking:
+    definitions_craterlake:
   far_forward:
     default:
   far_backward:


wdconinc (approved):
Good catch. Can you add a test for this? I thought we already had check-tracking-geometry which was supposed to catch this. Or is this a situation where it just doesn't include B0 tracking, but runs otherwise ok?

simonge:
Doesn't look like it would have been caught by the test even if the configuration was being tested. The most telling message I saw while running EICrecon was:

[acts] [info] CONV           Subdetector with name : 'B0Tracker' has inconsistent information for translation and is not of type 'compound'. If you want to have this DetElement be translated into the tracking geometry you need add the right DetectorType or VariantParameters (at this stage the subvolume needs to be declared as BEAMPIPE or BARREl, or have a VariantParameter passive_layer=true) or if it is a compound DetElement (containing a barrel-endcap hierarchy), the type needs to be set to 'compound'.

For some reason I can't run test_ACTS locally to confirm the output looks the same. I'll open a new PR with some extended tests

PR #697, opened 2024-04-04
Extended configs tested by check-tracking-geometry

simonge:
Adds tests to check if more configurations are pass the check-tracking-geometry check.

Initially this is before the epic_ip6 has been updated so should fail the test but probably initially won't. This should be followed by a change to acts_geo_check to cover the case or a request to change the logger level of messages which should cause it to fail.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No
No

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -173,6 +173,9 @@ jobs:
   check-tracking-geometry:
     runs-on: ubuntu-latest
     needs: build
+    strategy:
+      matrix:
+        detector_config: [epic_craterlake, epic_ip6]
     steps:
     - uses: actions/checkout@v4
     - uses: actions/download-artifact@v4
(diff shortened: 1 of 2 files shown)


simonge:
Is it worth requesting the log level in acts gets upgraded to a warning?
https://github.com/acts-project/acts/blob/main/Plugins/DD4hep/src/ConvertDD4hepDetector.cpp#L532

Presumably we wouldn't adopt an updated version for a while anyway so worth adding another test in the meantime.

wdconinc:
> Is it worth requesting the log level in acts gets upgraded to a warning?

Yes, they are happy about PRs. It will take even longer for us to pick it up if we don't request it :-) But changing log level from info to warning doesn't really address the fact that it will still accept it.

wdconinc (approved):
Nice find and test. I think we can just check all geometries as we do in some other checks already.

PR #698, opened 2024-04-05
feat: install setup.sh as FHS compliant bin/thisepic.sh

wdconinc:
Installing `setup.sh` in the prefix is not consistent with the linux filesystem hierarchy standard and leads to `/usr/local/setup.sh`. This PR installs `setup.sh` into e.g. `/usr/local/bin/thisepic.sh`, which is FHS compliant.

Next steps:
- Transition to using FHS location
- Add deprecation warning to old version (v24.MM++)
- Remove old version (v24.MM++)

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

Not yet, but recommended default changes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -113,6 +113,10 @@ configure_file(templates/setup.sh.in ${CMAKE_CURRENT_BINARY_DIR}/setup.sh @ONLY)
 install(FILES ${CMAKE_CURRENT_BINARY_DIR}/setup.sh
     DESTINATION ${CMAKE_INSTALL_PREFIX}
 )
+install(FILES ${CMAKE_CURRENT_BINARY_DIR}/setup.sh
+    DESTINATION ${CMAKE_INSTALL_BINDIR}
+    RENAME this${PROJECT_NAME}.sh
+)
 
 # install programs
 install(PROGRAMS bin/g4MaterialScan_to_csv


CMakeLists.txt
```diff
@@ -113,6 +113,10 @@ configure_file(templates/setup.sh.in ${CMAKE_CURRENT_BINARY_DIR}/setup.sh @ONLY)
 install(FILES ${CMAKE_CURRENT_BINARY_DIR}/setup.sh
     DESTINATION ${CMAKE_INSTALL_PREFIX}
 )
+install(FILES ${CMAKE_CURRENT_BINARY_DIR}/setup.sh
+    DESTINATION ${CMAKE_INSTALL_BINDIR}
```
veprbl, line 117: If we are talking FHS, then this would be the place
```suggestion
    DESTINATION ${CMAKE_INSTALL_LIBEXEC}
```
In practice, bindir is fine too, of course.

CMakeLists.txt
```diff
@@ -113,6 +113,10 @@ configure_file(templates/setup.sh.in ${CMAKE_CURRENT_BINARY_DIR}/setup.sh @ONLY)
 install(FILES ${CMAKE_CURRENT_BINARY_DIR}/setup.sh
     DESTINATION ${CMAKE_INSTALL_PREFIX}
```
veprbl, line 114: Any deprecation plans for this?

veprbl (approved):

CMakeLists.txt
wdconinc: Right so. I'm worried people aren't going to find libexec very convenient...

CMakeLists.txt
wdconinc: In advance of 24.05, modify the template for setup.sh to warn if $0 is setup.sh instead of thisepic.sh.

wdconinc: #700

PR #704, opened 2024-04-11
EEEMCal geometry update

veprbl:

Updates:
 - Change thickness of outer support volume
 - Removing steel from filling gaps between cells and outer support
 - New inner support collar
 - Remove copper between inner support and cells
 - Adjust cell width
 - Adjust intercell wrappers thicknesses

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __


veprbl:
Comparison of this geometry (salmon) to CAD drawing provided (emerald):
image
Funky irregular rows in the middle are not implemented on purpose.

veprbl:
As we are finalizing 24.06.0 geometry, this could use a review.

wdconinc:
> As we are finalizing 24.06.0 geometry, this could use a review.

Started looking but it's a fairly sizeable changeset.

src/HomogeneousCalorimeter_geo.cpp
```diff
@@ -1,22 +1,5 @@
 // SPDX-License-Identifier: LGPL-3.0-or-later
```
wdconinc, line 1: No one else is using this, right?

veprbl: This file? No other detector, currently. Possibly, in the future this would need to, again, support some beamtest-like configurations.

src/HomogeneousCalorimeter_geo.cpp
```diff
@@ -58,6 +40,63 @@ template <class XmlComp> Position get_xml_xyz(XmlComp& comp, dd4hep::xml::Strng_
...
+static Volume inner_support_collar(Detector& desc, xml_comp_t handle) {
+  // This consists of two circular tubes joined by straight sections
+
+  Material inner_ring_material = desc.material(handle.materialStr());
+
+  double electron_rmin         = handle.attr<double>(_Unicode(electron_rmin));
+  double electron_rmax         = handle.attr<double>(_Unicode(electron_rmax));
+  double proton_rmin           = handle.attr<double>(_Unicode(proton_rmin));
+  double proton_rmax           = handle.attr<double>(_Unicode(proton_rmax));
+  double straight_section_tilt = handle.attr<double>(_Unicode(straight_section_tilt));
```
wdconinc, line 52: Is the tilt of the section the basic parameter here? Should this not be the separation of the centers, from which the tilt is derived?

wdconinc (approved):
Looks good. Optional comment.

PR #713, opened 2024-04-20
feat: trigger eic_container build to catch build errors

wdconinc:
This PR causes changes in this repository to trigger a container rebuild on eicweb to catch errors like https://eicweb.phy.anl.gov/containers/eic_container/-/jobs/2955907 during container build which don't happen during detector benchmark custom build. Ultimately this should make the detector benchmark trigger unnecessary and simplify the workflows with which our benchmark pipelines are activated.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


wdconinc:
This PR now triggers eic_container builds, and then indirectly the containers for both default and nightly. At some point we can fix the container to only trigger default when the container is triggered by an epic geometry push.

veprbl (dismissed):
Looks good. This will allow to remove some custom geometry-building logic from common_bench.

veprbl (approved):

PR #716, opened 2024-04-29
scripts/view*: use bash from $PATH

veprbl:
Existence of /bin/bash is not guaranteed.


veprbl:
There is some utility to shellcheck in CI, but I don't think it would have an opinion on /bin/bash.

PR #717, opened 2024-04-29
bin/generate_prim_file: use POSIX process groups instead of psutil

veprbl:
psutil is an external package. I'd rather not have to worry about providing it.


wdconinc (approved):
Yeah, that looks a lot less janky too.

PR #720, opened 2024-05-05
Add more vacuum at z>40m

michael-pitt:
This RP will fill the hadron line downstream with a vacuum to allow faster simulation of ion fragments.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [X] Other: This makes the simulation run faster in case of eA collisions in which ions (or ion fragments) are defined as stable particles and propagate downstream the Hadron beamline (x10 improvement).

- [X] Changes have been communicated to collaborators: Presented at the FF WG meeting on May 14 https://indico.bnl.gov/event/23379/#15-forward-detectors-with-epic 

Besides the execution time, this PR should keep the results unchanged. 

no

### DIFF
--- a/src/magnetVacuumFF.cpp
+++ b/src/magnetVacuumFF.cpp
@@ -203,7 +203,7 @@ static Ref_t create_detector(Detector& det, xml_h e, SensitiveDetector /* sens *
     Volume vpiece(piece_name, magnetPiece, m_Vac);
     sdet.setAttributes(det, vpiece, x_det.regionStr(), x_det.limitsStr(), vis_name);
 
-    auto pv = assembly.placeVolume(
+    assembly.placeVolume(
         vpiece, Transform3D(RotationY(rotation_magnet[pieceIdx]),
                             Position(x_elem_magnet[pieceIdx], y_elem_magnet[pieceIdx],
                                      z_elem_magnet[pieceIdx])));
@@ -227,7 +227,7 @@ static Ref_t create_detector(Detector& det, xml_h e, SensitiveDetector /* sens *
     Volume vpiece(piece_name, gapPiece, m_Vac);
     sdet.setAttributes(det, vpiece, x_det.regionStr(), x_det.limitsStr(), vis_name);
 
-    auto pv = assembly.placeVolume(
+    assembly.placeVolume(
         vpiece, Transform3D(RotationY(angle_elem_gap[correctIdx]),
                             Position(x_gap[correctIdx], 0.0, z_gap[correctIdx])));
   }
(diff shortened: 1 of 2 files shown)


michael-pitt:
Hi @wdconinc  and @ajentsch 

This updated was communicated to the FF WG, and it is ready to be merged with the master. It should improve the execution time of the simulation for some processes involving particles that propagate downstream the hadronic beampipe

ajentsch (dismissed):

src/magnetVacuumFF.cpp
```diff
@@ -270,6 +271,33 @@ static Ref_t create_detector(Detector& det, xml_h e, SensitiveDetector /* sens *
...
+    sdet.setAttributes(det, vpiece, x_det.regionStr(), x_det.limitsStr(),
+                       "InvisibleNoDaughters"); // make invisible instead of AnlBlue
+
+    auto pv = assembly.placeVolume(vpiece, Transform3D(RotationY(rotation_magnet[pieceIdx]),
+                                                       Position(endGap_x, 0.0, endGap_z)));
+
+    pv.addPhysVolID("sector", 1);
+
+    DetElement de(sdet, Form("sector%d_de", numGaps + numMagnets + 1), 1);
+    de.setPlacement(pv);
```
veprbl, line 297:
```suggestion
```

Not needed for dummy volume.

michael-pitt: I copy paste this block from a different place. Does it mean that the dummy volume needs to be removed everywhere?  
https://github.com/eic/epic/blob/d6746500443b3e21bee7991998e2e12095d15006/src/magnetVacuumFF.cpp#L210-L213
https://github.com/eic/epic/blob/d6746500443b3e21bee7991998e2e12095d15006/src/magnetVacuumFF.cpp#L237-L240
https://github.com/eic/epic/blob/d6746500443b3e21bee7991998e2e12095d15006/src/magnetVacuumFF.cpp#L266-L269

veprbl: Yes. None of those make sense, could you please remove.

michael-pitt: ok, removed

veprbl:
There is a small extrusion from the world_volume, is that due to a rotation?

veprbl:
Perhaps some rotated width is not accounted for.

michael-pitt:
The introduced vacuum volume (tube) is placed between the last magnet (at z=40.74293743+4.4/2 meter) and the end of the physical volume (at z = 100 meters) and aligned with the last magnet rotation angle of -28.22mrad. The tube has radius of 4 times the B2 magnet radius (4*0.2 m). Hence at the end of the physical volume small part of the tube extends beyond the edge: 

image

I'm open to suggestions. The easiest will be to shorten the length by 0.2*4*tan(-28.22mrad)

simonge:
> > I'm open to suggestions. The easiest will be to shorten the length by 0.2x4xtan(-28.22mrad)
> 
> This is exactly what we want. Ideally, this is done in the compact xml file as an expression.

Would the CutTube approach (example present in the same file) be slightly better than shortening a tube.

veprbl (approved):
image

Looks fine

PR #721, opened 2024-05-06
calibrations.xml: update to use latest tof.lut and pfrich.lut

veprbl:
This updates to use the latest LUT from https://github.com/eic/epic-data/pull/12
No binning change, so we don't need any file renames.


wdconinc (dismissed):

wdconinc (approved):

wdconinc:
Force merged on account of just waiting for broken eicweb.

PR #723, opened 2024-05-08
linux-eic-shell.yml: provide PIPELINE_NAME to detector_benchmarks

veprbl:
This takes advantage of https://github.com/eic/detector_benchmarks/pull/20


veprbl:
This also could use https://github.com/eic/trigger-gitlab-ci/pull/11

wdconinc (approved):
Can I use PR titles to inject new variables when I use a `\n` in the title?

Issue #730, opened 2024-05-21
Geometry Matrix Changes


Chao1009:
There are some changes in the latest geometry matrix, which could potentially affect many detector subsystems:
https://docs.google.com/spreadsheets/d/1WEGXm7pbVmUkJ_gGimLgE9as6nY1O2QEwOB8Z6GEU1E/edit#gid=1829269715

Detector subsystem liaisons need to confirm the changes are within expectation and are implemented in the simulation.

Chao1009:
Related PRs:
Barrel EMCal: #729

Chao1009:
@mariakzurek @ruse-traveler @DrKKwd @ssedd1123 @chchatte92 @mposik1983 @ShujieL @kkauder @johnny8266 @jizhongling @FriederikeBock @rymilton @rdom 
Would you please check the latest geometry table and either confirm the simulation does not need any change or open an issue if any detector parameters need to be changed?

ruse-traveler:
Confirming: BHCal does not need any change.

chchatte92:
For the dRICH in the table the aerogel cone is extended up to 33 cm. However, in the attached image is still pointing that the aerogel cone should be around 25 cm. This 25 cm was chosen in order not to shadow the photons in the wall of the aerogel cone for high rapidity particles. 
The dRICH start has already been moved to 198 cm.

DevilsZ:
For the Barrel ToF, I don't see any changes in the recent matrix from the previous version.

ShujieL:
The Si disk volume changes from 0.75 to 0.8, but I don't think this parameter is  used explicitly anywhere in simulation.

mariakzurek:
#729 addresses the changes in the recent Barrel ECal geometry. So things are in order now. The Detector Envelopes Matrix requires an update (it looks like it is using a very old parameters for some of the values, e.g., depth of the BIC), and we will take care of that.

PR #732, opened 2024-05-21
Inner mpgd barrel geo update

ybedfer:
Update the MPGD inner barrel, making it a set of cylindrical, as opposed to flat, tiles. 

- [x] New feature (issue #730 )

Yes. Users need to update their DD4hep private repository to revision #c3b305ce of the AIDASoft/DD4hep github, see "https://github.com/AIDASoft/DD4hep/commit/c3b305ce".

Yes.


ybedfer:
### Plots showing the distributions of:
  - Simulated Hits,  
  - Reconstructed Hits,  
  - Residuals (Rec-Sim). 
This, for a gun of µ^- in the range 10 GeV, 15 < theta < 165 deg.
[cCyMBaLSim.pdf](https://github.com/user-attachments/files/16365929/cCyMBaLSim.pdf)
[cCyMBaLRec.pdf](https://github.com/user-attachments/files/16365939/cCyMBaLRec.pdf)
[cCyMBaLRes.pdf](https://github.com/user-attachments/files/16365941/cCyMBaLRes.pdf)

wdconinc:
FYI, https://eicweb.phy.anl.gov/containers/eic_container/-/merge_requests/963 for inclusion of https://github.com/AIDASoft/DD4hep/pull/1294.

compact/tracking/mpgd_barrel.xml
```diff
@@ -21,70 +21,91 @@
...
+    <constant name="MMInnerSensor_R" value="MMInnerSector_R + rmin2Sensor" />
+    <constant name="MMOuterSensor_R" value="MMOuterSector_R + rmin2Sensor" />
   </define>
 
   <detectors>
     <detector
       id="TrackerBarrel_2_ID"
       name="InnerMPGDBarrel"
-      type="epic_TrackerBarrel"
+      type="MPGDCylinderBarrelTracker"
```
wdconinc, line 68: We prepend `epic_` to the plugin names (detector types) to avoid name clashes when Detector 2 reuses our code (and to avoid clashes with detector types defined centally by DD4hep; though not likely to be an overlap here).

compact/tracking/mpgd_barrel.xml
```diff
@@ -21,70 +21,91 @@
...
-      <layer module="InnerMPGDBarrel_Mod1" id="1" vis="TrackerMMGASLayerVis">
+
+      <layer module="InnerMPGDBarrel_Mod1" id="1" vis="InvisibleWithDaughters">
         <barrel_envelope
           inner_r="InnerMPGDBarrelLayer_rmin"
           outer_r="InnerMPGDBarrelLayer_rmax"
           z_length="InnerMPGDBarrelLayer_length"
           z0="InnerMPGDBarrel_zoffset"/>
         <layer_material surface="inner" binning="binPhi,binZ" bins0="MPGDBarrelStave_count" bins1="100" />
-        <layer_material surface="outer" binning="binPhi,binZ" bins0="MPGDBarrelStave_count" bins1="100" />
```
wdconinc, line 85: @ShujieL may confirm, but we have both inner and outer surfaces to ensure that material maps can allow for both approaches.

compact/tracking/mpgd_barrel.xml
```diff
@@ -98,8 +119,16 @@
...
+      <comment> 768 strips in phi yield a segmentation of 2pi/8/768 ~= 1mrad. Which in turn, yields ~160 um in resolution.
+      </comment>
+      <segmentation type="MultiSegmentation" key="sensor">
+        <comment> In preparation for a 2-sensors-w/-strip setup. Presently sensors w/ odd key values do not exist. </comment>
+        <segmentation name="InnerPhi" type="CylindricalGridPhiZ" key_value="0" grid_size_phi="1*mrad" grid_size_z="0.150*mm*sqrt(12)" radius="MMInnerSensor_R" />
+        <segmentation name="InnerZ"   type="CylindricalGridPhiZ" key_value="1" grid_size_phi="1*mrad" grid_size_z="0.150*mm*sqrt(12)" radius="MMInnerSensor_R" />
+        <segmentation name="OuterPhi" type="CylindricalGridPhiZ" key_value="2" grid_size_phi="1*mrad" grid_size_z="0.150*mm*sqrt(12)" radius="MMOuterSensor_R" />
+        <segmentation name="OuterZ"   type="CylindricalGridPhiZ" key_value="3" grid_size_phi="1*mrad" grid_size_z="0.150*mm*sqrt(12)" radius="MMOuterSensor_R" />
+      </segmentation>
+      <id>system:8,layer:4,module:12,sensor:2,phi:32:-16,z:-16</id>
```
wdconinc, line 131: Does the 16 bits in `z` still capture the full extent of what was 18 bits in `y` with the same grid size?

src/MPGDCylinderBarrelTracker_geo.cpp
```diff
@@ -0,0 +1,563 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2022 Whitney Armstrong
```
wdconinc, line 2:
```suggestion
// Copyright (C) 2022-2024 Whitney Armstrong, Nivedith Ramasubramanian, Yann Bedfer
```

src/MPGDCylinderBarrelTracker_geo.cpp
```diff
@@ -0,0 +1,563 @@
...
+    // Present detector constructor can only handle ONE <module> tag
+    printout(ERROR, "MPGDCylinderBarrelTracker", "Number of modules = %d. Must be = 1",
+             (int)modules.size());
+    throw runtime_error("Logics error in building modules.");
+  }
+  xml_comp_t x_mod   = modules;
+  string m_nam       = x_mod.nameStr();
+  double stave_width = dimensions.width(), det_length = dimensions.length();
+  printout(DEBUG, "MPGDCylinderBarrelTracker", "Module \"%s\" width = %.2f cm length = %.2f cm",
+           m_nam.c_str(), stave_width, det_length);
```
wdconinc, line 99: Generally safer to divide out units when printing value with explicit unit:
```suggestion
  printout(DEBUG, "MPGDCylinderBarrelTracker", "Module \"%s\" width = %.2f cm length = %.2f cm",
           m_nam.c_str(), stave_width / cm, det_length / cm);
```

src/MPGDCylinderBarrelTracker_geo.cpp
```diff
@@ -0,0 +1,563 @@
...
+
+  sens.setType("tracker");
+
+  // ********** MODULE
+  // ***** ONE AND ONLY ONE MODULE
+  xml_coll_t modules(x_det, _U(module));
+  if (modules.size() != 1) {
+    // Present detector constructor can only handle ONE <module> tag
+    printout(ERROR, "MPGDCylinderBarrelTracker", "Number of modules = %d. Must be = 1",
+             (int)modules.size());
```
wdconinc, line 92: Unneeded cast:
```suggestion
    printout(ERROR, "MPGDCylinderBarrelTracker", "Number of modules = %u. Must be = 1",
             modules.size());
```

wdconinc: Applies below.

src/MPGDCylinderBarrelTracker_geo.cpp
wdconinc: Applies below.

src/MPGDCylinderBarrelTracker_geo.cpp
```diff
@@ -0,0 +1,563 @@
...
+    // "rsensor" is supposed to have been set equal to the radius specified in
+    // the segmentation and is used here to double-check the consistency
+    // between the stack of module components and that segmentation radius.
+    double rsensor;
+    int nphi;
+    double offset;
+    int io;
+  } StaveModel;
+  StaveModel staveModels[4];
+  int sector2Models[2][2];
```
wdconinc, line 163: Can this use a map?

src/MPGDCylinderBarrelTracker_geo.cpp
```diff
@@ -0,0 +1,563 @@
...
+#include <array>
+#include "DD4hepDetectorHelper.h"
+
+using namespace std;
+using namespace dd4hep;
+using namespace dd4hep::rec;
+using namespace dd4hep::detail;
+
+#include "Math/Vector2D.h"
+using namespace ROOT::Math;
```
wdconinc, line 28: Can we use only the XYVector (I think that is all that is used)?

wdconinc:
Updated branch to restart CI pipelines with updated DD4hep installation.

compact/tracking/mpgd_barrel.xml
```diff
@@ -21,70 +21,91 @@
     </comment>
 
     <constant name="InnerMPGDBarrelLayer_length"            value="InnerMPGDBarrel_length" />
-    <constant name="InnerMPGDBarrelLayer_thickness"         value="5*cm" />
-    <constant name="InnerMPGDBarrelLayer_rmin"              value="InnerMPGDBarrel_rmin"/>
-    <constant name="InnerMPGDBarrelLayer_rmax"              value="InnerMPGDBarrelLayer_rmin + InnerMPGDBarrelLayer_thickness"/>
+    <constant name="InnerMPGDBarrel_hack"                   value="0*cm"/>
```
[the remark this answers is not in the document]
ybedfer: I haven't followed the whole history of the file: this "hack" was there when I first got involved and I have understood it since as some kind of adjustment variable. But giving it now a second thought, I agree w/ you: it can be removed.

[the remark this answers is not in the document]
mposik1983: I agree this can be removed. I think it is a left over debug artifact from when Niv was working on the geometry

compact/tracking/mpgd_barrel.xml
ybedfer: But then "./src/MPGDCylinderBarrelTracker_geo.cpp" has to be also modified, hasn't it?

mposik1983: Yes, you will need to prepend epic_ to MPGDCylinderBarrelTracker in the .cpp file too

compact/tracking/mpgd_barrel.xml
mposik1983: @wdconinc I think that is correct. I remember when going over material mapping with @ShujieL she mentioned she wanted to add two surfaces for the material to obtain both approaches

ybedfer: I now understand that there is something wrong here, concerning "<layer_material/>", which I have overlooked. "MPGDBarrelStave_count" is set =8 having in mind that whatever do-loop including it must have it as its stopping  limit. => Therefore, something like bins0="0" bins1="MPGDBarrelStave_count - 1" />. To makes things clearer: CyMBaL is subdivided into 8 tiles of cylindrical shape along phi, whereas previously the barrel shape along phi was a polygon w/ 100 sides. But then I have to admit that I do not what this <layer_material/> line of instruction is supposed to do: I do not know myself how to modify it. And I would be willing to learn, btw. Where in the software is this piece of info used, e.g.?

compact/tracking/mpgd_barrel.xml
ybedfer: I think so. My experience with this parameter is that when it is set too tight, the processing aborts with an error. Why did I choose 16 in phi and 16 in Z? Because our tiles are approximately square in size. => phi and Z play tend to play a symmetric role. Also, it's a first guess, the geometry of the tiles is not yet finalized: it's some kind agnostic first guess.

src/MPGDCylinderBarrelTracker_geo.cpp
ybedfer: OK.

compact/tracking/mpgd_barrel.xml
mposik1983: I think adding the deleted line back in should be ok. These are related to ACTS and tell it how to map the material onto the inner and outer layer surface. "binning" give the bin types, e.g. phi,Z. bins0 is number of phi bins and bins1 is for bins in z that the material is mapped to. You can see some explanation of the "inner" and "outer" surfaces here:https://acts.readthedocs.io/en/v7.0.0/core/material.html 

and some of parameter meanings here: https://acts.readthedocs.io/en/latest/plugins/dd4hep.html#possible-parameters-acts-will-interpret

src/MPGDCylinderBarrelTracker_geo.cpp
wdconinc: I meant something like this:
```suggestion
using ROOT::Math::XYVector;
```

src/MPGDCylinderBarrelTracker_geo.cpp
ybedfer: I would keep it as is for the time being, with fixed dimensions. The reason being that, although I tried to make it as general as possible, I haven't checked that more complicated settings, with more than two models or more than two distinct radii per model, are consistently dealt with later on in the code. Fixed dimensions act as a reminder of the fact.

compact/tracking/mpgd_barrel.xml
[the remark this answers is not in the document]
ybedfer: Done. See commit #7403586 ("https://github.com/eic/epic/commit/7403586").

compact/tracking/mpgd_barrel.xml
ybedfer: Done. See commit #7403586 ("https://github.com/eic/epic/commit/7403586").

compact/tracking/mpgd_barrel.xml
ybedfer: I implemented the duplication (inner += outer). See commit #7403586 ("https://github.com/eic/epic/commit/7403586").
But I'm not sure  that the settings of "bins(0|1)" are correct. What binning is here at stake?: I haven't understood. And the setting of "bins0" was originally set =128 (via MPGDBarrelStave_count=128). Does it make sense that it now goes from 128 to only 8? Anyway that does not seem to have any impact on the distributions I am monitoring (Sim and Rec hits eicrecon log).

mposik1983: My understanding of this is that "inner_r" defined in the <barrel_envelope> tag creates an ACTS surface cylinder surface with radius of inner_r and length "z_length", which ACTS projects near by materials onto. The material on this cylindrical surface is binned in phi and z, with Bins0 = number of phi bins, and Bins1 = the number of Z bins used. There is a second surface, "outer_r", that does a similar thing to create an outer material surface. In the previous barrel, a cylinder was approximated using 128 staves, and the phi binning was defined based on the number of staves. For your updated cylinder, I think 8 phi bins would work, e.g. 1 phi bin per tile. A finer binning could be used and would result in a finer material mapping to these surfaces.

wdconinc: If you use 8 phi bins for a detector with 8-fold symmetry, then your bins are going to be filled with the same value and you might as well use 1 bin... What you are trying to account for here is the variation in material within these 2pi/8 bins: if a track is closer to the edge of a tile, the projection of material on a cylinder will be different by 1/cos(2pi/16) (if I did this correctly without a napkin), i.e. an 8% effect.

In some sense, you may want to think about the size of features between tiles, e.g. areas of overlap or gaps. The angular dimension of those features are what should define the bin size of the material map approximation.

ybedfer: I increased the binning so as to be sensitive to ~2 cm, i.e. the cumulated width of two neighboring detector frames.

ybedfer:
Hello,
What could I do concerning the failing check "eicweb/reconstruction_benchmark"? I can see that some (I haven't checked them all) jobs are crashing w/:
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
#5  0x00007fb9ab939b58 in draw_surfaces(std::shared_ptr<Acts::TrackingGeometry const>, Acts::ContextType, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) () from /opt/local/lib/libJugTrackPlugins.so
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Does this mean that it's connected w/ the specification of the sensitive surface? or the <layer_material/> which I modified lately?
The benchmarks seem to be available on github ("git@eicweb.phy.anl.gov:EIC/benchmarks/reconstruction_benchmarks.git"). "https://eicweb.phy.anl.gov/EIC/benchmarks". Can I run them, with a limited amount of training and effort?

wdconinc:
I noticed the failing reconstruction_benchmarks check yesterday and looked into it. The issue is in legacy code, and stems from an incomplete list of supported surfaces in our export to a 3D graphics file for visualization and debugging. We have fixed the equivalent code in our up-to-date code, and yesterday I propagated a fix to the legacy code as well. It takes a day or so for this to propagate to the containers in which we run the checks, so I suspect this will be resolved by Monday at the latest.

ybedfer:
Hello, 
There is no progress.
Is it because something went wrong in your fix this business of "incomplete list of supported surfaces"? Or is it simply that the job has not been restarted?

wdconinc:
I re-triggered the job and it now succeeded. All checks now pass.

compact/tracking/mpgd_barrel.xml
wdconinc: At 16 bits and 0.150*mm*sqrt(12) grid size, the maximum extent is 34 m, so that should be fine.

compact/tracking/mpgd_barrel.xml
mposik1983, line 1: @ybedfer you should add your name as author too

mposik1983 (approved):

compact/tracking/mpgd_barrel.xml
wdconinc: ecd07f0

wdconinc (approved):

PR #741, opened 2024-06-03
Update material map

ShujieL:
New material map for craterlake with updated service cones. See https://eicweb.phy.anl.gov/EIC/detectors/athena/-/issues/153#note_27320

Update material map for the new service cone

- [ ] Bug fix (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x ] Tests for the changes have been added
- [x ] Changes have been communicated to collaborators

N/A
it should improve the EICrecon/ACTS tracking performance.

### DIFF
--- a/compact/tracking/definitions_craterlake.xml
+++ b/compact/tracking/definitions_craterlake.xml
@@ -197,7 +197,7 @@
     <plugin name="epic_FileLoader">
       <arg value="cache:$DETECTOR_PATH:/opt/detector"/>
       <arg value="file:calibrations/materials-map.cbor"/>
-      <arg value="url:https://eicweb.phy.anl.gov/EIC/detectors/athena/uploads/6253bfcb92dd6bc236dd9a06084614ba/material-maps.cbor"/>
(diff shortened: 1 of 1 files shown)


veprbl (approved):
I don't quite understand how to use the material_map artifact to validate this. Can confirm that this doesn't completely break the reconstruction.

veprbl:
Looks like this improves number of hits per track?

image

veprbl:
full report: https://veprbl.github.io/capybara-reports/486edc424c3be862e990721e871421d3/

Simple-Shyam:
> Looks like this improves number of hits per track?
> 
> image

@veprbl @ShujieL I was expecting this behavior to increase the number of hits as mentioned during tracking meeting. That's why I see material map should be updated. When it will be added to calibrations/ directory?

ShujieL:
> Looks like this improves number of hits per track?
> 
> image

Great! That means the algorithm finds more measurements within the chi2 cut now. Let's see how much of this will reflect in the tracking performance study

veprbl:
@Simple-Shyam This will be available soon in 24.06.0

PR #744, opened 2024-06-06
HcalEndcapPInsert: Replace Tungsten layers with Steel layers

sebouh137:
Use only steel absorbers (replace tungsten layers with steel)

Changes the materials used in the calorimeter insert to use only steel layers for the aborbers, rather than using tungsten.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [X] Other: update to materials used in the detector

- [X] Tests for the changes have been added

This could cause issues with EICrecon, since the sampling fraction will need to be changed in the code.  
Yes.

### DIFF
--- a/compact/hcal/forward_insert.xml
+++ b/compact/hcal/forward_insert.xml
@@ -49,8 +49,7 @@
              HcalEndcapPInsertAirThickness "
     />
     <constant name="HcalEndcapPInsertBackplateThickness"    value="HcalEndcapPInsertAbsorberThickness"/>
-    <constant name="HcalEndcapPInsertLayer_NTungstenRepeat" value="10"/>
-    <constant name="HcalEndcapPInsertLayer_NSteelRepeat"    value="54"/>
+    <constant name="HcalEndcapPInsertLayer_NSteelRepeat"    value="65"/>
   </define>
 
   <detectors>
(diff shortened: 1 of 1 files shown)


compact/hcal/forward_insert.xml
```diff
@@ -84,27 +83,6 @@
       <backplate thickness="HcalEndcapPInsertBackplateThickness"/>
       <comment> Front layer to match front walls of LFHCAL modules </comment>
       <comment> Slices will be ordered according to the slice order listed here </comment>
-      <layer repeat="1" thickness="HcalEndcapPInsertFrontLayerThickness" vis="InvisibleWithDaughters">
-        <slice material="Steel235" thickness="LFHCAL_FrontWallThickness" vis="LFHCAL8MModVis" />
```
veprbl, line 88: Front wall is gone? Then, let's remove the constant definition for its thicknesses `HcalEndcapPInsertFrontLayerThickness` and `LFHCAL_FrontWallThickness`.

sebouh137: I mistakenly thought this was just an ordinary layer of the absorber and added it to the count of other absorbers, however apparently this has a somewhat different thickness.  I'll change this so that it is frontplate layer + 64 "normal" layers.

sebouh137: the front wall has been reintroduced in the most recent commit.

veprbl (approved):
LGTM

PR #763, opened 2024-08-05
configurations/: enable MARCO field for central single-detector configurations

veprbl:
Simulating without field is not very useful. Most of the configurations are already doing this, we'd like to make things more consistent.

### DIFF
--- a/configurations/eeemcal_only.yml
+++ b/configurations/eeemcal_only.yml
@@ -1,7 +1,5 @@
 features:
-  #  beampipe:
-  tracking:
-    definitions_craterlake:
-    support_service_craterlake:
+  fields:
+    marco:
   ecal:
     backward_PbWO4:
--- a/configurations/backward_hcal_only.yml
+++ b/configurations/backward_hcal_only.yml
@@ -1,4 +1,6 @@
 features:
+  fields:
+    marco:
   hcal:
     backward:
     backward_endcap_flux:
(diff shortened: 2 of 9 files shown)


configurations/eeemcal_only.yml
```diff
@@ -1,7 +1,5 @@
 features:
-  #  beampipe:
-  tracking:
-    definitions_craterlake:
-    support_service_craterlake:
```
wdconinc, line 5: There are some of these services in front of the EEEMCal, right?

wdconinc (approved):
Only question I'd have is whether someone in track reconstruction is maybe using zero-B field simulations to test reconstruction independent of bending... But I don't think so :-)

configurations/eeemcal_only.yml
veprbl: I think, the reason I removed the services is because none of the other "_only" configurations include those.

wdconinc: Then we have to make sure no one is using them for calorimeter studies.

veprbl: I'm not using them. They were added to get a visual view, and I didn't take them out from the file that was committed. Those services are not in front of EEEMCal.

veprbl: Yes, and the point is to be able to quickly rule-out situations when we can't just blame other detectors :)

PR #773, opened 2024-08-21
Make BIC configurable, add optional setup with 6 layers.

sly2j:
Add an optional setup with a BIC configuration that has all six silicon layers instrumented (versus the current default where only four out of six slots are instrumented). This fully instrumented setup is needed to do further layer optimization studies.

- [ ] Bug fix (issue #__)
- [X] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __


compact/ecal/bic/bic.xml
```diff
@@ -152,7 +156,27 @@
 
       <layer repeat="1" vis="EcalBarrelLayerVis"
              thickness="EcalBarrel_ImagingLayerThickness"
-             space_before="EcalBarrel_FrontSupportThickness">
+             space_before="EcalBarrel_FrontSupportThickness"
+             enable_tray="EcalBarrel_enable_tray_1">
```
wdconinc, line 160: I feel that the boolean `enable_tray` doesn't really make clear in the xml what is enable or disabled. Does it enable everything inside the layer? Does it keep the frame in place? There is nothing else called "tray" so that makes it hard to identify what the effect is. Would something like `enable_layer` be clearer?

wdconinc (commented):
Looks sensible. Minor comment on naming.

compact/ecal/bic/bic.xml
sly2j: I was playing around with it, currently it enables/disables the `<staves>` which is what we want. I could call it `enable_staves` instead. Or I could add it as a member to the `<staves>` field instead for clarity. In the long run we likely won't have staves, and need to change the implementation model of the detector to match our conceptual mechanical model (modules + trays). We could rename this to "tray" at this point if that would be preferred.

wdconinc: I like the approach of adding the enable flag to the element that is enabled or disabled.

sly2j: Done

wdconinc (approved):

PR #779, opened 2024-09-11
fix(ci): run_material_map_validation.sh use v33.1.0 scripts

wdconinc:
This PR updates the version of the scripts used for material map validation to the version of ACTS that is used in the environment.

- [x] Bug fix (issue: https://github.com/eic/epic/actions/runs/10800006504/job/29957261108)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.


wdconinc:
Succeeds now and looks fine in artifacts at https://github.com/eic/epic/actions/runs/10946548910/job/30393563281?pr=779. The FPE does not seem to make the entire workflow invalid.

veprbl (approved):

wdconinc:
@veprbl The geoviewers are failing since we don't have a build_detector phase. I'm adapting the geoviewer jobs to use /opt/detectors/epic-main, see https://eicweb.phy.anl.gov/EIC/benchmarks/geoviewer/-/merge_requests/4. This will require that we trigger the geoviewer job from the container build, and not directly from here. Or did you have a different idea for this?

veprbl:
> @veprbl The geoviewers are failing since we don't have a build_detector phase. I'm adapting the geoviewer jobs to use /opt/detectors/epic-main, see https://eicweb.phy.anl.gov/EIC/benchmarks/geoviewer/-/merge_requests/4. This will require that we trigger the geoviewer job from the container build, and not directly from here. Or did you have a different idea for this?

Yes. That would be in line with our current best approaches. My first reaction was to revert https://eicweb.phy.anl.gov/EIC/benchmarks/common_bench/-/merge_requests/91. In the long term, geoviewer should be just another detector benchmark, so it would do good for us to exercise it to run via a container build.

PR #784, opened 2024-09-19
Mpgd gas mix

mposik1983:
Change the gas mixture in the outer barrel and end cap MPGDs

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ x] Other: __

- [ x] Changes have been communicated to collaborators

No. None
Gas mixture change will lead to slightly different deposited energy distribution in the outer barrel and end cap MPGDs


wdconinc:
Can I suggest that the naming scheme for the gas mixtures can use improvement? Both `Ar10CO2` and `Ar90IsoButane` contain 90% argon...

mposik1983:
> Can I suggest that the naming scheme for the gas mixtures can use improvement? Both `Ar10CO2` and `Ar90IsoButane` contain 90% argon...

I can see how that will lead to confusion. I would like to keep the Ar90IsoButane structure because this is already being used in mpgd_barrel.xml. It looks like Ar10CO2 was used sometime ago as most files show up in depreciated directory. The files not in depreciated are in central_tracker_hybrid_v2.xml, can you remind me if this is still used? I can modify those and the gas mix definition to be Ar90CO2

wdconinc:
Sounds good. If you just rename the Ar10CO2 gas name, then even if it is used (which I don't think it is), it shouldn't matter.

mposik1983:
Ok, I just changed the gas name in materials.xml

wdconinc:
> Ok, I just changed the gas name in materials.xml

Well, we should also change it where it is used. Now we're in an inconsistent state (even if the files aren't used).

mposik1983:
Ok, I believe I have all cases of "Ar10CO2" --> "Ar90CO2"

wdconinc (approved):
Looks good. Thanks for making the changes in adjacent scope!

PR #796, opened 2024-10-12
Tof new gemo

ssedd1123:
Endcap TOF geometry modified to match the newest technical drawings from https://indico.bnl.gov/event/24949/attachments/57520/98759/ePICTOF_WP2_FTOFLayout_09272024v2.pdf.

The current epic geometry has each ETOF module located separately. But the newest drawing shows that a single power board house multiple modules together. This PR groups modules together according to the aforementioned drawing.

Furthermore, the current epic geometry shows that sensors are placed facing front and backing alternatively across rows. However, the latest drawing shows that ETOF are double sided on every row, but with sensors placed below the service hybrid on the front side, and above the service hybrid on the other side. Each module now contains 4 small sensors instead of one big sensor.

Here's the updated ETOF power board configuration,

image

Here's the design drawing,

image

Barrel TOF is also modified. Staves are broken into two so an engagement ring can fit in the middle. LGAD is also located on both sides alternatively, 

image

Cooling tubes are also incorporated into the staves,

image

Latest technical drawing shows that sensor width is half of what was previously proposed. Sensors right next to the central engagement ring are only half sized,

image

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ x] Other: BTOF and ETOF geometry modified according to new technical drawing

No

BTOF ETOF geometries are changed.


src/EndcapTOF_geo.cpp
```diff
@@ -1,9 +1,10 @@
 // SPDX-License-Identifier: LGPL-3.0-or-later
-// Copyright (C) 2022 Nicolas Schmidt
+// Copyright (C) 2022 - 2024, Nicolas Schmidt, Chun Yuen Tsang
 
 /** \addtogroup Trackers Trackers
  * \brief Type: **Endcap Tracker with TOF**.
  * \author N. Schmidt
+ * \modified by C.Y. Tsang 8th Oct, 2024
```
veprbl, line 7:
```suggestion
```
No need to duplicate this information - it's in copyright and in git log.

compact/tracking/tof_endcap.xml
```diff
@@ -107,7 +107,7 @@
     <comment> 1 um padding to not have layer and module touch (ACTS requirement) </comment>
     <constant name="TOFEndcapLayer_thickness"   value="TOFEndcapMod_thickness + 2 * TOFEndcapMod_dz + 1 * um" />
 
-    <constant name="ForwardTOFMod_rmin"         value="ForwardTOFRegion_minR+5.0*mm" />
+    <constant name="ForwardTOFMod_rmin"         value="ForwardTOFRegion_minR-1*cm" />
```
veprbl, line 110: What is this 1 cm? Does it violate some detector envelope? Is it possible to have a named constant for it?

veprbl (dismissed):
Looks great. Tracking benchmarks still look good.

compact/tracking/tof_endcap.xml
ssedd1123: That is correct. Volumes overlap if the minR is too large. Now that I think about it, I can just reduce ForwardTOFRegion_minR by 1 cm. checkOverlap shows 0 overlap when I do that. Let me push the updated version.

ssedd1123: The pull request is updated.

veprbl (approved):

compact/tracking/tof_endcap.xml
veprbl: Makes sense to me. Thanks!

ssedd1123:
@veprbl bench:backwards_ecal exceeded the 2 hours limit and thus eicweb/eic_container failes. I check the logs and the snake jobs are progressing slowly until the very end so it's not like it's getting stuck in a loop. It seems strange to me that my modifications to TOF would have any significant effects on backward ecal. Do you have any idea what's going on?

ssedd1123:
@veprbl Unfortunately I don't think this issue is solved. I looked at other jobs in the merge queue and snakemake is still spawning hundreds of jobs on bench:backwards_ecal.

veprbl:
> @veprbl Unfortunately I don't think this issue is solved. I looked at other jobs in the merge queue and snakemake is still spawning hundreds of jobs on bench:backwards_ecal.

Right, back to a drawing board.

PR #802, opened 2024-11-08
calibrations.xml: add calibrations/onnx/EcalEndcapN_pi_rejection.onnx

veprbl:
See https://github.com/eic/epic-data/pull/22
This is needed to enable https://github.com/eic/EICrecon/pull/1618


wdconinc (approved):
Looks fine to me.

PR #759, opened 2024-07-24
Implement sensor unit (with inactive area) for the silicon vertex layers

ShujieL:
### Briefly, what does this PR introduce?
Added a new geometry plugin to support staves with sensor unit (with inactive area) , and used it for the three inner silicon vertex layers.
The tracking performance will change due to the reduced acceptance. Study in progress

### What kind of change does this PR introduce?
- [ ] Bug fix (issue #__)
- [ x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [x ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [ ] Changes have been communicated to collaborators

### Does this PR introduce breaking changes? What changes might users need to make to their code?
N/A
### Does this PR change default behavior?
N/A

PR #803, opened 2024-11-11
Implement curved RSU structure for the three layers of vertex barrel trackers 

ShujieL:
created a new vertex barrel geometry file vertex_barrel_curved.xml with a new geo plugin to support smooth cylindrical surface and build-in inactive areas. 
This work is a follow-up of https://github.com/eic/epic/pull/759 which creates flat sensors and barrel with staves. 

09.2025 updates:
1. updated the RSU dimension 
Screenshot 2025-09-25 at 17 20 31

2.  added the mechanics and cables for Inner Barrels (L0,L1,L2) according to the drawings which are extracted from the 08.2025 IB step files (contact: Rosario Turrisi, INFN):
[Draft_Half_Asm_L0L1_note.pdf](https://github.com/user-attachments/files/22549897/Draft_Half_Asm_L0L1_note.pdf)
[Draft_Half_Asm_L2_note.pdf](https://github.com/user-attachments/files/22549898/Draft_Half_Asm_L2_note.pdf)
Note: The FPC cable (digital, analog, or signal) contains 70um copper according to MOSAIX FPC stack-up V2, see also ITS3 [report](https://inspirehep.net/literature/2960403). In this simulation, use 100um of aluminum instead. There are 3 layers of cables on the hadron going direction, and 2 layers on the electron going direction. Applied to sheet 5 and 9 of the above draft).  

- [ ] Bug fix (issue #__)
- [ x ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [ x] Changes have been communicated to collaborators

The # of hits on vertex barrels will reduce by a few percent due to the inactive areas.
Didn't see a significant change in tracking efficiency. Performance study on going.


ShujieL:
VertexBarrelEnvelope_length and VertexTrackingRegion_length are not used anywhere. Can we remove them from dump-parameter-table? 

Also, not sure why we keep VertexCheck in the xml file.

ShujieL:
Also, validate-material-map failed due to "Propagation reached the step count limit of 1000 (did 1000 steps)". I remember seeing this error many times in the past but they didn't interrupt the program, a material map would still be generated. Any idea on how to deal with this?   @wdconinc

wdconinc:
> Also, validate-material-map failed due to "Propagation reached the step count limit of 1000 (did 1000 steps)". I remember seeing this error many times in the past but they didn't interrupt the program, a material map would still be generated. Any idea on how to deal with this? @wdconinc

no time to look into it until next week (not even sure then)

veprbl:
Also:
```
MultiSegmentation ERROR Invalid sub-segmentation identifier!
VertexBarrelHandler ERROR cellID: failed to access segmentation for VolumeID: 000000000103B21F [17019423]  [MultiSegmentation: Invalid sub-segmentation identifier!]
VertexBarrelHandler ERROR ....... G4-local:   (49.820797, 1.685985, 1.892819) G4-global:   (34.036452, 36.420795, -19.331181)
VertexBarrelHandler ERROR ....... TGeo-local: (4.982080, 0.168598, 0.189282) TGeo-global: (3.403645, 3.642079, -1.933118)
VertexBarrelHandler ERROR ....... Pre-step:  AV_1555!Module2_upper_58#58!RSU_upper_3#3  SD: VertexBarrel
VertexBarrelHandler ERROR ....... Post-step: AV_11865!VertexBarrelSubAssembly_3#3!VertexBarrel_layer2_1#1  SD: 
DDSim            ERROR Simulation failed!
Geant4Kernel     FATAL +++ Exception while simulating:MultiSegmentation: Invalid sub-segmentation identifier!
```

compact/tracking/vertex_barrel_curved.xml
```diff
@@ -0,0 +1,239 @@
...
+          grid_size_phi="0.02*mm/VertexBarrelSeg1_r" grid_size_z="0.02*mm"
+          radius="VertexBarrelSeg1_r" />
+        <segmentation name="L1" type="CylindricalGridPhiZ" key_value="2"
+          grid_size_phi="0.02*mm/VertexBarrelSeg2_r" grid_size_z="0.02*mm"
+          radius="VertexBarrelSeg2_r" />
+        <segmentation name="L2" type="CylindricalGridPhiZ" key_value="4"
+          grid_size_phi="0.02*mm/VertexBarrelSeg3_r" grid_size_z="0.02*mm"
+          radius="VertexBarrelSeg3_r" />
+      </segmentation>
+      <id>system:8,layer:4,module:12,sensor:3,phi:30:-18,z:-16</id>
```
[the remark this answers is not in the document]
wdconinc: Also, you're allocating 4 bits for layers, which allows for 16 layers. You only use 3 layers, so 3 bits would be sufficient.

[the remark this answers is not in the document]
wdconinc: Also, your bit fields don't add up cleanly to 64 due to sensor being 3 bits wide.

src/SVTBarrelTracker_geo.cpp
```diff
@@ -0,0 +1,324 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+// Copyright (C) 2022 Whitney Armstrong, Jonathan Witte, Shujie Li
```
wdconinc, line 2: Update date.

wdconinc: By which I mean
```suggestion
// Copyright (C) 2022-2024 Whitney Armstrong, Jonathan Witte, Shujie Li
```

src/SVTBarrelTracker_geo.cpp
wdconinc, line 1: Avoid naming the plugin to the specific detector. As far as I understand this will need to be done for other detectors too. We should try to make sure we can reuse this geometry plugin for the other detectors too. Therefore, naming this as `CurvedBarrelTracker_geo` or so is probably useful to consider.

src/SVTBarrelTracker_geo.cpp
```diff
@@ -0,0 +1,324 @@
...
+
+/** Curved Silicon Vertex Tracker Barrel with RSU.
+ *
+ * - Designed to process "vertex_barrel_curved.xml"
+ *
+ * - Derived from "BarrelTrackerWithFrame_geo.cpp".
+ * - Build-in RSU structure with 12 tiles and inactive areas
```
wdconinc, line 9: Number of tiles is likely configurable parameter, not inherent to the geometry plugin.

src/SVTBarrelTracker_geo.cpp
```diff
@@ -0,0 +1,324 @@
...
+  for (xml_coll_t bmat(x_det, _Unicode(boundary_material)); bmat; ++bmat) {
+    xml_comp_t x_boundary_material = bmat;
+    DD4hepDetectorHelper::xmlToProtoSurfaceMaterial(x_boundary_material, params,
+                                                    "boundary_material");
+  }
+
+  Assembly assembly(det_name);
+
+  sens.setType("tracker");
+
```
wdconinc, line 63: I noticed you removed the support and frame sections here. Did you not want to keep the support for this plugin?

src/SVTBarrelTracker_geo.cpp
```diff
@@ -0,0 +1,324 @@
...
+  // loop over the modules
+  for (xml_coll_t mi(x_det, _U(module)); mi; ++mi) {
+    xml_comp_t x_mod = mi;
+    string m_nam     = x_mod.nameStr();
+    double m_rmin    = x_mod.rmin();
+    double m_length  = x_mod.length();
+    double m_width   = x_mod.width();
+    module_length[m_nam].push_back(m_length);
```
wdconinc, line 71: I don't understand why the module length is necessary as a map to vector. Will there be multiple modules with the same name? Do you need the length outside of this scope?

ShujieL:
> > Also, validate-material-map failed due to "Propagation reached the step count limit of 1000 (did 1000 steps)". I remember seeing this error many times in the past but they didn't interrupt the program, a material map would still be generated. Any idea on how to deal with this? @wdconinc
> 
> no time to look into it until next week (not even sure then)

@wdconinc there is a new material mapping script to allow mapping without propagation but requires MaterialMapper from ACTS v34. 
https://github.com/acts-project/acts/blob/main/Examples/Scripts/Python/material_mapping_core.py

Any plan for ACTS version upgrade?

src/SVTBarrelTracker_geo.cpp
ShujieL: I use it to accommodate the upper and lower modules and pass to layer assembly. Is there a  better way?

ShujieL:
This new IB barrel implementation: 12 RSU along z ~ 26cm long, introduces an acceptance gap at eta=0.9 b/c the sensor coverage is 1cm shorter than the nominal 27cm. 

For this reason, we will not merge this PR until the tracker geometry is tweaked to solve this issue.

ShujieL:
@wdconinc @veprbl the IB update is ready for review.

compact/tracking/vertex_barrel.xml
```diff
@@ -1,157 +1,248 @@
...
-      - Currently there are 3 sensor layers. Each is composed of 2 half-cylinders modules with only 40um of silicon thickness.
-      - Both support shells are 300um thick, implemented as the integrated tracker support/service setup
-
+      for the segmentation. Assume sensors are placed at the inner most surface for each layer
     </documentation>
+    <constant name="VertexBarrelSegL0_r" value="VertexBarrelModL0_rmin+SiVertexSensor_thickness/2.0" />
+    <constant name="VertexBarrelSegL1_r" value="VertexBarrelModL1_rmin+SiVertexSensor_thickness/2.0" />
+    <constant name="VertexBarrelSegL2_r" value="VertexBarrelModL2_rmin+SiVertexSensor_thickness/2.0" />
 
-    <constant name="VertexBarrelEnvelope_length"    value="VertexTrackingRegion_length"/>
```
veprbl, line 29: This causes
https://github.com/eic/epic/actions/runs/18144969161/job/51644821946?pr=803#step:5:2226
```
jinja2.exceptions.UndefinedError: 'VertexBarrelEnvelope_length' is undefined
```
Perhaps we can somehow provide some reasonable value for envelope length? Or we can discontinue its output in comparison template https://github.com/eic/epic/blob/fc2eb7a5f35264d46c5c5c04e0404e3b816b8e82/templates/DetectorParameterTable.csv.jinja2#L44-L46

compact/definitions.xml
```diff
@@ -105,15 +105,20 @@ The unused IDs below are saved for future use.
...
+    <!-- <constant name="TrackerCompositeBarrel_0_ID"  value="50"/>
     <constant name="TrackerCompositeEndcapN_0_ID" value="51"/>
     <constant name="TrackerCompositeEndcapN_1_ID" value="52"/>
     <constant name="TrackerCompositeEndcapN_2_ID" value="53"/>
     <constant name="TrackerCompositeEndcapN_3_ID" value="54"/>
     <constant name="TrackerCompositeEndcapP_0_ID" value="55"/>
     <constant name="TrackerCompositeEndcapP_1_ID" value="56"/>
     <constant name="TrackerCompositeEndcapP_2_ID" value="57"/>
-    <constant name="TrackerCompositeEndcapP_3_ID" value="58"/>
+    <constant name="TrackerCompositeEndcapP_3_ID" value="58"/> -->
```
veprbl, line 121: Can remove these.

src/CurvedBarrelTracker_geo.cpp
```diff
@@ -0,0 +1,326 @@
...
+          for (size_t ic = 0; ic < sensVols[kk].size(); ++ic) {
+            PlacedVolume sens_pv = sensVols[kk][ic];
+            DetElement comp_de(mod_elt, std::string("de_") + sens_pv.volume().name(), module);
+            comp_de.setPlacement(sens_pv);
+            auto& comp_de_params =
+                DD4hepDetectorHelper::ensureExtension<dd4hep::rec::VariantParameters>(comp_de);
+            comp_de_params.set<string>("axis_definitions", "XYZ");
+            // comp_de.setAttributes(description, sens_pv.volume(), x_layer.regionStr(), x_layer.limitsStr(),
+            //                       xml_det_t(xmleles[m_nam]).visStr());
+            //
```
veprbl, line 300: What is this doing/not doing for us?
cc @wdconinc

veprbl:
> It keeps failing the backwards_ecal check due to some issue with curve_fit, which I think is irrelevant. Can we give this PR a pass? @wdconinc @veprbl

Yes. We can restart/fix those benchmarks, no issue. For review we care about code itself and what are tracking benchmarks saying about relative performance change.

veprbl (commented):
Checked that material map generation still works after this is merged.

PR #815, opened 2024-12-06
backward_PbWO4: adjust geometry to match latest CAD design

veprbl:

This updates tower geometry to match the latest CAD design.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

Yes


veprbl:
This is ready. Could somebody take a look at his, please?

wdconinc (approved):
Hard to judge. Don't have CAD. Code looks fine.

veprbl:
Good point. I forgot to add some context.

This is comparison of CAD (emerald) to the current main (salmon):
before

And this is comparison to this PR
after

Note that inner support CAD has cooling pipes, which I was are not final and encouraged to not implement yet.

Also note that the outer dodecagon wall appears thinner, but that's because some material implemented in the envelope is not rendered here.

PR #832, opened 2025-02-25
add scaled field files for the early science magnet settings

ajentsch:

This PR adds the necessary XML files to steer the beam correctly for 6 of the 7 early science settings. I have not added the 10x100 GeV Au setting for right now since we are not even producing files for it.

[remedied with new commit] One thing that has NOT been added in the configuration YAML files for craterlake. This is because those YAML files don't have a field for the ion species. 

- [ ] Bug fix (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [ x] Documentation has been added / updated

No, it just adds the magnet settings to steer the ion beams for early science.

No, just adds additional functionality for ions.


veprbl:
I'm confused, how these are supposed to be used if they are not included from any configurations?

ajentsch:
They would have to be included "by hand" in a configuration. If we want to automate it, we will need to adjust the YAML file which is used to generate the "craterlake_[EBEAM}x[PBEAM}.xml" files.

veprbl:
I think that should be done.

ajentsch:
Okay, cool - I just pushed a new commit with those files.

veprbl (approved):
All LGTM

PR #834, opened 2025-02-27
Bug fix on Barrel TOF sensors having incorrect segmentation. 

ssedd1123:

We realize that the cell positions are not in the correct places on some BTOF sensors. Specifically, the cell center for the half-sensors immediately to the left and right of the central gap for the engagement ring are all incorrect. This issue is caused by the segmentation class (CartesianGridXY) applying the wrong offset on the half-sensors. Unfortunately, I do not know of a way to apply a different offset value just for the half-sensor while keeping the original value for the normal sensor.

This pull request resolves that issue by making all sensors half-sized. A full sensor is simply two half-sensors placed side-by-side.

Please read our discussion on the ePIC AC-LGAD TOF weekly meeting for details: https://indico.bnl.gov/event/27008/contributions/103896/attachments/60237/103439/BTOF_geometry_bug_EIC_EPIC.pdf 

- [x ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [X ] Changes have been communicated to collaborators

BTOF now uses CartesianStripX for segmentation rather than CartesianGridXY. However, the field names for cellID remains unchanged since I renamed the identifies for CartesianStripX to match the behavior of CartesianGridXY.

No.


veprbl:
Just a note, I think you can solve it simpler, with multi segmentation defining which offset is used. That would be also future-proof if you need to reposition half sensors with respect to the others.
https://github.com/eic/epic/blob/2e62d1507c2ed3ad8fe3af5bc9d1e798663b064a/compact/tracking/mpgd_barrel.xml#L126-L138

ssedd1123:
Thanks @veprbl for your suggestion. Yeah, that implementation make more sense. I did not know that segmentation class exist. Now I've learnt something new. I will update my pull request soon using MultiSegmentation.

ssedd1123:
The pull request is updated to using MultiSegmentation.

veprbl (approved):
Diff LGTM

PR #859, opened 2025-05-06
linux-eic-shell.yml: fix download step in build-artifacts-page

veprbl:
This should help with issues like https://github.com/eic/epic/actions/runs/14817178151/job/41749142343#step:7:7

Doesn't look like `artifacts/` were supposed to be used https://github.com/eic/EICrecon/blob/734cd480d8637ac50837db0d70da0ac48a1498d5/.github/workflows/linux-eic-shell.yml#L1161C11-L1162

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -779,7 +779,8 @@ jobs:
     - name: Download GitHub Pages staging artifact
       uses: actions/download-artifact@v4
       with:
-        path: artifacts/
+        name: github-pages-staging
+        path: publishing_docs/
     - name: Build jekyll page
       uses: actions/jekyll-build-pages@v1
     - name: Make PR summary page


.github/workflows/linux-eic-shell.yml
```diff
@@ -779,7 +779,8 @@ jobs:
     - name: Download GitHub Pages staging artifact
       uses: actions/download-artifact@v4
       with:
-        path: artifacts/
```
wdconinc, line 782: I think this is the wrong direction of this change.  See links in README.md (which becomes https://epic-eic.org/epic/index.html), and `reports/artifacts.md`.

I think what we may want to do is keep artifacts, and then publish _site or so.

wdconinc: And we've been advertising links like https://eic.github.io/epic//artifacts/epic_craterlake_no_bhcal.stp/epic_craterlake_no_bhcal.stp to others, like project and similar gdml files to Makoto Asai. We should try not to have them break.

wdconinc: Maybe we should just revert #856 and make it work first.

PR #864, opened 2025-05-12
fix: handle dead symlinks in FileLoaderHelper

wdconinc:
This PR fixes an infrequent recurring issue in the handling of calibrations and fieldmaps. It occurs when a file handle symlink is pointing to a hash symlink that is itself pointing to a non-existent location (the following is the situation after the error occurs):
```
jug_dev> wdconinc@menelaos:~/git/epic$ ls -al calibrations/
total 152616
drwxr-xr-x  6 wdconinc wdconinc     4096 May 12 13:16 .
drwxrwxr-x 32 wdconinc wdconinc     4096 May 12 13:16 ..
lrwxrwxrwx  1 wdconinc wdconinc      109 May 12 13:16 0820c1b054653fef -> /opt/detector/epic-git.63e5b941500f64c35449c29850b0bd60717adb2c_main/share/epic/calibrations/0820c1b054653fef
-rwxrwxrwx  1 wdconinc wdconinc      178 Apr 27 11:14 5a54959cc63eeb5b -> /opt/software/linux-debian12-x86_64_v2/gcc-12.2.0/epic-git.4f29e576c85b7342bf0b3b3f0329e4f4583b3f5f_main-
lrwxrwxrwx  1 wdconinc wdconinc       16 Apr 27 11:14 materials-map.cbor -> 5a54959cc63eeb5b
```
Above, the link `0820c1b054653fef` to the new `materials-map.cbor` in `/opt` has already been created, but `materials-map.cbor` cannot be relinked to point to `0820c1b054653fef`. The file handle symlink `materials-map.cbor` is indicated to not exist even if the symlink exists but the file it points to does not. Since it does not exist, the code tries to create the symlink, but fails because it already exists.

- [x] Bug fix (issue: headaches when updating calibrations files)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/FileLoaderHelper.h
+++ b/src/FileLoaderHelper.h
@@ -188,7 +188,7 @@ inline void EnsureFileFromURLExists(std::string url, std::string file, std::stri
                  "but a file already exists there and we will not remove it automatically");
     printout(ERROR, "FileLoader", "hint: backup the file, remove it manually, and retry");
     std::_Exit(EXIT_FAILURE);
-  }    
+  }
   // file_path now does not exist
 
   // symlink file_path to hash_path
--- a/src/FileLoaderHelper.h
+++ b/src/FileLoaderHelper.h
@@ -180,14 +180,16 @@ inline void EnsureFileFromURLExists(std::string url, std::string file, std::stri
       }
     }
   } else {
-    // failure mode: file exists but not symlink, and we won't remove files
-    printout(ERROR, "FileLoader",
-             "file " + file_path.string() + " already exists but is not a symlink");
-    printout(ERROR, "FileLoader",
-             "we tried to create a symlink " + file_path.string() + " to the actual resource, " +
-                 "but a file already exists there and we will not remove it automatically");
-    printout(ERROR, "FileLoader", "hint: backup the file, remove it manually, and retry");
-    std::_Exit(EXIT_FAILURE);
+    if (fs::exists(file_path)) {
+      // failure mode: file exists but not symlink, and we won't remove files
+      printout(ERROR, "FileLoader",
+               "file " + file_path.string() + " already exists but is not a symlink");
+      printout(ERROR, "FileLoader",
+               "we tried to create a symlink " + file_path.string() + " to the actual resource, " +
+                   "but a file already exists there and we will not remove it automatically");
+      printout(ERROR, "FileLoader", "hint: backup the file, remove it manually, and retry");
+      std::_Exit(EXIT_FAILURE);
+    }
   }
   // file_path now does not exist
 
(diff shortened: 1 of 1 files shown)


veprbl:
This branch is taken when `(!fs::exists(file_path)) && fs::is_symlink(file_path)`. Previous solutions were already aimed at detecting broken symlinks. I wonder if there is a way to correct the original logic. I think we just need to invert the hierarchy of checks to not let `fs::exits` follow the symlinks:
```c++
  if (fs::is_symlink(file_path)) {
    auto file_status = fs::symlink_status(file_path);
    if (fs::exists(file_status)) { // redundant
      // file already exists
      // ...
    }
  } else {
    auto file_status = fs::symlink_status(file_path);
    if (fs::exists(file_status)) {
      // file already exists
      // ...
    }
  }
```

wdconinc:
> This branch is taken when `(!fs::exists(file_path)) && fs::is_symlink(file_path)`. Previous solutions were already aimed at detecting broken symlinks. I wonder if there is a way to correct the original logic.

We tried to fix this previously in https://github.com/eic/epic/pull/762, but failed due to a partial understanding of the failure mode. There is no other dead symlink handling in the code that should catch this.

> I think we just need to invert the hierarchy of checks to not let `fs::exits` follow the symlinks

Sensible idea. I don't know if it will make this plugin more maintainable though.

veprbl:
My point is that we should be ending up in the already existing 
```
// link points to incorrect path
```
branch instead of adding a new one.

wdconinc:
... Regretting not going with `system("ln -sf")` ;-)

src/FileLoaderHelper.h
```diff
@@ -171,15 +172,22 @@ inline void EnsureFileFromURLExists(std::string url, std::string file, std::stri
...
     }
+  } else {
+    // failure mode: file exists but not symlink, and we won't remove files
+    printout(ERROR, "FileLoader",
+             "file " + file_path.string() + " already exists but is not a symlink");
+    printout(ERROR, "FileLoader",
+             "we tried to create a symlink " + file_path.string() + " to the actual resource, " +
+                 "but a file already exists there and we will not remove it automatically");
+    printout(ERROR, "FileLoader", "hint: backup the file, remove it manually, and retry");
+    std::_Exit(EXIT_FAILURE);
```
veprbl, line 190:
```suggestion
    if (fs::exists(file_path)) {
      // failure mode: file exists but not symlink, and we won't remove files
      printout(ERROR, "FileLoader",
               "file " + file_path.string() + " already exists but is not a symlink");
      printout(ERROR, "FileLoader",
               "we tried to create a symlink " + file_path.string() + " to the actual resource, " +
                   "but a file already exists there and we will not remove it automatically");
      printout(ERROR, "FileLoader", "hint: backup the file, remove it manually, and retry");
      std::_Exit(EXIT_FAILURE);
    }
```

wdconinc: I'll look at this tomorrow. And maybe write a ci job just to test this.

veprbl: The CI agrees with me.

wdconinc: In this case in particular: just because it compiles doesn't mean this will now work.

PR #867, opened 2025-05-13
Centralize IP6 Beampipe constant definitions

simonge:
Fixes an airgap in the Far-Backward beamline by centralizing the definitions for the beampipe ends.

TaggerTrackerTracks have gone missing from the EICRecon output since the IP6 beampipe update. A visual inspection shows an air gap between the central beampipe and the current far-backward electron beamline (and an overlap with the incoming hadron extension). The air is _hopefully_ responsible for breaking the electron track reconstruction...

image

Centralizing the variables used to define the positions of the ends should fix this and stop something like this happening in the future. The small gap in the incoming hadron beamline is much less important.

image

A more detailed revision of the far-backward beamline is in progress but this is required sooner.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

Hopefully recovers a full vacuum

### DIFF
--- a/compact/definitions.xml
+++ b/compact/definitions.xml
@@ -411,6 +411,50 @@ Examples:
     <constant name="SolenoidBarrel_length"     value="Solenoid_length"/>
 
     <documentation>
+      ## IP6 Beam Pipe Parameters - Note might need cleaning
+    </documentation>
+    <constant name="CrossingAngle" value="-0.025*rad "/>
+    <constant name="ionCrossingAngle" value="CrossingAngle"/>
+    <constant name="electronCrossingAngle" value="0.0"/>
+    <constant name="CrossingSlope" value="CrossingAngle"/>
+
+    <constant name="IPBeampipe_rmax" value="2.501*25.4*mm/2.0"/>
+    <constant name="Beampipe_rmax"   value="IPBeampipe_rmax"/>
+    <constant name="IPBeampipeID" desc="IP6 beam pipe inner diam w/o coating" value="62*mm"/>
+
+    <documentation>
+      ## Central Beam Pipe Parameters
+    </documentation>
+    <constant name="IPBeampipeUpstreamStraightLength" value="74.8 * cm"/>
+    <constant name="IPBeampipeDownstreamStraightLength" value="64.8 * cm"/>
+    <constant name="BeampipeOD" value="62 * mm"/>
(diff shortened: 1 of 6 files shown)


nat93 (approved):
Thanks for requesting my review. Since nothing changed in the central beam pipe, except for variable names and place where they are defined, I am okay with these updates.

compact/definitions.xml
```diff
@@ -410,6 +410,50 @@ Examples:
     <constant name="SolenoidBarrel_rmin"       value="Solenoid_rmin"/>
     <constant name="SolenoidBarrel_length"     value="Solenoid_length"/>
 
+    <documentation>
+      ## IP6 Beam Pipe Parameters - Note might need cleaning
+    </documentation>
+    <constant name="CrossingAngle" value="-0.025*rad "/>
+    <constant name="ionCrossingAngle" value="CrossingAngle"/>
+    <constant name="electronCrossingAngle" value="0.0"/>
+    <constant name="CrossingSlope" value="CrossingAngle"/>
```
wdconinc, line 419: Is this used, and do we want to use "slope" which has vertical angle implications? In my interpretation, slope is zero at the IP.

compact/definitions.xml
```diff
@@ -410,6 +410,50 @@ Examples:
     <constant name="SolenoidBarrel_rmin"       value="Solenoid_rmin"/>
     <constant name="SolenoidBarrel_length"     value="Solenoid_length"/>
 
+    <documentation>
+      ## IP6 Beam Pipe Parameters - Note might need cleaning
+    </documentation>
+    <constant name="CrossingAngle" value="-0.025*rad "/>
+    <constant name="ionCrossingAngle" value="CrossingAngle"/>
+    <constant name="electronCrossingAngle" value="0.0"/>
```
wdconinc, line 418: Arguably, these are not *crossing* angles, since they refer to one beam only. Can we just call these "angle"?

wdconinc (approved):
Looks good. Optional comments since I understand you're just trying to move things around, not rewrite things.

compact/definitions.xml
simonge: I'll open a separate issue where these kind of naming comments, unused and copied variables can be highlighted. If we started digging into this here it could take a while.

PR #876, opened 2025-06-02
Add a switch for making all nHCal layers sensitive

lkosarz:

Added a switch to make all layers sensitive for sampling fraction calculation. Use *sampF.xml geometry for that. Needed for sampling fraction benchmark.

- [ ] Bug fix (issue #__)
- [x] New feature: Add a switch to make all layers sensitive for sampling fraction calculation
- [ ] Documentation update
- [ ] Other: __


src/PolyhedraEndcapCalorimeter2_geo.cpp
```diff
@@ -75,7 +76,7 @@ static Ref_t create_detector(Detector& description, xml_h e, SensitiveDetector s
       sliceZ += s_thick / 2;
       PlacedVolume s_phv = l_vol.placeVolume(s_vol, Position(0, 0, sliceZ));
       s_phv.addPhysVolID("slice", s_num);
-      if (x_slice.isSensitive()) {
+      if (x_slice.isSensitive() || allSensitive == 1) {
```
wdconinc, line 79: Avoid bool to int conversion.
```suggestion
      if (x_slice.isSensitive() || allSensitive) {
```

lkosarz:
I'm not sure why some checks failed. Looks like container failed to setup, which is not connected to the updates. I'll try to rerun.

veprbl:
I think we need to review/merge https://github.com/eic/epic/pull/885 first.

lkosarz:
I originally prepared it in such a way, that the steel was updated - to be merged after #885, but I see the extra changes are removed. This looks good to me.

veprbl:
> I originally prepared it in such a way, that the steel was updated - to be merged after #885, but I see the extra changes are removed. This looks good to me.

We still to wait for #885 as there are merge conflicts between the two that will will have to resolve.

veprbl (dismissed):
Diff LGTM

compact/hcal/backward.xml
wdconinc, line 1: This file misses copyright header.

Issue #884, opened 2025-06-04
StainlessSteel density fix based on SAE 304


lkosarz:
Update density of StainlessSteel from 8.3 to 7.9 g/cm^3 of SAE 304. The density is wrong.

PR #885, opened 2025-06-04
Update materials.xml

lkosarz:
Fix density of StainlessSteel from 8.3 to 7.9 g/cm^3

- [ x] Bug fix (issue #884)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### DIFF
--- a/compact/materials.xml
+++ b/compact/materials.xml
@@ -412,7 +412,7 @@
     <composite n="2" ref="O"/>
   </material>
   <material name="StainlessSteel">
-    <D type="density" value="8.3" unit="g / cm3"/>
+    <D type="density" value="7.9" unit="g / cm3"/>
     <fraction n="0.74" ref="Fe"/>
     <fraction n="0.18" ref="Cr"/>
     <fraction n="0.08" ref="Ni"/>


lkosarz:
Yes, this is exactly an attempt to do it, by opening an issue. I can post it on Mattermost etc. Initially I wanted to fix it for our subsystem only, but I'm pretty sure, the density is not correct.

nat93 (dismissed):
The real central beam pipe SS is still undefined. It might be P506. 
SAE304 should work for now.

simonge (dismissed):

ajentsch:
Hey guys,

I have been on travel dealing with family stuff so I haven't been checking
email as much.

My number was not incorrect - it was for standard, construction grade 80/20
steel used for beam pipes and support structure.

I'm fine with labeling the exact Stainless steel used, but please do not
remove the original one.

wdconinc:
> it was for standard, construction grade 80/20 steel used for support structure

@ajentsch Can you clarify this at some point? 80/20 is mostly used to describe extruded _aluminum_ profiles, not typically a steel grade. Typical vacuum beam pipes and conflat flanges would be 304 or 316 stainless (e.g. https://www.mcmaster.com/products/conflat-flanges/)

ajentsch:
Typically yes, but there is an 80/20 stainless alloy that had a higher
density of 8.3 g/cm3. When I was finding the correct density number a few
years ago I was told to just use one type of steel for everything since we
didn't have details of connections, flanges etc. worked out. When I asked
what kind I should used (I'm not an expert in beam pipe design) I was sent
a link with some specs. I'd have to dig it up because it was a while ago.

But as I said before, Andrii will know the correct numbers, or can check
with Charlie.

I just want to make sure the right numbers are used for the pipes vs.
absorber since I assume they will be different.

veprbl:
Original Leszek's point was that the material composition of "StainlessSteel" does not correspond to any existing material with a 8.3 g/cm^3 density. I've looked up 80/20 alloy and its showing me a completely different material http://www.metalcor.de/en/datenblatt/114/ of nickel-chromium, which makes me conclude that the current "StainlessSteel" is not consistent.

ajentsch:
SAE304 is a good option. If we need to add other steel variations we can do
so as the need arises.

veprbl:
Required CI hasn't finished yet, I restarted the job manually. We can merge tomorrow unless objections arise.

PR #886, opened 2025-06-04
Automate beam energy configurations

simonge:
Aiming to address https://github.com/eic/EICrecon/issues/1899 by automatically adding xml configurations for all beam energy combinations to detector setups which include far_forward or far_backward subsystems.

Also changes the default energy from 18x275 to 5x41 when no values are provided.

Some of the yml files probably don't need the far detector subsystems for the studies they are used for so could be removed.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No
Adds a lot more xml files to confuse the users.


wdconinc:
We run some ci for every configuration and I think it may be based on the files in configurations/. Does that still work and does it blow up the CI load?

PR #890, opened 2025-06-18
Add parameters needed for 6.3 lattice and magnet visualization.

simonge:
Moves the code and visualization changes from https://github.com/eic/epic/pull/838/ so that it will reflect the minimal differences required to update the lattice.

Mostly this adds parameters which are set to 0 in the current configuration but need to be set in the update.

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No
No

### DIFF
--- a/src/CylindricalMagnetChain_geo.cpp
+++ b/src/CylindricalMagnetChain_geo.cpp
@@ -33,6 +33,7 @@ static Ref_t create_magnet(Detector& description, xml_h e, SensitiveDetector /* 
     xml_comp_t magnet(magnet_coll);
 
     string name   = getAttrOrDefault<string>(magnet, _Unicode(name), "");
+    int id        = getAttrOrDefault<int>(magnet, _Unicode(id), 0);
     double x      = getAttrOrDefault<double>(magnet, _Unicode(x), 0);
     double y      = getAttrOrDefault<double>(magnet, _Unicode(y), 0);
     double z      = getAttrOrDefault<double>(magnet, _Unicode(z), 0);
@@ -45,9 +46,14 @@ static Ref_t create_magnet(Detector& description, xml_h e, SensitiveDetector /* 
     Tube yoke_tube(rin, rout, 0.5 * length);
     Volume v_yoke("v_yoke_" + name, yoke_tube, m_Iron);
 
-    v_yoke.setVisAttributes(description.visAttributes(vis_name));
+    v_yoke.setVisAttributes(x_det.visStr());
+    
+    auto yoke_pv = assembly.placeVolume(v_yoke, Transform3D(RotationY(theta), Position(x, y, z)));
 
-    assembly.placeVolume(v_yoke, Transform3D(RotationY(theta), Position(x, y, z)));
+    yoke_pv.addPhysVolID("element", id);
+    DetElement yoke_de(sdet, name, id);
+    yoke_de.setPlacement(yoke_pv);
+    yoke_de.setAttributes(description, v_yoke, x_det.regionStr(), x_det.limitsStr(), vis_name);
   }
 
   // Final placement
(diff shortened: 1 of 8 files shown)


wdconinc (approved):
Looks good to me. Have not done a one to one comparison with capybara.

Question: what is the element id for the yoke for? Do you intend to make that sensitive for some reason?

compact/far_backward/definitions.xml
```diff
@@ -28,7 +28,9 @@
 
     <constant name="B2BeR_InnerRadius"    value="111.0*mm"/>
     <constant name="B2BeR_Length"         value="3.18*m"/>
+    <constant name="B2BeR_CenterX"       value="0*mm"/>
     <constant name="B2BeR_CenterPosition" value="-13.275*m"/>
```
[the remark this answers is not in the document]
simonge: Good point, I've tried to make the naming more consistent throughout.

simonge:
> Question: what is the element id for the yoke for? Do you intend to make that sensitive for some reason?

It was a while ago I added that so not entirely sure. I believe the visualization wasn't working correctly without a detector element and the detector element didn't work without an id. Some/all of what has been added might be unnecessary.

simonge:
> Looks good to me. Have not done a one to one comparison with capybara.

What was the problem with publishing the capybara here as is done in EICRecon? I didn't follow why it was added then reverted.

compact/far_backward/definitions.xml
```diff
@@ -174,7 +176,7 @@
     <constant name="Vacuum_BB_MaxX"       value="0.5*m"/>
     <constant name="Vacuum_BB_MinY"       value="-1.5*m"/>
     <constant name="Vacuum_BB_MaxY"       value="1.5*m"/>
-    <constant name="Vacuum_BB_MinZ"       value="B2BeR_CenterPosition-B2BeR_Length/2"/>
+    <constant name="Vacuum_BB_MinZ"       value="B2BeR_CenterZ-       B2BeR_Length/2"/>
```
wdconinc, line 179:
```suggestion
    <constant name="Vacuum_BB_MinZ"       value="B2BeR_CenterZ-B2BeR_Length/2"/>
```

src/CylindricalMagnetChain_geo.cpp
```diff
@@ -45,9 +46,14 @@ static Ref_t create_magnet(Detector& description, xml_h e, SensitiveDetector /*
...
-    v_yoke.setVisAttributes(description.visAttributes(vis_name));
+    v_yoke.setVisAttributes(x_det.visStr());
 
-    assembly.placeVolume(v_yoke, Transform3D(RotationY(theta), Position(x, y, z)));
+    auto yoke_pv = assembly.placeVolume(v_yoke, Transform3D(RotationY(theta), Position(x, y, z)));
+
+    yoke_pv.addPhysVolID("element", id);
+    DetElement yoke_de(sdet, name, id);
+    yoke_de.setPlacement(yoke_pv);
+    yoke_de.setAttributes(description, v_yoke, x_det.regionStr(), x_det.limitsStr(), vis_name);
```
wdconinc, line 56: I'd still like to avoid this, if we can. We only have 255 detector IDs in our 8-bit field for segmentations, and less than half are left. If we start giving IDs to all magnets, we're going to run out quickly.

Note that you can add regions and limits to volumes too:
```
DDCore/include/DD4hep/Volumes.h:
    const Volume& setAttributes(const Detector& description, const std::string& region, const std::string& limits,
                                const std::string& vis) const;
```

Can you verify that visualization is not affected when avoiding the ID for this?

simonge: I have removed the DetElement and recovered the visualization using setAttributes. The physical volume ID is still required to give the volume a unique identifier in the tree but shouldn't count towards the detector IDs

PR #894, opened 2025-06-19
fix: BarrelExtraSpace_thickness zeroed; DIRC support rails modified to avoid overlap

wdconinc:
This PR removes 1.5 cm of extra space between the DIRC and the BIC. This places the rmin of the BIC at 82.5 cm instead of 84 cm. This is closer to how it should be, since in CAD it should be 82 cm. Despite the remaining discrepancy, this should be merged.

This PR also hard-codes the carbon fiber inner detector support cylinder radius in the definitions, so it can be used.

This PR also slightly modifies how the DIRC/MPGD support rails are encoded. They now touch the inner detector support cylinder and the BIC rmin.

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No. This extra space only affects the EcalBarrel_rmin, and does not propagate to other subsystems.

Yes, BIC smaller.

### DIFF
--- a/src/DIRC_geo.cpp
+++ b/src/DIRC_geo.cpp
@@ -250,10 +250,11 @@ static Ref_t createDetector(Detector& desc, xml_h e, SensitiveDetector sens) {
   dirc_support.setVisAttributes(desc.visAttributes(xml_support.visStr()));
 
   // Rail
-  xml_comp_t xml_rail       = xml_support.child(_Unicode(rail));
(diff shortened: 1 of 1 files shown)


mariakzurek (dismissed):

wdconinc:
No more overlaps. This is ready for review and merge.

sly2j (approved):

PR #896, opened 2025-06-24
feat: tighten up the world to a more minimal geometry

wdconinc:
This PR shrinks the world volume from 60 m x 60 m x 200 m to something that is just sufficient to enclose the relevant volumes. This way we won't be tracking slow neutrons on their way to death in air.

- [x] Bug fix (issue: the world is too big)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

- [x] Changes have been communicated to collaborators: @ajentsch @simonge this makes the forward and backwards regions quite narrow, chosen based on what's currently in the geometry

No.

No.

### DIFF
--- a/compact/definitions.xml
+++ b/compact/definitions.xml
@@ -9,6 +9,7 @@
     <documentation>
       ## World volume
     </documentation>
+    <constant name="backward_x0" value="0.5*m"/>
     <constant name="backward_dx" value="2*m"/>
     <constant name="backward_dy" value="1*m"/>
     <constant name="backward_dz" value="35*m"/>
--- a/compact/definitions.xml
+++ b/compact/definitions.xml
@@ -10,14 +10,14 @@
       ## World volume
     </documentation>
     <constant name="backward_dx" value="2*m"/>
-    <constant name="backward_dy" value="0.5*m"/>
-    <constant name="backward_dz" value="65*m"/>
+    <constant name="backward_dy" value="1*m"/>
+    <constant name="backward_dz" value="35*m"/>
     <constant name="exp_hall_dx" value="3.5*m"/>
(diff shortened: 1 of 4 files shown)


wdconinc:
Hmm, "Unknown Shape TGeoCompositeShape" in convert-to-step...

wdconinc:
There was a hidden vacuum volume that is now also fixed, so finally overlaps succeed:
image

wdconinc:
For reference, this is the world before, which is clearly more than necessary (and it's all air, not vacuum):
image

simonge:
Looks very useful, do you have any metrics which show the performance improvements?

It would be nice to be able to control the world volume more precisely for different configurations. Could the parameters be moved into a separate e.g. worlddefinition_default.xml file.

simonge:
> I don't think the speed-up is above the percent level, though, so I don't think there is too much benefit in tweaking this much. We could make it a compact file that is listed in the yml configs, but I would also want to be mindful of the potential for errors there when stuff extends outside the world.

I would expected more significant gains in the tracking only geometries where there aren't many secondaries produced in the detector but lots might come from the air. When I was looking at this for the specific case of beamline tracking, killing the single electron track in the beampipe vacuum rather than allowing it to continue through 50m of air resulted in a 200x speed up. It won't be quite the same for this but might still be significant.

wdconinc:
Good point! Maybe sharpening up the world for those is useful then.

veprbl (dismissed):
Looks good to me.

wdconinc:
> Looks good to me.

This needs npsim v1.4.5.

veprbl (approved):

veprbl (approved):

PR #900, opened 2025-06-27
feat: add DIRC to Acts geometry for truth hits and projection surfaces

wdconinc:
The DIRC group needs track projections to the DIRC, a way to assess the quality of these track projections, and truth information to generate lookup tables.

This PR adds Acts support to the DIRC. This will ensure consistency between truth and reconstruction surfaces, and the material budgets around the DIRC (inner/outer). The sim hits on the DIRC will be generated for non-optical photons only, and will not be used in the track reconstruction (naturally).

- [ ] Bug fix (issue #__)
- [x] New feature (issue: DIRC needs the track position at the quartz bars)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -150,7 +150,7 @@ jobs:
     needs: build
     uses: ./.github/workflows/check-tracking-geometry.yml
     with:
-      detector_configs: "['epic_craterlake', 'epic_ip6', 'epic_dirc_only']"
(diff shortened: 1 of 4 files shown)


ShujieL:
can we attach an ACTS measurement surface to the geometry like [this](https://github.com/eic/epic/blob/27bf0215a74a3ca5441f3bad638b79bfd9b59aa2/src/TrapEndcapTracker_geo.cpp#L180) ? 
As long as we don't put this in the hit collection for CKF it should not interfere with the tracking performance.

PR #893, opened 2025-06-19
New far forward beampipe and cryostat model v2025 implementation

nat93:
### Briefly, what does this PR introduce?
This PR introduces a complete model of the cryostat (fwd and bwd) with detailed description of superconducting coils, support tubes, yokes, yoke shielding, heat shielding, and cryostat vessel. The fwd and rear sides were adjusted in order to avoid overlaps with off-momentum detectors and low-Q^2 tagger pipes.

The corresponding STEP file can be found on SharePoint: EIC Public Sharing Docs → Documents → Experimental Program → ePIC → Engineering → STR-Files → IR6_CRYOSTAT_2200m_top_4-24-2025.stp [https://brookhavenlab.sharepoint.com/:u:/s/EICPublicSharingDocs/EdR44ODny1BEmMTTks61CCwBe8tJYMc0iWkPIYY_EjYEuw?e=2XLUEt]

The changes have been communicated to TIC, IR, and MDP communities.

### What kind of change does this PR introduce?
- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [x] Changes have been communicated to collaborators

### Does this PR change default behavior?
Yes, it adds extra materials between the beam pipe (beam loss hot-spots) and the ePIC detector. It is crucial for the realistic detector background and dose calculations.

PR #903, opened 2025-07-03
Fixes for stuck particles

### LINKED ISSUE #902: Navigation issue in far-backward
beamline benchmark https://eicweb.phy.anl.gov/EIC/benchmarks/detector_benchmarks/-/jobs/5792714#L7870 revealed a potential issue:
```
Stuck Track: potential geometry or navigation problem.
  Track stuck, not moving for 10 steps.
  Current  phys volume: 'AV_65!Pipe_Q1APR_to_B1PR_assembly_44#44!v_bend_vac_Pipe_in_Q1APR_5#5'
   - at position : (164.6801976241447,-2.120532151842422,-7098.101365046259)
     in direction: (0.03451755227922471,-0.003668044780671342,-0.9993973604288425)
    (local position: (33.38031232104109,9.094947017729282e-13,2.120532151842422))
    (local direction: (-0.009524434611176597,0.9999479139399334,0.003668044780671281)).
  Previous phys volume: 'world_volume_1'
  Likely geometry overlap - else navigation problem !
   *** Trying to get *unstuck* using a push - expanding step to 1e-07 (mm) ...       Potential overlap in geometry !
```
This is severe enough to cause a stuck simulation.

simonge:
Appears to fix #902.

Tiny torus segments were being created where the alignment of one tube didn't match the next due to floating point rounding errors. Below a limit of 0.01 mrad difference now is rounded down to 0.

Length of `Pipe_to_Q1APR` has also been fixed so it correctly joins with the central beampipe.

Most of this may be reworked again after #893 and #838

- [x] Bug fix (issue #902 )
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### DIFF
--- a/src/BeamPipeChain_geo.cpp
+++ b/src/BeamPipeChain_geo.cpp
@@ -97,9 +97,6 @@ static Ref_t create_detector(Detector& description, xml_h e, SensitiveDetector /
   for (uint i = 1; i < thetas.size(); i++) {
 
 
-    std::cout << "Pipe " << i<< ": " << names[i] << std::endl;
-    std::cout << "theta: " << thetas[i-1] << std::endl;
-    std::cout << "theta: " << thetas[i] << std::endl;
     // Start at the join between the first two pipes ending at the join between the last two pipes N-1
     double bendAngle  = thetas[i] - thetas[i - 1];
     if (std::abs(bendAngle) < 0.00001) {
@@ -108,8 +105,6 @@ static Ref_t create_detector(Detector& description, xml_h e, SensitiveDetector /
     {
       double bendLength = abs(rOuters1[i] * tan(bendAngle / 2));
       bendLengths.push_back(bendLength + bendRadius);
-      std::cout << "bendAngle: " << bendAngle << std::endl;
-      std::cout << "bendLength: " << bendLength << std::endl;
     }
   }
 
--- a/src/BeamPipeChain_geo.cpp
+++ b/src/BeamPipeChain_geo.cpp
@@ -84,6 +84,7 @@ static Ref_t create_detector(Detector& description, xml_h e, SensitiveDetector /
     double l      = sqrt(deltaX * deltaX + deltaZ * deltaZ);
     double theta  = atan2(deltaX, deltaZ);
 
+
     xCenters[pipeN] = x;
     zCenters[pipeN] = z;
     lengths[pipeN]  = l;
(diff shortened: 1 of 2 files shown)


src/BeamPipeChain_geo.cpp
```diff
@@ -94,12 +94,13 @@ static Ref_t create_detector(Detector& description, xml_h e, SensitiveDetector /
 
   // If there is an bend in the pipe, calculate the length reduction of the pipe and joint length
   for (uint i = 1; i < thetas.size(); i++) {
+
     // Start at the join between the first two pipes ending at the join between the last two pipes N-1
-    if (thetas[i - 1] == thetas[i]) {
+    double bendAngle = thetas[i] - thetas[i - 1];
+    if (std::abs(bendAngle) < 0.00001) {
```
wdconinc, line 100: (Optional) Use explicit units, maybe in mrad.

wdconinc:
Is there a way we can catch these overlaps without waiting until someone sees a warning in a Geant4 simulation? Smaller tolerance on the tgeo-based overlap check?

simonge:
> Is there a way we can catch these overlaps without waiting until someone sees a warning in a Geant4 simulation? Smaller tolerance on the tgeo-based overlap check?

I believe this snuck through because currently the effected beampipes are only in the extended geometry which isn't tested automatically.

simonge:
Getting back to this as I'm looking to finish the benchmarks which highlighted the issue. Looks like there were no navigation errors in the beamline benchmark but there are still some overlaps in the new tests added for the ip6_extended configuration and smaller tolerances.
https://eicweb.phy.anl.gov/EIC/benchmarks/detector_benchmarks/-/jobs/6428140

simonge:
The overlaps relevant to the far-backward extension have been fixed by expanding the inner yoke of the placeholder cylindrical magnets by 1%. 

The remaining overlaps from the 0.01mm configuration were unrelated to the issue or the far-backward region so I have reverted the check to 0.1mm. If the geant4 overlap gets changed to 0.01mm in the future as the root one has, these will obviously need to be addressed.

I think this is ready for review and merging.

simonge:
Is this stuck because there are isn't a `check-overlap-geant4 (epic_craterlake)` job anymore? It is now split into `check-overlap-geant4 (epic_craterlake, 0.1)` and `check-overlap-geant4 (epic_ip6_extended, 0.1)`. I can't see how/where the requirement for a the specific job is being set.

wdconinc:
That's set in the repository settings. @veprbl are we ready to merge this?

PR #904, opened 2025-07-05
test: add epic_ip6_extended to check-overlap-tgeo

veprbl:
This is a test for https://github.com/eic/epic/pull/903#issuecomment-3034675600

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No
No

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -293,7 +293,7 @@ jobs:
     strategy:
       matrix:
         option: ['m'] #, 's'] # FIXME
-        detector_config: [epic_craterlake] # FIXME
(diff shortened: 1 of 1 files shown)


simonge:
> @simonge I don't see an issue in the ip6_extended configuration. By any chance, did you have a better test that we could implement? Or, maybe, we need to do geant4 scan from e.g. different origin?

Just checked and my local overlap check was running with a limit of 0.01 mm rather than 0.1 mm. I wouldn't recommend doing this for the full geometry at the moment. Although the overlaps between the main and my branch which should fix the stuck tracks doesn't seem to report any different detector elements, this might just be that these torus slithers are even smaller than the overlap checks at all.

PR #906, opened 2025-07-11
Update material map

veprbl:


veprbl:
Thanks for spotting the B0. That one I'm not sure about.

The BIC issue is trivial. After current CI finishes running validation-map, We should apply:
```diff
diff --git a/compact/ecal/bic/bic_layer1_only.xml b/compact/ecal/bic/bic_layer1_only.xml
index b97dd14fc..ef459ea54 100644
--- a/compact/ecal/bic/bic_layer1_only.xml
+++ b/compact/ecal/bic/bic_layer1_only.xml
@@ -223,4 +223,11 @@
     </readout>
   </readouts>
 
+  <plugins>
+    <plugin name="DD4hep_ParametersPlugin">
+      <argument value="EcalBarrelImaging"/>
+      <argument value="layer_pattern: str=envelope"/>
+    </plugin>
+  </plugins>
+
 </lccdd>
```

veprbl:
Looks like "missing" B0 is just a drawing artifact with the current scripts.

Strange how it could not process 10000 events and just hang. The reason for increase was that I'm noticing how my statistics are not as high as in previous plots.

veprbl:
Some distributions for material around the BIC layer from the CI artifacts
image
image
image
image

ShujieL (approved):

PR #917, opened 2025-08-11
fix: tighter overlap tolerances (10 um)

wdconinc:
This PR tightens the overlap checks from 100 um to 10 um (for cause: 100 um allows silicon tracker overlaps to sneak through).

- [x] Bug fix (issue: silicon tracker overlaps sneak through)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -309,7 +309,7 @@ jobs:
         setup: install/bin/thisepic.sh
         run: |
           mkdir -p doc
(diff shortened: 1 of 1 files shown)


simonge:
Is the plan here to enforce <10 um overlaps or is this branch just being used to keep track of where they are?

wdconinc:
RFR @veprbl Let's just merge this. It's probably overkill but it also doesn't cost us anything when applied to the ROOT TGeo overlap checks. We do use O(10um) layers so this would catch issues of that magnitude.

PR #925, opened 2025-08-22
Add support materials to FTOF

ssedd1123:

There is currently nothing between the front and back layers of FTOF. However, dRICH group wants to study the effect of FTOF material and having the right material is important for their study. 

A uniform slab of carbon foam of width 1 cm is now sandwiched between the two layers. Details such as cooling tubes will be added later. The width of the support structure is given by slide 6 of https://indico.bnl.gov/event/24949/attachments/57520/98759/ePICTOF_WP2_FTOFLayout_09272024v2.pdf

The thickness of the module board and PCB are not final and will be changed in the future. 

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No

No

### DIFF
--- a/compact/tracking/tof_endcap.xml
+++ b/compact/tracking/tof_endcap.xml
@@ -121,12 +121,12 @@
     <constant name="EndcapTOF_HybridTrace_thickness"  value="0.002500*cm"/>
     <constant name="EndcapTOF_HybridPCB_thickness"    value="0.1*cm"/>
     <constant name="EndcapTOF_CFSkin_thickness"       value="2*0.0075*cm"/>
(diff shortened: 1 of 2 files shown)


wdconinc:
> Rebasing just to trigger check-tracking-geometry, but the failure in https://github.com/eic/epic/actions/runs/17691331481/job/50285341117?pr=925 appears caused by this branch since main passes that check...

Looking at the issue, it's likely identical to what caused the CI checks in ACTS PR#4620 to fail for ODD. Let me update our patch in eic-shell.

veprbl (approved):
Looks reasonable.

PR #930, opened 2025-08-26
MPGD Geo Validation

mposik1983:
Check that MPGD simulation geometry matches CAD drawing ([CAD File](https://indico.bnl.gov/event/27186/contributions/104526/attachments/60501/104351/EPIC%20Active%20Area%20Envelope%20-%2003-20-2025.pdf))

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [x] Other: __

Doesn't break anything, but geometry is changed and material map will need to be reproduced.
MPGD hit positions will be modified due to geometry changes

### DIFF
--- a/compact/tracking/mpgd_barrel.xml
+++ b/compact/tracking/mpgd_barrel.xml
@@ -52,9 +52,10 @@
     <constant name="MMCentral_gap"                          value="0.2*cm"/>
 
     <comment> In principle, stave count should not be independenty specified: it is fixed by above intrinsic parameters </comment>
+    <comment> 1.3cm is to make envelope match CAD drawings </comment>
     <constant name="MPGDBarrelStave_count"                  value="8"/>
     <constant name="InnerMPGDBarrelLayer_rmin"              value="InnerMPGDBarrel_rmin"/>
-    <constant name="InnerMPGDBarrelLayer_rmax"              value="MMOuterSector_R + MMRadial_offset / 2 + 1*cm"/>
+    <constant name="InnerMPGDBarrelLayer_rmax"              value="MMOuterSector_R + MMRadial_offset / 2 + 1*cm +1.3*cm"/>
     <constant name="rmin2Sensor" value="MMDriftCuGround_thickness + MMDriftKapton_thickness + MMDriftCuElectrode_thickness + MMGasGap_thickness / 2" />
     <constant name="MMInnerSensor_R" value="MMInnerSector_R + rmin2Sensor" />
     <constant name="MMOuterSensor_R" value="MMOuterSector_R + rmin2Sensor" />
--- a/compact/tracking/mpgd_barrel.xml
+++ b/compact/tracking/mpgd_barrel.xml
@@ -52,7 +52,7 @@
     <constant name="MMCentral_gap"                          value="0.2*cm"/>
 
     <comment> In principle, stave count should not be independenty specified: it is fixed by above intrinsic parameters </comment>
-    <comment> 1.3cm is to make envelope match CAD drawings </comment>
+    <comment> Factor of 1.3cm on InnerMPGDBarrelLayer_rmax is to extend radial envelope to match 03-12-2025 CAD envelope </comment>
     <constant name="MPGDBarrelStave_count"                  value="8"/>
     <constant name="InnerMPGDBarrelLayer_rmin"              value="InnerMPGDBarrel_rmin"/>
     <constant name="InnerMPGDBarrelLayer_rmax"              value="MMOuterSector_R + MMRadial_offset / 2 + 1*cm +1.3*cm"/>
--- a/compact/tracking/mpgd_barrel.xml
+++ b/compact/tracking/mpgd_barrel.xml
@@ -52,10 +52,7 @@
     <constant name="MMCentral_gap"                          value="0.2*cm"/>
 
     <comment> In principle, stave count should not be independenty specified: it is fixed by above intrinsic parameters </comment>
-    <comment> Factor of 1.3cm on InnerMPGDBarrelLayer_rmax is to extend radial envelope to match 03-12-2025 CAD envelope </comment>
     <constant name="MPGDBarrelStave_count"                  value="8"/>
-    <constant name="InnerMPGDBarrelLayer_rmin"              value="InnerMPGDBarrel_rmin"/>
-    <constant name="InnerMPGDBarrelLayer_rmax"              value="MMOuterSector_R + MMRadial_offset / 2 + 1*cm +1.3*cm"/>
     <constant name="rmin2Sensor" value="MMDriftCuGround_thickness + MMDriftKapton_thickness + MMDriftCuElectrode_thickness + MMGasGap_thickness / 2" />
     <constant name="MMInnerSensor_R" value="MMInnerSector_R + rmin2Sensor" />
     <constant name="MMOuterSensor_R" value="MMOuterSector_R + rmin2Sensor" />
@@ -101,8 +98,8 @@
 
       <layer module="InnerMPGDBarrel_Mod1" id="1" vis="InvisibleWithDaughters">
         <barrel_envelope
-          inner_r="InnerMPGDBarrelLayer_rmin"
-          outer_r="InnerMPGDBarrelLayer_rmax"
+          inner_r="InnerMPGDBarrel_rmin"
+          outer_r="InnerMPGDBarrel_rmax"
           z_length="InnerMPGDBarrelLayer_length"
           z0="InnerMPGDBarrel_zoffset"/>
         <layer_material surface="inner" binning="binPhi,binZ" bins0="MPGDBarrelStave_count*24" bins1="100" />
--- a/compact/tracking/mpgd_outerbarrel.xml
+++ b/compact/tracking/mpgd_outerbarrel.xml
@@ -46,7 +46,7 @@
    <detector id="TrackerBarrel_5_ID" name="MPGDOuterBarrel" type="epic_OuterMPGDBarrel" readout="OuterMPGDBarrelHits" vis="TrackerVis">
      <type_flags type="DetType_TRACKER + DetType_BARREL"/>
      <dimensions width="MPGDOuterBarrelModule_width" length="MPGDOuterBarrelModule_length" height="MPGDOuterBarrelModule_allowed_space" />
-          <position x="0" y="0" z="MPGDOuterBarrelModule_offset" />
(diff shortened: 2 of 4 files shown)


compact/tracking/mpgd_outerbarrel.xml
```diff
@@ -45,9 +46,10 @@
    <detector id="TrackerBarrel_5_ID" name="MPGDOuterBarrel" type="epic_OuterMPGDBarrel" readout="OuterMPGDBarrelHits" vis="TrackerVis">
      <type_flags type="DetType_TRACKER + DetType_BARREL"/>
      <dimensions width="MPGDOuterBarrelModule_width" length="MPGDOuterBarrelModule_length" height="MPGDOuterBarrelModule_allowed_space" />
-     <position x="0" y="0" z="MPGDOuterBarrelModule_offset" />
+          <position x="0" y="0" z="MPGDOuterBarrelModule_offset" />
```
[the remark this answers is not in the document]
mposik1983: No, its not. Should be fixed now.

rahmans1:
@mposik1983 It seems like the tracking geometry check is complaining about some overlaps. Do you see it locally as well? https://github.com/eic/epic/actions/runs/17614403464/job/50043978042?pr=930

rahmans1:
> @mposik1983 It seems like the tracking geometry check is complaining about some overlaps. Do you see it locally as well? https://github.com/eic/epic/actions/runs/17614403464/job/50043978042?pr=930

apparently should be fixed by an acts update which is not on cvmfs yet. I checked on my local nightly and it seemed fine. One thing that remains now is the material map check.

mposik1983:
Looking through my local npsim and eicrecon logs (what is printed by default), I did not see any overlap errors related to ACTS. The output that I got from eicrecon with the new mpgd geo seems reasonable (plot attached, generated using the previous acts material map). I remember hearing of an ACTS geo issue, but do not know the details. I have a hard time parsing through and making sense of the acts log outputs, can you tell what detector(s) are causing the overlaps?
Screenshot 2025-09-10 at 9 54 13 AM

compact/tracking/mpgd_barrel.xml
```diff
@@ -52,9 +52,10 @@
     <constant name="MMCentral_gap"                          value="0.2*cm"/>
 
     <comment> In principle, stave count should not be independenty specified: it is fixed by above intrinsic parameters </comment>
+    <comment> 1.3cm is to make envelope match CAD drawings </comment>
     <constant name="MPGDBarrelStave_count"                  value="8"/>
     <constant name="InnerMPGDBarrelLayer_rmin"              value="InnerMPGDBarrel_rmin"/>
-    <constant name="InnerMPGDBarrelLayer_rmax"              value="MMOuterSector_R + MMRadial_offset / 2 + 1*cm"/>
+    <constant name="InnerMPGDBarrelLayer_rmax"              value="MMOuterSector_R + MMRadial_offset / 2 + 1*cm +1.3*cm"/>
```
wdconinc, line 58: Don't like the magic. Is this a placement different from what the CAD specifies?

compact/tracking/mpgd_outerbarrel.xml
```diff
@@ -13,20 +13,21 @@
...
-    <constant name="MPGDOuterBarrelModule_allowed_space"           value="2.5*cm"/>
+    <constant name="MPGDOuterBarrelModule_allowed_space"           value="25*mm"/>
     <constant name="MPGDOuterBarrelModule_rmax"                    value="MPGDOuterBarrelModule_rmin + MPGDOuterBarrelModule_allowed_space" />
     <constant name="MPGDOuterBarrelModule_roffset"                 value="0.5*cm" />
-    <constant name="MPGDOuterBarrelModule_width"                   value="36.0*cm"/>
+    <constant name="MPGDOuterBarrelModule_width"                   value="360*mm"/>
     <constant name="MPGDOuterBarrel_length"                        value="MPGDOuterBarrelModule_zmin1 + MPGDOuterBarrelModule_zmin2"/>
     <constant name="MPGDOuterBarrelModule_length"                  value="0.5*(MPGDOuterBarrelModule_zmin1 + MPGDOuterBarrelModule_zmin2 + MPGDOuterBarrelModule_zoverlap)"/>
-    <constant name="MPGDOuterBarrelModule_offset"                  value="(MPGDOuterBarrelModule_zmin2 - MPGDOuterBarrelModule_zmin1)/2.0"/>
+    <constant name="MPGDOuterBarrelModule_offset"                  value="(MPGDOuterBarrelModule_zmin2 - MPGDOuterBarrelModule_zmin1)"/>
```
wdconinc, line 29: This is not how offset is usually defined. Why no div by 2? Normally it's the half the difference since that is where the center of the volume is compared to the origin. Can you add a comment here?

```
    <constant name="MPGDOuterBarrelModule_zmin1"     value="1795*mm"/>
    <constant name="MPGDOuterBarrelModule_zmin2"     value="1845*mm"/>
```
This means that the barrel extends from `-1795` to `+1845`. That means the center is at `+25`, not at `+50`.

compact/tracking/mpgd_outerbarrel.xml
```diff
@@ -13,20 +13,21 @@
...
+    <constant name="MPGDOuterBarrelModule_feb_ext"                 value="110*mm"/>
```
wdconinc, line 30: This is used as an offset for the PCB below. The name is unclear, and it may be better to keep calling this `offset`, e.g.

```suggestion
    <constant name="MPGDOuterBarrelModule_PCB_offset"                 value="110*mm"/>
```

wdconinc:
Rebased to trigger new set of workflows with the (finally) updated eic_xl on cvmfs.

compact/tracking/mpgd_barrel.xml
ybedfer: Hello Matt,
And sorry for the late reply...
The modules should fit into the envelope. This is the only constraint, in my understanding. Right?
And this constraint is fulfilled...
...As to the the inner surface of the envelope, by:
 - setting its radius = 550mm, as in the CAD. 
 - setting the minimum radius of the inner sectors ("sector" being the subdivision along Z and the 
 inner sectors being the smallest ones) large enough that  when it's combined w/ a negative radial 
 offset , it still fits.
  From the histogram of the R  distribution, one can see that it's indeed the case: everything sits above 550.
...As to the outer surface, the opposite strategy (so to say) is chosen:
 - the outer radius of the envelope is set from the parameters of the outer sectors: radius + positive
  radial offset plus a 1cm margin accounting for the thickness of the modules.
  Since the modules have at present a total thickness of 0.4180 cm (which one learns by summing the
 the thicknesses of all the components, or, simpler, by enabling the debugging option of the CyMBaL
 source code, which does the calculation for you), the 1cm margin is large enough.
I agree that the comments in the xml file are too sibylline for the reader to grasp all this.
They (the comments, that is) could be improved.
But, as I understand, the whole thing is about to be changed completely, w/ the new foreseen scheme of 12 sectors ('sector' being, this time, to be understood as a subdivision along phi) for all of
the barrel detectors.

src/BarrelPlanarMPGDTracker_geo.cpp
```diff
@@ -283,14 +291,20 @@ static Ref_t create_BarrelPlanarMPGDTracker_geo(Detector& description, xml_h e,
         string module_name = _toString(module, "module%d");
         DetElement mod_elt(lay_elt, module_name, module);
         double mod_z       = 0.5 * dimensions.length();
-        double z_placement = mod_z - j * nz * mod_z; // z location for module placement
+        double z_placement = mod_z - 0.5 * pcb_feb_ext -
+                             j * (nz * mod_z - pcb_feb_ext); // z location for module placement
```
veprbl, line 295: Even the original expression looks a bit fishy to me. I doubt it works for nz != 2. Maybe, it should have `2` instead of `nz`, because mod_z is half-length.

src/BarrelPlanarMPGDTracker_geo.cpp
```diff
@@ -159,7 +161,13 @@ static Ref_t create_BarrelPlanarMPGDTracker_geo(Detector& description, xml_h e,
...
       c_vol.setVisAttributes(description, x_comp.visStr());
 
-      pv = m_vol.placeVolume(c_vol, Position(0, 0, thickness_sum + x_comp.thickness() / 2.0));
+      if (comp_name == "PCB") {
+        pcb_feb_ext = x_comp.offset();
+        pv          = m_vol.placeVolume(
+            c_vol, Position(0, -pcb_feb_ext / 2.0, thickness_sum + x_comp.thickness() / 2.0));
+      } else {
+        pv = m_vol.placeVolume(c_vol, Position(0, 0, thickness_sum + x_comp.thickness() / 2.0));
+      }
```
veprbl, line 170: Can we just mark every non-PCB layer to have
```
offset="-MPGDOuterBarrelModule_PCB_offset / 2"
```
and for PCB do
```
offset="MPGDOuterBarrelModule_PCB_offset / 2"
```
Then you don't need to have separate cases here and extra math for `z_placement`.

src/BarrelPlanarMPGDTracker_geo.cpp
```diff
@@ -283,14 +291,20 @@ static Ref_t create_BarrelPlanarMPGDTracker_geo(Detector& description, xml_h e,
...
                 : z0 / 2.0; // determine the amount of overlap in z the z nz modules have
 
-        Transform3D tr(RotationZYX(0.0, ((M_PI / 2) - phic - phi_tilt), -M_PI / 2),
-                       Position(xc, yc, mpgd_pos.z() + z_placement + z_offset)); // in x-y plane,
+        Transform3D tr(
+            RotationZYX(0, ((M_PI / 2) - phic - phi_tilt), -M_PI / 2) * RotationZ(j * M_PI),
+            Position(
+                xc, yc,
+                mpgd_pos.z() + z_placement +
+                    z_offset)); //RotZYX rotat planes around azimuth, RotZ flip plane so pcb_feb_ext is faceing endcaps
```
veprbl, line 306:
```suggestion
        Transform3D tr(
            // RotationZYX rotates planes around azimuth, RotZ flips plane so pcb_feb_ext is facing endcaps
            RotationZYX(0, ((M_PI / 2) - phic - phi_tilt), -M_PI / 2) * RotationZ(j * M_PI),
            Position(
                xc, yc,
                mpgd_pos.z() + z_placement +
                    z_offset));
```
Needs spelling corrections.

PR #931, opened 2025-08-26
ci: Add reviewers to GitHub Actions Dependabot config

wdconinc:
This PR adds @eic/epic-detector-geometry-review as reviewer to dependabot updates.


veprbl (approved):
Reviewed so that we can keep reviewing going.

Issue #963, opened 2025-10-10
Update the FarBackward beamline, magnets and taggers


simonge:
The Low-Q2 Taggers and beamline geometry needs a significant overhaul. This issue intends to keep track of the various steps required and will probably see additional steps added along the way.

Shouldn't effect current detector setup - just prepares for changes
- [x] Separate the code defining the backwards taggers from the vacuum. #968
- [x] Move beamline definition into single file, separate from magnets. #968
- [x] Address the overlaps and stuck particles in the epic_ip6_extended configuration #903 
- [ ] Revisit minimal changes required for update after cryostats added #890
- [x] Add new magnet geometry code based on drawings (https://indico.bnl.gov/event/29346/) #978 
- [x] Adapt beamline tracking code to allow non-circular cross sections (antechamber) #989

Required for efficient changes to the geometry to be made and effects understood.
- [x] Retraining of neural network on benchmark failure https://github.com/eic/detector_benchmarks/pull/171
- [x] Update of onnx model attached to new PR for updated geometry https://github.com/eic/detector_benchmarks/pull/201

Updates required to geometry
- [ ] Exchange current drift vacuum with beampipe+antechamber based off CAD (https://indico.bnl.gov/event/29346/)
- [ ] Update beamline layout and magnetic fields. (https://indico.bnl.gov/event/29346/)

Further planned updates
- [ ] Split tagger readout to match currently proposed design
- [ ] Make layer materials more realistic
- [ ] Include 2D field maps for dipole magnets

PR #968, opened 2025-10-17
Separate tagger from beampipe

simonge:
Separates the definition of the Tagger detector from the beamline vacuum. The taggers are now inserted into the world volume rather than added into a section of the beampipe which is split between vacuum and air.

There are some very small changes to the positions of the detectors to match small edits to the beamline vacuum.

- [x] Bug fix (issue #963)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No
No

### DIFF
--- a/compact/far_backward/taggers.xml
+++ b/compact/far_backward/taggers.xml
@@ -15,8 +15,8 @@
 
       <comment> Tagger stations </comment>
       <module id="1" name="Tagger1">
-        <comment> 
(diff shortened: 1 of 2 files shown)


simonge:
> I believe it aligns with #989 PR. Right?

They are both essential for making the beamline changes but separate enough that they shouldn't need each other before merging. I think they are the final two small pieces before I can change the beamline while keeping everything else functioning and directly track the effect of the changes.

PR #979, opened 2025-11-07
nHCal update to 24 mm scintillator

lkosarz:
Update of nHCal geometry to 10x layers of 4 cm steel and 24 mm scintillator
Fixed potential overlap between nHCal and flux return - missed by overlap check, but may affect sampling fraction calculation

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,7 @@ repos:
     -   id: end-of-file-fixer
     -   id: trailing-whitespace
 -   repo: https://github.com/pre-commit/mirrors-clang-format
-    rev: v21.1.2
+    rev: v21.1.5
     hooks:
     -   id: clang-format
 -   repo: https://github.com/Lucas-C/pre-commit-hooks
--- a/compact/colors.xml
+++ b/compact/colors.xml
@@ -16,15 +16,11 @@
   <vis name="AnlBlue"         alpha="1" r="0/256"   g="96/256"  b="156/256"/>
   <vis name="AnlTeal"         alpha="1" r="0/256"   g="161/256" b="156/256"/>
   <vis name="AnlGray"         alpha="1" r="102/256" g="102/256" b="102/256"/>
-  <vis name="AnlLight_Gray"   alpha="1" r="209/256" g="209/256" b="209/256"/>
-  <vis name="AnlLightGray"    ref="AnlLight_Gray" />
+  <vis name="AnlLightGray"    alpha="1" r="209/256" g="209/256" b="209/256"/>
   <!-- off-white (or white) don't work! -->
   <!--vis name="AnlOff_White"    alpha="1" r="242/256" g="242/256" b="242/256"/-->
   <vis name="AnlDarkRed"      alpha="1" r="161/256" g="45/256"  b="45/256"/>
   <vis name="AnlDarkGreen"    alpha="1" r="0"       g="110/256" b="50/256"/>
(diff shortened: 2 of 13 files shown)


compact/hcal/backward_sampF.xml
```diff
@@ -12,6 +12,7 @@
 
   <define>
     <constant name="HcalEndcapN_allSensitive" value="1"/>
+    <constant name="HcalEndcapN_length_new"              value="5.0*HcalEndcapN_length"/>
```
veprbl: What's the point of this?

compact/hcal/backward_LFHCAL_geo.xml
veprbl: Did you intend to add this? Why?

veprbl (commented):
There are more changes than advertised. I'd like to confirm what are we reviewing?

compact/hcal/backward_sampF.xml
lkosarz: This is necessary to have a correct sampling fraction calculation and benchmarks will use it. Making all layers sensitive is not enough and gives wrong e/h response. The correct way is to make it thicker to include the leakage. We tested this.

compact/hcal/backward_template.xml
```diff
@@ -21,7 +21,7 @@
 
     <constant name="HcalEndcapNSingleLayerThickness"
              value="HcalEndcapNSteelThickness + HcalEndcapNPolystyreneThickness + HcalEndcapNLayerGap"/>
-    <constant name="HcalEndcapNLayer_NRepeat"   value="floor(HcalEndcapN_length / HcalEndcapNSingleLayerThickness)"/>
+    <constant name="HcalEndcapNLayer_NRepeat"   value="floor(HcalEndcapN_length_new / HcalEndcapNSingleLayerThickness)"/>
```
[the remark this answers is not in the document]
lkosarz: I'll keep it with "HcalEndcapN_length_new", to make it work for the sampling fraction calculation.

configurations/backward_hcal_only_sampF.yml
```diff
@@ -3,4 +3,3 @@ features:
     marco:
   hcal:
     backward_sampF:
-    backward_endcap_flux:
```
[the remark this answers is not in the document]
lkosarz: We have to remove "backward_endcap_flux" for the sampling fraction calculation. Otherwise it makes an overlap.

compact/hcal/backward_sampF.xml
veprbl: But what is the point of having an alias `HcalEndcapN_length_new` of the same value as `HcalEndcapN_length`?

lkosarz: The point is that it allows it to make the detector 5x thicker. Do you prefer another solution to make it work?

veprbl: Okay, I think I start to get it. Can we come up with more descriptive names.
Would if be okay to rename
`HcalEndcapN_length_new` -> `HcalEndcapN_length`
HcalEndcapN_length` -> `HcalEndcapN_baseline_length` (or `HcalEndcapN_nominal_length`)

compact/hcal/backward_LFHCAL_geo.xml
veprbl: Great. Please remove it from the submission.

configurations/backward_hcal_only_sampF.yml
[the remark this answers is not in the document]
veprbl: Please see the CI checks failing, if you haven't already.

Issue #993, opened 2025-11-14
Set `DetType_{ELECTROMAGNETIC, HADRONIC}` flags


ruse-traveler:
**Is your feature request related to a problem? Please describe.**

It was pointed out in discussion on [EICrecon#2078](https://github.com/eic/EICrecon/pull/2078) that DD4hep provides dedicated flags for electromagnetic vs. hadronic calorimeters (see [here](https://github.com/AIDASoft/DD4hep/blob/e369490a18a9857d5a61771f340c1791e8f9f7a3/DDCore/include/DD4hep/DetType.h#L53-L54)). These tags could be used for generically sorting electromagnetic vs. hadronic energy in later reconstruction algorithms such as PF or eID.

Unfortunately, only 1 calorimeter has set them.

**Describe the solution you'd like**

Each calorimeter should always include one or the other in its type flags.

**Describe alternatives you've considered**

Alternatively, we could rely on `enum`s maintained in either EICrecon or EDM4eic. However, this is not ideal as it introduces another set of flags to maintain.

PR #994, opened 2025-11-14
Add missing detector type flags

ruse-traveler:

This PR adds appropriate [DD4hep `DetType` flags](https://github.com/AIDASoft/DD4hep/blob/master/DDCore/include/DD4hep/DetType.h) to all calorimeters missing them. An example of how to extract these flags from calorimeter and tracker hits (where the latter includes PID output) is provided in snippets [here](https://github.com/eic/snippets/tree/main/Calorimetery/ExtractDetectorTypes).

Note that when this PR was opened, there was no `FAR` flag defined, and so I opted to label the FF calorimeters as `ENDCAP`. However, [DD4hep#1573](https://github.com/AIDASoft/DD4hep/pull/1573) adds this (alongside `BACKWARD`). We should follow up to add these once we pick up this DD4hep version.

- [x] Bug fix (issue #993 )
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/forwardEcal_geo.cpp
+++ b/src/forwardEcal_geo.cpp
@@ -12,6 +12,7 @@
 #include "DD4hep/Printout.h"
 #include <XML/Helper.h>
 #include <XML/Layering.h>
+#include <XML/Utilities.h>
 #include "forwardEcalMap.h"
 
 using namespace dd4hep;
@@ -293,7 +294,7 @@ static Ref_t createDetector(Detector& desc, xml_h handle, SensitiveDetector sens
   pv               = motherVol.placeVolume(envelopeVol, tr);
 
   // apply any detector type flags set in XML
-  dd4hep::xml::setDetectorTypeFlag(x_det, sdet);
+  dd4hep::xml::setDetectorTypeFlag(detElem, det);
 
   pv.addPhysVolID("system", detID);
   det.setPlacement(pv);


veprbl:
Have you tested querying these values? I don't think this works without explicit calls to `dd4hep::xml::setDetectorTypeFlag()` in each constructor.

ruse-traveler:
It took a minute to work through it, but it looks like things are working mostly as expected! For some reason, the type flags don't seem to be propagated for the DIRC, pfRICH, FHCal insert, and far backward detectors. I haven't figured out why yet...

detRegionRxEta epic994 py8ncdis18x275minq1Knevt1K d5m12y2025
detTypeRxEta epic994 py8ncdis18x275minq1Knevt1K d5m12y2025
detRegionRxZ epic994 py8ncdis18x275minq1Knevt1K d5m12y2025
detTypeRxZ epic994 py8ncdis18x275minq1Knevt1K d5m12y2025

veprbl:
Should we propose `DetFlag_FAR`? To be used as `DetFlag_Far + DetFlag_FORWARD`.

veprbl (dismissed):

ruse-traveler:
> Should we propose `DetFlag_FAR`? To be used as `DetFlag_FAR + DetFlag_FORWARD`.

I think that also makes sense! It definitely would be useful to be able to distinguish between detectors that are physically in the endcap vs. those that are along the beamline.

veprbl (dismissed):

veprbl:
Acts is not happy with this one:
```
22:50:50    Acts           VERBOSE   --> Layer candidate match: VertexBarrel_layer\d -> VertexBarrel_layer2
22:50:50    Acts           VERBOSE   Volume: 'VertexBarrelSubAssembly' is a compound volume -> resolve the sub volumes
```

veprbl:
Marking irrelevant detectors as beampipe (itself) messes with Acts geometry conversion.

src/BarrelTrackerWithFrame_geo.cpp
```diff
@@ -52,6 +52,9 @@ static Ref_t create_BarrelTrackerWithFrame(Detector& description, xml_h e, Sensi
   string det_name = x_det.nameStr();
   DetElement sdet(det_name, det_id);
 
+  // apply any detector type flags set in XML
+  dd4hep::xml::setDetectorTypeFlag(x_det, sdet);
+
```
ruse-traveler:
```suggestion
```

src/BeamPipeTracking_geo.cpp
```diff
@@ -29,6 +30,9 @@ static Ref_t create_detector(Detector& description, xml_h e, SensitiveDetector s
   DetElement sdet(det_name, det_id);
   Assembly assembly(det_name + "_assembly");
 
+  // apply any detector type flags set in XML
+  dd4hep::xml::setDetectorTypeFlag(x_det, sdet);
+
```
ruse-traveler:
```suggestion
```

veprbl:
Perhaps, this is because BIC is also a tracking detector.

ruse-traveler:
> Perhaps, this is because BIC is also a tracking detector.

Yeah, that was it 🫤 I was getting into hot water with the `setTypeFlag(...)` calls in `src/BarrelCalorimeterImaging_geo.cpp`. I reverted those changes, and confirmed that `test_ACTS.cxx` runs successfully on craterlake after the reversion. I think we can live with the AstroPix hits not having a type for now.

In the process, I caught that (1) I wasn't propagating the type flags through to the BIC ScFi hits, and (2) I hadn't updated the newer FEMC files. After doing that, I'm seeing the correct types everywhere I would expect them! From my side, I think this is ready to go.

typeRxEta compareTypeFlags_afterFixingFEMCTypos_onlyScFiHits py8ncdis18x275q1Knevt100 d3m4y2026
typeRxZ compareTypeFlags_afterFixingFEMCTypos_onlyScFiHits py8ncdis18x275q1Knevt100 d3m4y2026

veprbl (approved):
LGTM

PR #999, opened 2025-11-17
Delete src/BackwardsLumiVac_geo.cpp

veprbl:
Not used since 736b993c6eaab9f8ba3167ac672dd689fdc7fe29


sjdkay:
> Looks fine, I think any reintroduction of this would need to be more detailed.
> 
> I haven't kept up to date with the lumi beamline plans and how much the main ePIC simulation is being used and itself kept up to date. @sjdkay

Typically we don't simulate the full beamline as brem photons are propagated to the converter foil position and pre-converted before simulating, at least of the lumi PS. I can do a quick check of the geometry with/without this and confirm, but I don't think it's currently used.

PR #978, opened 2025-11-07
New far backward dipole magnet volumes added

simonge:
### Briefly, what does this PR introduce?
Creates the volumes of the dipole magnets in the far backward region. 

Currently in the wrong position, this will be updated to the correct location, or merged into a separate PR when the beamline changes have been made.

Opened to check the naming of the components, materials are appropriate and the simplification of the CAD geometry is acceptable.

### What kind of change does this PR introduce?
- [ ] Bug fix (issue #__)
- [x] New feature (issue #963 )
- [ ] Documentation update
- [ ] Other: __

### Please check if this PR fulfills the following:
- [ ] Tests for the changes have been added
- [ ] Documentation has been added / updated
- [ ] Changes have been communicated to collaborators

PR #1004, opened 2025-11-19
Setup new far backwards geometry

simonge:
Makes the changes required to match the current EIC accelerator beamline setup in the Far Backward region. The vacuum pipe spanning the tagger drift volume has needed some significant revisions, simplifying the implementation but removing some functionality to allow the "antechamber" design.

Incorporates the volumes reviewed in https://github.com/eic/epic/pull/978, placed correctly.

Introduces a new script which automatically updates the xml description of the magnets, strengths and positions based on a provided twiss and survey file. (The current files have been added to the calibrations but could be removed as don't do anything other than act as an input to the script.)

Some extra tinkering of positions may be required. In particular the luminosity pipe along the z-axis doesn't extend into the B2AeR magnet. There is a small vacuum/air boundary between the tagger exit windows and the beampipe cylinder. These could be fixed in this PR or a future PR as they won't have the most significant impacts.

The Luminosity components will need their positions updated to match the new layout.

Not the cleanest PR.

- [x] Bug fix (issue #963)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

Hopefully nothing.
There is likely to be significant changes to the acceptance and reconstruction of the taggers.

### DIFF
--- a/compact/tracking/definitions_craterlake.xml
+++ b/compact/tracking/definitions_craterlake.xml
@@ -214,7 +214,7 @@
     <plugin name="epic_FileLoader">
       <arg value="cache:$DETECTOR_PATH:/opt/detector"/>
       <arg value="file:calibrations/materials-map.cbor"/>
-      <arg value="url:https://eicweb.phy.anl.gov/-/project/473/uploads/c21564c0821bad5f8d0f1d21f9bc6b9c/material-map.cbor"/>
+      <arg value="url:https://eicweb.phy.anl.gov/-/project/473/uploads/e4042d004e28ce252e969a7cde5e21b8/material-map.cbor"/>
 
     </plugin>
   </plugins>


simonge:
Visual comparison between the CAD and simulation. Looks like definitely some positions to adjust still.
image
image

simonge:
Great, thanks @nat93 just thought I'd double check. Where have the values you've used come from, was this the magnet CAD drawings?

Presumably there will need to be some gap between the pipe and support tube, do you have an idea of what this would need to be?

simonge:
Most of the issues and overlaps have been addressed.

There is currently still a discontinuity linking the luminosity line with the electron line coming out of B2AeR. This will need changes to BeamPipeChain, where I plan to create a separate PR which allows merging the a chain of the current volumes into boolean solids and also adding an optional branched pipe which will also need to be a boolean solid with the summed vacuum subtracted from the summed pipes.

image

Issue #987, opened 2025-11-04
Regression in tracking in the B0 with newer version of epic


ajentsch:
As the first samples from the 25.10.1 exclusive (DVCS, in particular, for the example here) physics campaign came out, an issue was noted in the acceptance of the B0 tracker.

Using files from 25.06.1 (they were on-hand), the following acceptance is noted for the B0: 

<img width="529" height="538" alt="Image" src="https://github.com/user-attachments/assets/5d5bba11-6d33-427d-bd1c-d8525dd1dc3c" />

The B0 is the portion in the dark blue.

This is the result from the 25.10.1 files:

<img width="647" height="647" alt="Image" src="https://github.com/user-attachments/assets/de4f5826-e96b-4113-9297-a49c7eab84d4" />

Focusing on the dark blue curve, the B0 angular acceptance has been reduced significantly, with a cutoff almost 1.5 mrad higher in theta than before. 

Looking at samples from 25.08.0 (left) and 25.10.1 (right), the hit distribution look as the are expected to look:

<img width="1920" height="626" alt="Image" src="https://github.com/user-attachments/assets/b4852566-1200-4fd4-ac7f-a2a9bade616d" />

The only geometry change is seen on the left side of these plots, and that would only affect very high-angles in the B0, not the ones at the lower end, and certainly not to the degree we are seeing.

Using small samples from 25.08.0 (left) and 25.10.1 (right), we also see the number of reconstructed tracks is significantly reduced, for the same MC events:

<img width="1600" height="527" alt="Image" src="https://github.com/user-attachments/assets/f8c6edc4-1776-4002-9cc4-72f4bc1d1b81" />

The conclusion is that some changes in ACTS have affected the B0 tracking and resulted in fewer reconstructed tracks.

ajentsch:
@veprbl I do not understand this. The adjustment to the geometry was only on the tiny section of the tracker which is between the beam pipes. That would only affect the acceptance in a tiny region of the detector, and at higher pT, not the region we see the loss. How does this change affect the tracking acceptance but in the wrong region of pT?

PR #1009, opened 2025-11-26
B0_tracker: set envelope_z_{min,max} to 1 mm

veprbl:
Since update to Acts 39, @wdconinc had implemented several fixes to correct Acts routines used for automatic DD4hep to Acts surface conversion (https://github.com/acts-project/acts/pull/4496, https://github.com/acts-project/acts/pull/4502, https://github.com/acts-project/acts/pull/4620). Those patches were applied to Acts in our nightly containers since July, but we had to revert those for the stable releases because they were killing the B0 reconstruction as seen in before & after of the tracking_performance_dis benchmark:
image
Recent issue like #987 can take advantage of resolving some of the issues with Acts surface dimensions. We also need to make sure that EICrecon can advance to using the newer versions of Acts that include the patches.

It looks like the issue with broken tracking has to do with placement of "approach" surfaces. The surfaces are simple shapes (e.g. disk) that are supposed to enclose the actual sensitive surfaces to which hits belong. CKF tracking relies on being able to correctly visit sensitive surfaces that belong to those bins.

This schematic shows the coordinates of the surfaces in Acts 39 before all the fixes:
before
and this is how they look with the patches
nominal
the issue is apparently is with how tight is the space between the surfaces.

One way to fix this is to see that current implementation of the geometry uses the Assembly volume associated with layer's DetElement and sets the dimensions of the layer via
https://github.com/eic/epic/blob/cf46b7c046d56dd57ff0df7e85fa6a03a03b2037/src/B0Tracker_geo.cpp#L229-L232
https://github.com/eic/epic/blob/cf46b7c046d56dd57ff0df7e85fa6a03a03b2037/compact/far_forward/B0_tracker.xml#L114-L115
https://github.com/eic/epic/blob/cf46b7c046d56dd57ff0df7e85fa6a03a03b2037/compact/far_forward/B0_tracker.xml#L127-L128
https://github.com/eic/epic/blob/cf46b7c046d56dd57ff0df7e85fa6a03a03b2037/compact/far_forward/B0_tracker.xml#L140-L141
https://github.com/eic/epic/blob/cf46b7c046d56dd57ff0df7e85fa6a03a03b2037/compact/far_forward/B0_tracker.xml#L153-L154
The zero-valued envelopes appear to be consistent with approach surfaces almost touching each other, it must be the right behaviour, and the setting is inadequate. This PR adjusts the envelope to 1 mm, which results in the following arrangement
1mm

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

### DIFF
--- a/compact/far_forward/B0_tracker.xml
+++ b/compact/far_forward/B0_tracker.xml
@@ -112,7 +112,7 @@
       </module>
       <layer id="1">
         <envelope  vis="FFTrackerLayerVis" rmin_tolerance="0*mm" rmax_tolerance="0*mm"
-                   zmin_tolerance="0*mm" zmax_tolerance="0*mm" length="1.0*cm"
+                   zmin_tolerance="1*mm" zmax_tolerance="1*mm" length="1.0*cm"
                    zstart="B0TrackerLayer1_zstart" />
         <ring phi0="B0TrackerLayerBigMod_phi0" dphi="B0TrackerModOpeningAngle"
           r="B0TrackerMod1Inner_r+B0TrackerMod1_y/2.0" zstart="0.0*mm"
@@ -125,7 +125,7 @@
       </layer>
       <layer id="2" >
         <envelope vis="FFTrackerLayerVis" rmin_tolerance="0*mm" rmax_tolerance="0*mm"
-                  zmin_tolerance="0*mm" zmax_tolerance="0*mm" length="1.0*cm"
+                  zmin_tolerance="1*mm" zmax_tolerance="1*mm" length="1.0*cm"
                   zstart="B0TrackerLayer2_zstart" />
         <ring phi0="B0TrackerLayerBigMod_phi0" dphi="B0TrackerModOpeningAngle"
           r="B0TrackerMod1Inner_r+B0TrackerMod1_y/2.0" zstart="0.0*mm"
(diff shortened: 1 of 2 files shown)


wdconinc:
Nice investigation! Looks like an improvement without downsides to me.

I am still wondering if we get reconstructed tracks that use B0 hits everywhere, or if those B0 tracks are only reconstructed when the hits are outside the 38 (?) mm beam pipe volume around the z axis.

veprbl:
> This is awesome - thank you for figuring a solution!

Fine print was that this is not meant as the solution to 25.08.0 regression. The benchmarks above already look suspiciously good though.

wdconinc:
> Let's say that's outside of the scope for this PR

Agreed.

> I don't really get the discussion about the beam pipe volume. I think, it could be made as small as necessary, since our current acts beam pipe is smaller than the actual machine beam pipe.

I think there's a minimum we need to stay above: track seeds have to start from within the beam pipe. With our crabbing angle we have vertices that are not on the z axis. We have to have an acts beam pipe volume that at least encompasses those (I think that may be a few mm, not cm though).

Also, as small as necessary with the size of the B0 right now is actually zero. The circumscribed circle around center of the B0 module placement that contains all B0 modules is going through the z-axis origin where the acts beam pipe is centered. That's why the B0 may just need to be a few mm smaller in radius to fit properly.

PR #1011, opened 2025-12-01
feat: Acts Gen3 geometry test

wdconinc:
This PR creates a version of `test_ACTS.cxx` for Gen3 geometry, in order to test changes that will be made in the context of the Acts geometry hackathon this week.


wdconinc:
See https://github.com/eic/EICrecon/pull/2227 instead.

PR #1021, opened 2025-12-19
Multi sensitive volume clusters

ybedfer:
Implement realistic pitch into Multi-SensitiveVolume-MPGD, extending PR#https://github.com/eic/epic/pull/1014.

- [* ] New feature, see PR#2264 ("https://github.com/eic/EICrecon/pull/2264").

- [ *] Changes have been communicated to collaborators

No

Eventually yes.


compact/tracking/mpgd_outerbarrel_2DStrip.xml
```diff
@@ -181,9 +184,9 @@
...
-        <segmentation name="UStrips" type="CartesianGridUV" key_value="1" grid_size_u="0.150*mm*sqrt(12)"    grid_size_v="MPGDOuterStrip_range" grid_angle="pi*rad/4" />
-        <segmentation name="VStrips" type="CartesianGridUV" key_value="2" grid_size_u="MPGDOuterStrip_range" grid_size_v="0.150*mm*sqrt(12)"    grid_angle="pi*rad/4"/>
-        <segmentation name="BotRad"  type="CartesianGridUV" key_value="3" grid_size_u="0.150*mm*sqrt(12)"    grid_size_v="0.150*mm*sqrt(12)"    grid_angle="pi*rad/4" />
+        <segmentation name="UStrips" type="CartesianGridUV" key_value="1" grid_size_u="MPGDOuterBarrelPitch" grid_size_v="MPGDOuterStrip_range" grid_angle="pi*rad/4" offset_u="MPGDOuterBarrelPitch/2" />
+        <segmentation name="VStrips" type="CartesianGridUV" key_value="2" grid_size_u="MPGDOuterStrip_range" grid_size_v="MPGDOuterBarrelPitch" grid_angle="pi*rad/4" offset_v="MPGDOuterBarrelPitch/2" />
+        <segmentation name="BotRad"  type="CartesianGridUV" key_value="3" grid_size_u="0.150*mm*sqrt(12)"    grid_size_v="MPGDOuterBarrelPitch" grid_angle="pi*rad/4" />
```
veprbl: Why only v is updated only for BotRad and not TopRad?

veprbl (dismissed):
LGTM

compact/tracking/mpgd_outerbarrel_2DStrip.xml
ybedfer: Hello,
It's a mistake.
But a mistake that has no consequence. Or so I think. The Radiators are only used at simulation time (`npsim`) and the CellID that is then computed _via_ segmentation and assigned to the SimHits is later on disregarded. Only the EDep of the SimHits. matters.
Yet, I think that the best is to have a pixel segmentation in those Radiators. I have corrected the XML file and committed.
Thank you for your careful review.

veprbl (dismissed):

compact/tracking/mpgd_outerbarrel_2DStrip.xml
veprbl: I take it that pixel resolution is better than resolution of true strip dimensions, so that does not hurt because you rely on coordinates. Why not make pixels match the dimensions of strips? Otherwise you might experience small effects due to pitch mismatch.

ybedfer: Hello,

> Why not make pixels match the dimensions of strips? Otherwise you might experience small effects due to pitch mismatch.

I don't know whether it's a good idea.
On the one hand, it worsens what I would call the readability of the podio `TTree`:  the readability of  what I get when I  `TTree:Scan` it, that is. When the segmentation has a pitch equal to the resolution, I can find back in the CellIDs of my RecHits, pieces of the CellIDs of my original SimHits. That's this is a genuine advantage is questionable, I agree. So more so as we have the opposite situation when comparing SimHits to RawHits.
On the other hand, there's what you call "pitch mismatch". But I don't get what this might be. Can you be more precise?

veprbl: All I'm saying is that if you sample at discrete pixel pitch coordinates and then essentially histogram those with strip pitch you will get "beat signal" modulations in the resulting distribution.

veprbl: If this is not an issue, let's mark it resolved and we can move to merging.

ybedfer: Hello,
I have added comments describing the alternative for what I call Helper Sensitive Volumes: either resolution or pitch.

veprbl: Thanks for updating the comments. I want to check with you if they are misleading. You discuss helper volumes that may not necessarily have adequate segmentation, e.g.:

> - Two of the Sensitive Volumes, '(p|n)ThinGap", have a meaningful Segmentation, with a pitch corresponding to that of the actual detector.

Yet those correspond to BotRad and TopRad segmentation, and don't have the physical pitch.

ybedfer: I have been fighting with the English of the comments. Not sure I have been successful...
The confusion may come from the fact that it's not clearly stated that to each of the 5 sensitive volumes corresponds a distinct Segmentation. Added to the fact  that Volumes and Segmentation entries do not have the same name. These names cannot be strictly the same (because of CyMBaL where one has to in addition distinguish between inner and outer sections). But they can be made more similar. I am preparing a new commit where I implement a better naming scheme.

veprbl: Okay, maybe I got confused. Those are not BotRad and TopRad. Those are U/VStrips.

Maybe my confusion now is with then you say
> The other three Volumes are Helpers, used only temporarily in the production of MC data.

Not sure if "Helpers" is a good word. They define a pixel binning that is used by the simulation.

ybedfer: Hello,
I committed a new attempt.
Where I still have the word "Helpers" (sorry: I had not read your message before committing).
By "Helpers" I mean something that comes into play only temporarily, for the time MC data are produced and digitized, and are then forgotten. In the processing of real data, they are even completely ignored.

veprbl (approved):
LGTM

compact/tracking/mpgd_outerbarrel_2DStrip.xml
veprbl: I like latest change to the naming. Before you could still look at `key`, but this helps even more.

PR #1048, opened 2026-02-25
feat: use BoundedMultipoleMagnet with AABB pre-filter to MultipoleMagnet

wdconinc:
This PR adds a pre-filter to the MultipoleMagnet to apply  an axis-aligned bounding box. This avoids expensive rotated r-z Tube::Contains evaluation in many cases, in particular since our fields are all added in an OverlayField, and assigned to the global field manager (i.e. evaluate for the far forward and far backward regions for every particle in the EM calorimeter showers).

Profiling indicates we spend 30% of the total ddsim (no optical photons) simulation time (for DIS 18x275 Q2>1000) in dd4hep::MultipoleField::fieldComponents, of which 25% (of the total simulation time) in the Tube::Contains (this is based on callgrind). This PR aims to address that (more profiling below).

Other options considered:
- assigning the multipole fields to a geometry volume so we only need to evaluate them when inside those: discarded since it doesn't handle the stray fields in the gaps between magnets well (and I think @ajentsch once said that's important).
- placing the entire far forward and far backward regions inside their own volume (not assembly as currently): discarded since this would require changes to the geometry structure and a clean interface between inside/outside these volumes, which is hard to do without clear axis-aligned boundaries.

Comparing profiling in the ctree artifacts (for DIS 10x100 Q2 > 1000 this time, sorry), we see the total time of simulation go from 822s to 768s. The time in dd4hep::MultipoleField::fieldComponents (in the various places in the call tree where it is called) goes from 45+5+6+4s to 12+1+1+1s, for a 6% speed increase.

- [ ] Bug fix (issue #__)
- [x] New feature (issue: MultipoleMagnet is performance hog)
- [ ] Documentation update
- [ ] Other: __

No.

No.

### DIFF
--- a/src/BoundedMultipoleMagnet.cpp
+++ b/src/BoundedMultipoleMagnet.cpp
@@ -12,9 +12,9 @@
  *  work is skipped for the vast majority of tracks that are nowhere near the
  *  far-forward magnets.
  *
- *  Usage – replace `type="MultipoleMagnet"` with `type="epic_BoundedMultipoleMagnet"`:
+ *  Usage – replace `type="MultipoleMagnet"` with `type="BoundedMultipoleMagnet"`:
  *  \code
- *  <field name="B0PF_Magnet" type="epic_BoundedMultipoleMagnet">
+ *  <field name="B0PF_Magnet" type="BoundedMultipoleMagnet">
  *    <position x="..." y="0" z="..."/>
  *    <rotation x="0" y="..." z="0"/>
  *    <shape type="Tube" rmin="0" rmax="..." dz="..."/>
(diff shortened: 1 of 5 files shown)


src/BoundedMultipoleMagnet.cpp
```diff
@@ -0,0 +1,148 @@
...
+ *
+ *  Drop-in replacement for DD4hep's built-in `MultipoleMagnet` field type.
+ *  The upstream factory parses `<position>`, `<rotation>`, `<shape>`, and
+ *  `<coefficient>` from the compact XML unchanged.  Before delegating to the
+ *  full tube-containment check and multipole calculation, a cheap rectangular
+ *  AABB test in global coordinates is applied so that the expensive per-step
+ *  work is skipped for the vast majority of tracks that are nowhere near the
+ *  far-forward magnets.
+ *
+ *  Usage – replace `type="MultipoleMagnet"` with `type="epic_BoundedMultipoleMagnet"`:
```
wdconinc:
```suggestion
 *  Usage – replace `type="MultipoleMagnet"` with `type="BoundedMultipoleMagnet"`:
```

wdconinc:
Upstreamed in https://github.com/AIDASoft/DD4hep/pull/1584, included in DD4hep v1.36, https://github.com/eic/containers/pull/232. Closing here.

PR #1051, opened 2026-02-26
feat: switch BIC, FEMC, and EcalBarrelScFi to paramVolume1D

wdconinc:
This PR converts loops of explicit placement to parametrized placement in the BIC, FEMC, and EcalBarrelScFi detectors. This reduces the TGeoNode count, speeds up geometry initialization, and reduces memory use.

For **BIC and FEMC**: modules/towers are placed via `paramVolume1D`.

For **EcalBarrelScFi**: scintillating fibers are placed via `paramVolume1D`, one call per HCP z-row per readout grid. Each z-row is wrapped in a thin Assembly with `physVolID("z", row_index)` to give unique VolumeManager cell-ID codes across rows (required because DD4hep parameterised volumes can only carry `volID=0`; the Geant4 copy number encodes the column index into the "fiber" readout field at runtime).

**Readout change for EcalBarrelScFi**: the `z` field now encodes the fiber row index within the grid (previously always 0), and `fiber` encodes the column index within that row (previously a sequential per-grid ID). The total set of unique cell IDs is unchanged; the encoding has changed.

Also fixes an O(N_fibers × N_grids) performance issue in geometry initialization: the TGeoPolygon for each grid is now built once instead of once per fiber.

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Documentation update
- [ ] Other: __

EcalBarrelScFi readout: the `z` and `fiber` fields have new semantics (row index and column index within row, respectively). Reconstruction code that uses these fields for individual fiber identification needs updating.

Yes — EcalBarrelScFi readout IDs change encoding.

### DIFF
--- a/src/BarrelCalorimeterImaging_geo.cpp
+++ b/src/BarrelCalorimeterImaging_geo.cpp
@@ -339,36 +339,40 @@ static Ref_t create_detector(Detector& desc, xml_h e, SensitiveDetector sens) {
           printout(DEBUG, "BarrelCalorimeterImaging", "Stave %s modules starting at x=%f, y=%f",
                    stave_name.c_str(), x0, y0);
 
-          // Place modules
+          // Place modules using parameterised placement (G4PVParameterised), reducing
+          // N individual G4PVPlacement objects to one parameterised volume. Only a
+          // 1D layout (nx == 1) along the stave Y axis is supported; nx > 1 requires
+          // a 2D parameterisation with a different module-ID encoding.
+          if (nx != 1) {
+            except("BarrelCalorimeterImaging",
+                   "xy_layout with nx=%d is not supported; only nx=1 is implemented "
+                   "(stave %s, layer %d)",
+                   nx, stave_name.c_str(), layer_num);
+          }
+          // With G4PVParameterised the copy number (0..ny-1) IS the module ID.
+          // physVolID is set only on the first (anchor) placement with value 0;
+          // all copies share that ID and use their copy number as the actual module ID.
           int i_module = 0;
-          for (auto i_x = 0; i_x < nx; ++i_x) {
-            for (auto i_y = 0; i_y < ny; ++i_y) {
-
-              // Create module
-              std::string module_name = _toString(i_module, "module%d");
-              DetElement module_element(stave_element, module_name, i_module);
-
-              // Place module
-              auto x = x0 + i_x * dx;
-              auto y = y0 + i_y * dy;
-              Position module_pos(x, y, 0);
-              PlacedVolume module_physvol = stave_volume.placeVolume(module_volume, module_pos);
-              module_physvol.addPhysVolID("module", i_module);
-              module_element.setPlacement(module_physvol);
-
-              // Add sensitive volumes
-              for (auto& sensitive_physvol : module_sensitives) {
-                DetElement sensitive_element(module_element, sensitive_physvol.volume().name(),
-                                             i_module);
-                sensitive_element.setPlacement(sensitive_physvol);
-
-                auto& sensitive_element_params =
-                    DD4hepDetectorHelper::ensureExtension<dd4hep::rec::VariantParameters>(
-                        sensitive_element);
-                sensitive_element_params.set<std::string>("axis_definitions", "XYZ");
-              }
-
-              i_module++;
+          PlacedVolume first_pv =
+              stave_volume.paramVolume1D(Transform3D(Position(x0, y0, 0)), module_volume,
+                                         static_cast<size_t>(ny), Transform3D(Position(0, dy, 0)));
+          auto& placements = first_pv.data()->params->placements;
+          for (auto i_y = 0; i_y < ny; ++i_y, ++i_module) {
+            PlacedVolume module_physvol = placements[i_y];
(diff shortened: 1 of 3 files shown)


src/BarrelCalorimeterImaging_geo.cpp
```diff
@@ -339,36 +339,40 @@ static Ref_t create_detector(Detector& desc, xml_h e, SensitiveDetector sens) {
...
+          if (nx != 1) {
+            except("BarrelCalorimeterImaging",
+                   "xy_layout with nx=%d is not supported; only nx=1 is implemented "
+                   "(stave %s, layer %d)",
+                   nx, stave_name.c_str(), layer_num);
+          }
+          // With G4PVParameterised the copy number (0..ny-1) IS the module ID.
+          // physVolID is set only on the first (anchor) placement with value 0;
+          // all copies share that ID and use their copy number as the actual module ID.
           int i_module = 0;
```
wdconinc:
```suggestion
```

src/BarrelCalorimeterImaging_geo.cpp
```diff
@@ -339,36 +339,40 @@ static Ref_t create_detector(Detector& desc, xml_h e, SensitiveDetector sens) {
...
-                        sensitive_element);
-                sensitive_element_params.set<std::string>("axis_definitions", "XYZ");
-              }
-
-              i_module++;
+          PlacedVolume first_pv =
+              stave_volume.paramVolume1D(Transform3D(Position(x0, y0, 0)), module_volume,
+                                         static_cast<size_t>(ny), Transform3D(Position(0, dy, 0)));
+          auto& placements = first_pv.data()->params->placements;
+          for (auto i_y = 0; i_y < ny; ++i_y, ++i_module) {
```
wdconinc:
```suggestion
          for (auto i_y = 0; i_y < ny; ++i_y) {
```

src/BarrelCalorimeterImaging_geo.cpp
```diff
@@ -339,36 +339,40 @@ static Ref_t create_detector(Detector& desc, xml_h e, SensitiveDetector sens) {
...
+            PlacedVolume module_physvol = placements[i_y];
+            std::string module_name     = _toString(i_module, "module%d");
+            DetElement module_element(stave_element, module_name, i_module);
```
wdconinc:
```suggestion
            std::string module_name     = _toString(i_y, "module%d");
            DetElement module_element(stave_element, module_name, i_y);
```

src/BarrelCalorimeterImaging_geo.cpp
```diff
@@ -339,36 +339,40 @@ static Ref_t create_detector(Detector& desc, xml_h e, SensitiveDetector sens) {
...
+            if (i_y == 0) {
+              module_physvol.addPhysVolID("module", 0);
+            }
+            module_element.setPlacement(module_physvol);
+            for (auto& sensitive_physvol : module_sensitives) {
+              DetElement sensitive_element(module_element, sensitive_physvol.volume().name(),
+                                           i_module);
```
wdconinc:
```suggestion
                                           i_y);
```

wdconinc (commented):
For BIC, `i_module` is now identical to `i_y`.

wdconinc:
I think copilot has successfully convinced me that this geometry change modifies stepping in such a way that we don't expect the same hits for tracks that go through the modified geometry since the navigator behaves differently for parametrized and non-parametrized volumes.

wdconinc:
EcalBarrelScFi `z` encoding is wrong here.

src/forwardEcal_geo.cpp
akioogawa, line 196: Can this loop be also using paramVolume1D?

akioogawa, line 244: And this loop?

akioogawa:
Sorry I did not realize this was up. Now I'm checking out and compiling. I will be testing this (maybe next week).

src/forwardEcal_geo.cpp
wdconinc: No, there is staggering of the layers inside it. That makes it harder to do. It may (@veprbl claims) be possible to do this with two paramVolume1D instances (one for even, one for odd), or it is may be possible to do this with a 2-fiber unit cell, but in any case it is more complicated than just a parametrized placement. And it would likely change the numbering scheme of the fibers. And it would get is proportionally less gain than this part gets us since we are now starting to bump up against other irreducible effects.

wdconinc: > No, there is staggering of the layers inside it.

I guess what I meant here is that you can't easily use paramVolume2D.

veprbl: > with two paramVolume1D

with one paramVolume3D

wdconinc: Ok, I'll need a picture for this...

akioogawa:
something is wrong... https://www.star.bnl.gov/public/spin/akio/epic/map.homo.png. Nope that was my bug in plotting. Geometry looks good in dd_web_display for both Homogeneous and SiPM. Hit mapping seems to be same as before. dE for gamma looks good.

akioogawa (approved):
Looks good

PR #1062, opened 2026-03-13
calibrations.xml: use HTTPS(epic-data) instead of XRootD for `Low-Q2_Steering_Reconstruction.onnx`

veprbl:
I'm told `xrdtest` is not a great path for the file. While we are moving out (to Rucio), I'd like to suggest decommissioning the direct URL for XRootD.

We have a CI action that may upload to XRootD, but we can agree that it's still okay to write to xrdtest for temporary artifacts.


veprbl:
With ongoing outage, I'm reminded that this is a change we might want to implement.
cc @wdconinc

wdconinc (approved):

simonge:
I will have to remember the approach to automatically updating the file on epic-data. It was a bit less convenient than using xrootd.

veprbl:
I think we need to revise it to use Rucio (at least for uploads) anyway.

PR #1072, opened 2026-04-07
Convert thisepic.sh to POSIX shell, separate from setup.sh

veprbl:
While testing the https://github.com/eic/containers/pull/228 I discovered that the thisepic.sh does not work in container initialization due to it being a bash script. This converts it to a POSIX shell script. setup.sh logic is moved to a separate file, and it now reports that it's really an error.


wdconinc (approved):
LGTM. Thanks!

PR #1076, opened 2026-04-13
run_material_map_validation.sh: re-enable dist and 1D scripts

veprbl:
This is a small follow-up on #1008, resolving some remaining technical debt.


veprbl:
Appears to be working:
image
image

wdconinc (approved):

PR #1081, opened 2026-04-16
Implement a tracking region volume to support Geant4TVUserParticleHandler

### LINKED ISSUE #43: Extend tracking region to include far forward decays
We should take advantage of facilities that address https://github.com/AIDASoft/DD4hep/issues/1302. Some study need to be performed to ensure that the file sizes remain adequate with the tracking region configuration that we choose.

This may address one of the issues brought up in https://github.com/eic/epic/issues/1069

veprbl:
To be used with https://github.com/eic/npsim/pull/44
This is needed to resolve https://github.com/eic/npsim/issues/43

### DIFF
--- a/compact/tracking_region.xml
+++ b/compact/tracking_region.xml
@@ -19,7 +19,7 @@
     <comment>
       Additional constants for the purpose of defining the additional extrusion
     </comment>
-    <constant name="tracker_region_forward_extrusion_zmax"   value="38 * m"/>
+    <constant name="tracker_region_forward_extrusion_zmax"   value="36 * m"/>
     <constant name="tracker_region_forward_extrusion_rmax"   value="15 * cm"/>
     <constant name="tracker_region_forward_extrusion_rot_y"  value="ionCrossingAngle"/>
 
--- a/compact/tracking_region.xml
+++ b/compact/tracking_region.xml
@@ -23,8 +23,10 @@
 
   <comment>
     This volume is used by the *Geant4TVUserParticleHandler* (default in the latest npsim)
+
+    Set `connected="true"` to make it visible in a geometry display
   </comment>
-  <parallelworld_volume name="tracking_volume" anchor="/world" material="Air" connected="true" vis="VisibleBlue">
+  <parallelworld_volume name="tracking_volume" anchor="/world" material="Air" connected="false" vis="VisibleBlue">
     <comment>
       We first build a tube using the legacy parameters and add a ion beamline extrusion
     </comment>
--- a/compact/tracking_region.xml
+++ b/compact/tracking_region.xml
@@ -40,7 +40,7 @@
       <rotation x="0" y="tracker_region_forward_extrusion_rot_y" z="0" />
       <position x="sin(tracker_region_forward_extrusion_rot_y) * tracker_region_forward_extrusion_zmax / 2"
                 y="0"
-                z="cos(tracker_region_forward_extrusion_rot_y) * tracker_region_forward_extrusion_zmax / 2" />
+                z="cos(tracker_region_forward_extrusion_rot_y) * tracker_region_forward_extrusion_zmax / 2 - (tracker_region_zmax + tracker_region_zmin) / 2" />
     </shape>
     <position x="0" y="0" z="(tracker_region_zmax + tracker_region_zmin) / 2" />
     <rotation x="0" y="0" z="0" />
--- a/compact/tracking_region.xml
+++ b/compact/tracking_region.xml
@@ -33,7 +33,7 @@
 
     Set `connected="true"` to make it visible in a geometry display
   </comment>
-  <parallelworld_volume name="tracking_volume" anchor="/world" material="Air" connected="true" vis="VisibleBlue">
+  <parallelworld_volume name="tracking_volume" anchor="/world" material="Air" connected="false" vis="VisibleBlue">
     <comment>
       We first build a tube using the legacy parameters and add a ion beamline extrusion
     </comment>
--- a/compact/tracking_region.xml
+++ b/compact/tracking_region.xml
@@ -32,7 +32,7 @@
   </comment>
   <parallelworld_volume name="tracking_volume" anchor="/world" material="Air" connected="false" vis="VisibleBlue">
     <comment>
-      We first build a tube using the legacy parameters and add a ion beamline extrusion
+      We first build a tube using the legacy parameters and add an ion beamline extrusion
     </comment>
     <shape type="BooleanShape" operation="Union">
       <shape type="Tube" rmin="0 * cm" rmax="tracker_region_rmax" dz="(tracker_region_zmax - tracker_region_zmin) / 2" />
--- a/compact/tracking_region.xml
+++ b/compact/tracking_region.xml
@@ -30,7 +30,7 @@
 
     Set `connected="true"` to make it visible in a geometry display
   </comment>
-  <parallelworld_volume name="tracking_volume" anchor="/world" material="Air" connected="false" vis="VisibleBlue">
+  <parallelworld_volume name="tracking_volume" anchor="/world" material="Air" connected="false" vis="BlueVis">
     <comment>
       We first build a tube using the legacy parameters and add an ion beamline extrusion
     </comment>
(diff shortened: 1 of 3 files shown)


veprbl:
It could use some guidance on what will be the extent of the region. It's currently 10 m in both directions with 1 cm radius.

compact/tracking_region.xml
```diff
@@ -0,0 +1,42 @@
...
+    For now they are forced to be symmetric, and we cover up to the start of the backward ECAL.
+    (this means in the forward region we stop in the middle of the RICH).
+  </comment>
+
+  <define>
+
+    <comment>
+      These constants are used directly by the *Geant4TCUserParticleHandler* (default in ddsim)
+    </comment>
+    <constant name="tracker_region_rmax"   value="EcalBarrel_rmin"/>
```
wdconinc, line 15: We should probably unset these in compact/definitions.xml then.

veprbl: Thanks. Committed now.

compact/tracking_region.xml
```diff
@@ -0,0 +1,42 @@
...
+
+    Set `connected="true"` to make it visible in a geometry display
+  </comment>
+  <parallelworld_volume name="tracking_volume" anchor="/world" material="Air" connected="false" vis="VisibleBlue">
+    <comment>
+      We first build a tube using the legacy parameters and add a ion beamline extrusion
+    </comment>
+    <shape type="BooleanShape" operation="Union">
+      <shape type="Tube" rmin="0*cm" rmax="tracker_region_rmax" dz="(tracker_region_zmax - tracker_region_zmin) / 2" />
+      <shape type="Tube" rmin="0*cm" rmax="1 * cm" dz="10 * m" />
```
DraTeots: Could we please run it to 36 meters or ZDC. Then I will run the simulations with this settings to test how it works.

DraTeots:
I have this rationale to be challenged for the tracking volume


1. It should be wider than just the beampipe for B0 area as lambdas decaying outside of beampipe has their chance to be detected (very sketchy assumption)
2. It should be the same as beampipe to ZDC after. 

Steps I am doing now: 

1. We have pretty detailed acceptance studies but they all rely on MCParticles (e.g. take lambda daughters and see where they have hits associations) so we need to really save all particles to be reliable (that was not true for the previous data we have). Now we know, that to truly save all particles we switch off Geant4TCUserParticleHandler completely, but all resulting dd4hep outputs bloated from 3gb => 60+gb and the processing is slooow. I will show the results as I have them. 

2. As we have expand the tube to ZDC (or decline this idea), I will run the simulations for this branch and we compare the plots.

veprbl:
Updated parameters based on suggestions from @ajentsch

veprbl:
Here is how it looks now:
image

DraTeots:
Thank you! I will now Simul8

DraTeots:
The good part - we now see the normal lambda decay distribution. I will plot all plots tomorrow, but for now: 

This is 18x275 lambda end point Z- all particles saved (particle handler switched off)

image

This is 18x275  lambda end point Z from this branch:

image

For some reasons there is no ZDC peak there

This is 18x275 ALL PARTICLES decay types: 

image

This is 18x275 this branch decay types

image

So it seems a lot is not captured especially with ZDC

Also interesting. File sizes are smaller than 26.03 image: 

THIS BRANCH:
```
-rw-r--r--. 1 romanov eic 2.5G Apr 21 20:46 k_lambda_10x100_5000evt_0001.edm4hep.root
-rw-r--r--. 1 romanov eic 2.6G Apr 21 20:42 k_lambda_10x100_5000evt_0002.edm4hep.root
-rw-r--r--. 1 romanov eic 2.5G Apr 21 20:26 k_lambda_10x100_5000evt_0003.edm4hep.root
-rw-r--r--. 1 romanov eic 2.6G Apr 21 16:36 k_lambda_10x100_5000evt_0005.edm4hep.root
-rw-r--r--. 1 romanov eic 2.6G Apr 21 19:06 k_lambda_10x100_5000evt_0006.edm4hep.root
-rw-r--r--. 1 romanov eic 2.6G Apr 21 19:11 k_lambda_10x100_5000evt_0007.edm4hep.root
-rw-r--r--. 1 romanov eic 2.6G Apr 21 20:00 k_lambda_10x100_5000evt_0008.edm4hep.root
-rw-r--r--. 1 romanov eic 2.5G Apr 21 20:27 k_lambda_10x100_5000evt_0009.edm4hep.root
-rw-r--r--. 1 romanov eic 2.5G Apr 21 19:54 k_lambda_10x100_5000evt_0010.edm4hep.root
```

26.03 campaign

```
-rw-r--r--. 1 romanov eic 3.6G Apr 11 15:58 k_lambda_10x100_5000evt_0001.edm4hep.root
-rw-r--r--. 1 romanov eic 3.6G Apr 11 15:59 k_lambda_10x100_5000evt_0002.edm4hep.root
-rw-r--r--. 1 romanov eic 3.6G Apr 11 16:00 k_lambda_10x100_5000evt_0003.edm4hep.root
-rw-r--r--. 1 romanov eic 3.6G Apr 11 16:01 k_lambda_10x100_5000evt_0004.edm4hep.root
-rw-r--r--. 1 romanov eic 3.6G Apr 11 16:02 k_lambda_10x100_5000evt_0005.edm4hep.root
-rw-r--r--. 1 romanov eic 3.6G Apr 11 16:03 k_lambda_10x100_5000evt_0006.edm4hep.root
-rw-r--r--. 1 romanov eic 3.6G Apr 11 16:04 k_lambda_10x100_5000evt_0007.edm4hep.root
-rw-r--r--. 1 romanov eic 3.7G Apr 11 16:06 k_lambda_10x100_5000evt_0008.edm4hep.root
-rw-r--r--. 1 romanov eic 3.6G Apr 11 16:07 k_lambda_10x100_5000evt_0009.edm4hep.root
-rw-r--r--. 1 romanov eic 3.6G Apr 11 16:08 k_lambda_10x100_5000evt_0010.edm4hep.root
```

DraTeots:
And we see this on z decay plot above. The plot captures decay Z only of lambda with daughters. You can see like initially the histogram is identical with the histogram where everything is saved. But then differences increase with Z. So on the plot we see how more and more lambdas start decaying outside of the tracking volume, not being saved and vanish from the histogram.  Finally there is no peak corresponding to ZDC as ZDC is completely outside of the tracking volume. 

I believe the best solution would be to add yet another tracking volume arm, so there are 2 arms - one going straight to zdc and the second one going along ion beam. 

The rules for saving particles are: 
1. Particle decays in tracking volume or one of daughters return to tracking volume
2. Particle hits **tracking** sensitive volume (even outside of TV). 

So if we leave just one arm going straight to ZDC we should be fine in general as charged particles going into ion pipe will be saved as there are trackers. But with only one hand we may miss rare something (doesn't come to mind instantly what). But if we still want to be careful two arms are optimal physics wise. 

I will now run simulations of the TW without rotation going directly to ZDC and post here how it compares.

DraTeots:
image
Reverting my change. Lets say "it was evening". But the question about why "Finally there is no peak corresponding to ZDC as ZDC is completely outside of the tracking volume." needs more analysis now.

veprbl:
If this is working, we should merge ASAP to allow for  some grace period before rolling out the npsim change that depends on this.

DraTeots:
Not fully fixed yet. I am working on understanding where is our ZDC peak. Will give the update later today.

DraTeots:
Lets merge this to main (as appropriate). It is much better than was before and we'll create a separate issue as we have a better understanding what else needs to be changed

veprbl:
I agree. Let's merge to get some version of epic with TV to the users ASAP. Not having TV is a fatal error in Geant4TVUserParticleHandler, so once we enable it by default users will have to take actions. Hence I'd like to separate those changes in time as much as possible.

Do you have any estimate on file size impact?

DraTeots:
> Do you have any estimate on file size impact?

No significant impact. It is only MCParticles change but the hits and associations stay the same (maybe pointing to another particles but the number is the same).

DraTeots (dismissed):

compact/tracking_region.xml
veprbl: Done. There is no centralized definition for ZDC front face (even different values of `ZDC_r_pos` defined in different variants), so it can't be a constant just yet.

PR #1089, opened 2026-04-27
ForwardTOF: add 1mm air gaps to decouple layer approach surfaces from volume boundaries

wdconinc:

Add thin (1 mm) Air gap layers before and after the physical ForwardTOF detector layers to ensure ACTS layer approach surfaces remain interior to the tracking volume boundary.

ACTS generates approach surfaces at the inner and outer faces of each layer envelope. When the outer approach surface of the last physical layer coincides exactly with the outer boundary of the ACTS tracking volume, the navigator can encounter floating-point precision issues: a propagation step landing exactly on the boundary may be classified as outside the volume, causing a navigation break. For the ForwardTOF, the Back layer outer face sits at z = 1880.975 mm, which is simultaneously the outer approach surface of the Back ACTS layer and the outer boundary of the ForwardTOF tracking volume.

No physical detector components are moved.

- [x] Medium
- [ ] Low

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Updated documentation
- [x] other: geometry robustness fix (decouple ACTS approach surfaces from tracking volume boundaries)

AI (GitHub Copilot) was used to analyze the anomalous track-projection times observed between ACTS v44.4.0 and v45.1.1, trace the root cause to the ForwardTOF volume-boundary coincidence, and prepare this fix. The C++ and XML changes were reviewed and validated by the author.

---

**Changes:**

- `compact/tracking/definitions_craterlake.xml`: add `ForwardTOF_gap = 1*mm` constant
- `compact/tracking/tof_endcap.xml`:
  - Add `FrontGap` layer (id=5) at z = 1859–1860 mm (before the physical Front layer), using `zstart="ForwardTOF_zmin - ForwardTOF_gap"`
  - Change `Front` layer from `zstart` to `zstack=true` (physical position unchanged at z = 1860 mm)
  - Add `BackGap` layer (id=6) at z = 1880.975–1881.975 mm (after the physical Back layer), using `zstack=true`
  - Gap layers have only a `<envelope thickness="ForwardTOF_gap" .../>` element and no layout or sensitive components
- `src/EndcapTOF_geo.cpp`: support explicit `thickness` attribute on `<envelope>` to allow layout-free passive layers (falls back to computing thickness from module layouts when not provided)

Gap layers contain only Air, produce no hits, do not affect the cellID encoding, and are invisible in visualisation.


ssedd1123 (approved):
Looks good to me.

PR #1103, opened 2026-05-11
DRICH_geo: construct mirror from sphere without phi cuts

veprbl:

The current implementation constructs mirror from a sphere section that spans some hardcoded 80 degrees in phi. In current GPU-accelerated opticks and simphony codes the phi cut is not implemented and produces an error. Since it does not (yet?) matter for DRICH implementation whether spheres can be cut in phi, let's not do it for now.

To better understand the meaning of phi in question see the following renderings:

image

**Legend:** Left to right:
1. nominal geometry
2. geometry with 40 degrees values (80 degrees span) replaced by 10 degrees values (20 degrees span)
3. present PR

- [x] Medium
- [ ] Low

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Updated documentation
- [ ] other: __

### DIFF
--- a/src/DRICH_geo.cpp
+++ b/src/DRICH_geo.cpp
@@ -512,8 +512,8 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
       mirrorTheta2 = M_PI;
     }
 
-    // solid : create sphere at origin, with specified angular limits;
-    // phi limits are increased to fill gaps (overlaps are cut away later)
+    // solid: create sphere at origin with the specified radial and theta limits;
+    // phi trimming is applied later via intersection with the pie-slice wedge
     Sphere mirrorSolid1(mirrorRadius, mirrorRadius + mirrorThickness, mirrorTheta1, mirrorTheta2);
 
     // mirror placement transformation (note: transformations are in reverse order)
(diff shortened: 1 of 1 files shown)


chchatte92:
Hi @veprbl I have difficulty to understand the changes. Is there a image of the mirror-sensor pairs?

c-dilks:
IIRC, this should be fine as long as the pie-slice $z$-length $\lt$ sphere diameter.

veprbl:
@ggalgoczi kindly agreed to test this on our side, to see if it works. Let's wait on his results.

> Hi @veprbl I have difficulty to understand the changes. Is there a image of the mirror-sensor pairs?

This only affects the way in which the mirror volume is constructed. You'd like to see a single segment?

ggalgoczi:
> @ggalgoczi kindly agreed to test this on our side, to see if it works. Let's wait on his results.
> 
> > Hi @veprbl I have difficulty to understand the changes. Is there a image of the mirror-sensor pairs?
> 
> This only affects the way in which the mirror volume is constructed. You'd like to see a single segment?

Working on it, will need a few more days. I am facing some other issues with validating simphony for dRICH and need to decouple them before coming to a conclusion, will follow up.

ggalgoczi:
@veprbl yes, this change to the geometry would be very welcome. There is an other geometrical feature still in that we do not support yet: Cylinder/tube phi cut . However if that is important for physics I can implement it, no problem.

ggalgoczi (changes requested):
This is not needed, because I implemented support for mentioned geometries in simphony in the following PR: 

https://github.com/BNLNPPS/simphony/pull/345

PR #1109, opened 2026-05-25
0x(1)1 issue in pfRICH/dRICH; tracking_and_drich_and_pfrich.yml

veprbl:
The current version of IRT2 indexes detector entries using G4 volume pointer. Using same pointer results in entry overwrite when both pfRICH and dRICH are enabled. This change allows both detectors to work with the current version of IRT.

Submitting on behalf of @alexander-kiselev.


veprbl:
I understand this workaround and approve the change. Please approve the PR, @alexander-kiselev.

wdconinc:
This PR needs a better description.

wdconinc (approved):
Fine with me.

Indexing by pointer in an ordered container (e.g. map) and then iterating over that container is a great way to introduce irreproducible behavior, so maybe there's some follow-up evaluation of this needed in IRT as well.

PR #1110, opened 2026-05-25
DRICH: reuse single PDU assembly template per sector

wdconinc:
## Problem

The dRICH geometry was creating a new `TGeoVolumeAssembly` for every PDU position inside the `(thetaGen, phiGen)` placement loop. Since each of the 210 PDU positions per sector has identical internal geometry, this produced:

- 210 unique TGeoVolumes per sector × 6 sectors = **1260 unique `DRICH_pdu_sec*` volumes**
- Similarly, `sensorAssembly`, `pssVol`, and `resinVol` were recreated per PDU

This inflated TGeo's volume count, increased memory footprint, and slowed geometry lookups.

## Fix

Hoist PDU assembly construction outside the placement loop:
- One `Assembly` template per sector is built **once**, then placed N times with different rotations/translations
- `pssVol`, `resinVol`, and the `sensorAssembly` grid are also created once per sector

TGeo then shares the same volume objects for all PDU copies, reducing unique volume count, memory use, and lookup cost.

**PhysVolID allocation:**
- `sipm` on `pssVol` placement inside `sensorAssembly` (varies per SiPM slot within a PDU)
- `sector` + `pdu` on the `pduAssembly` placement in `gasvolVol` (varies per PDU position)

The DD4hep VolumeManager accumulates physVolIDs from all levels in the navigation path, so the full `{sector, pdu, sipm}` triple is correctly reconstructed at hit time — no change to cellID encoding.

## Results

| Metric | Before | After |
|--------|--------|-------|
| Unique `DRICH_pdu_sec*` TGeoVolumes | 1260 | **6** |

### DIFF
--- a/src/DRICH_geo.cpp
+++ b/src/DRICH_geo.cpp
@@ -707,6 +707,163 @@ static Ref_t createDetector(Detector& desc, xml::Handle_t handle, SensitiveDetec
     // calculate PDU pitch: the distance between two adjacent PDUs
     double pduPitch = pduNumSensors * resinSide + (pduNumSensors + 1) * pduSensorGap + pduGap;
 
+    // BUILD PDU ASSEMBLY TEMPLATE -------------------------------------------------------
+    /* All PDUs within a sector have identical internal geometry; only their placement in
+     * the gas volume differs (each PDU points to a different spot on the sensor sphere).
+     * By building one Assembly template per sector and placing it N times, TGeo can share
+     * the same B-rep description for all PDU copies, dramatically reducing volume count
+     * and STEP export size.
+     *
+     * PhysVolID allocation:
+     *  - "sipm"   is set on the pssVol placement inside sensorAssembly (varies per slot)
+     *  - "sector" and "pdu" are set on the pduAssembly placement in gasvolVol (varies per PDU)
+     * The VolumeManager accumulates IDs from all levels when encoding a hit's cellID, so the
+     * full {sector, pdu, sipm} triple is correctly reconstructed at simulation time.
+     */
+
+    /* begin building sensors and PDUs, where:
+     * - sensor assembly: collection of all objects for a single SiPM
+     * - photodetector unit (PDU) assembly: matrix of SiPMs with services
+     *   - coordinate system: the "origin" of the assembly will be the center of the
+     *     outermost surface of the photosensitive surface (pss)
+     *     - reconstruction can access the sensor surface position from the sensor
+     *       assembly origin, which will ultimately have coordinates w.r.t. to the IP after
+     *       placement in the dRICH vessel
+     *     - the pss is segmented into SiPM pixels; gaps between the pixels
+     *       are accounted for in reconstruction, and each pixel reads out as a unique `cellID`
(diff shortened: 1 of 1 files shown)


wdconinc:
No diffs in e.g. https://eic.github.io/epic/pr/1110/capybara/sim_dis_18x275_minQ2=1000_epic_craterlake/index.html#DRICHHits, but I wonder if it is possible that lsb CellID differences are not captured there...

wdconinc:
Before (this thing has a height):
epic_drich_only

After:
epic_drich_only_new

wdconinc:
Slight geometry initialization speedup, over 5 runs:
```
  ┌───────────────────────────────────┬─────────────────────────────┐
  │                                   │ Wall time (mean ± stdev)    │
  ├───────────────────────────────────┼─────────────────────────────┤
  │ OLD (1260 unique PDU volumes)     │ 3.477 ± 0.189 s             │
  ├───────────────────────────────────┼─────────────────────────────┤
  │ NEW (6 unique PDU volumes)        │ 3.244 ± 0.146 s             │
  ├───────────────────────────────────┼─────────────────────────────┤
  │ Delta                             │ −0.233 s (−6.7%), t = −2.19 │
  └───────────────────────────────────┴─────────────────────────────┘
```
Memory use change is negligible (as expected). Geometry tree balancing is the main benefit here (and why the duplication into 6 sectors is not a big concern).

chchatte92:
> No diffs in e.g. https://eic.github.io/epic/pr/1110/capybara/sim_dis_18x275_minQ2=1000_epic_craterlake/index.html#DRICHHits, but I wonder if it is possible that lsb CellID differences are not captured there...

Hi @wdconinc I will run the full chain (locally) with single particles and see if the Cherenkov angles are meaningful. I guess there is no reason to believe that they would change, given the cellID does not change. Otherwise, I think this is fine.

chchatte92:
> Before (this thing has a height): epic_drich_only
> 
> After: epic_drich_only_new

Sorry for my ignorance, but this plot is not super clear to me. What does it change BTW?

wdconinc:
> > Before (this thing has a height): epic_drich_only
> > After: epic_drich_only_new
> 
> Sorry for my ignorance, but this plot is not super clear to me. What does it change BTW?

The geometry before has a lot of identical volumes that are recreated over and over and each placed. Now the geometry tree just places the same volume multiple times. This leads to a geometry tree that is easier to navigate.

veprbl (approved):
To me this looks fine. Capybara report claims there are no difference in output. I looked into pssDE and pssSkin changes. For pssDE, I don't think placedVolume is used by IRT1, it's probably all for those constants that are passed around. For SkinSurface, the second argument defines more like namespace, so that should be fine as well.

chchatte92:
@kumardeepraman tested the differences in Cherenkov angle and Npe, no differences seen.

chchatte92 (approved):

wdconinc:
> @kumardeepraman tested the differences in Cherenkov angle and Npe, no differences seen.

Well, I guess there are differences but they are statistically insignificant and caused by the known irreproducibility in IRT.

PR #1112, opened 2026-05-26
LFHCAL: reuse slice volume templates across readout sections

wdconinc:
## Summary

The LFHCAL XML defines 7 readout sections (`readoutlayer=0..6`), all with identical slice geometry (Polystyrene + Kapton + Steel235 with the same thicknesses). Previously, `layer_num` was reset to 0 at each readout section boundary inside `createDetector`, causing `createEightMModule`/`createFourMModule` to create separate `TGeoVolume` objects for every layer index in every section — producing 1704 unique volumes when ~57 suffice.

## Fix

Pre-build one template per slice type (absorber, filler, scintillator assembly) from the first occurrence of each `slice_partID`. For all subsequent layers, place the same template. The per-layer `layerz` and `rlayerz` physVolIDs are set on each **placement node** (consistent with how DD4hep's VolumeManager collects IDs along the full path).

The `createScintillatorPlateEightM`/`createScintillatorPlateFourM` helper signatures no longer receive `layerID`/`roLayer`; foil volume names are now fixed strings instead of layer-indexed.

## Impact

Measured locally using `npdet_to_dot` and `checkGeometry`:

| Metric | Before | After |
|--------|--------|-------|
| Unique TGeoVolumes (DOT nodes) | 1704 | 57 |
| Reduction | — | **97%** |
| Peak PSS (`checkGeometry` on `epic_lfhcal_only.xml`) | ~596 MB | ~585 MB |
| Geometry overlaps | 0 | 0 |

The primary benefit is reduced unique volume count — fewer `G4LogicalVolume` objects during Geant4 initialization, a smaller `VolumeManager` tree in EICRecon, and lower memory footprint at startup.

### DIFF
--- a/src/LFHCAL_geo.cpp
+++ b/src/LFHCAL_geo.cpp
@@ -210,13 +210,11 @@ Volume createScintillatorTower(Detector& desc, std::string basename, double w_to
 //************************************************************************************************************
 //************************** create scintillator plate with separations for 8M *******************************
 //************************************************************************************************************
-Assembly createScintillatorPlateEightM(Detector& desc, std::string basename,
-                                       //                                         int modID,
-                                       int layerID, double h_mod, double w_mod, double t_mod_tp,
-                                       double t_mod_sp, double t_slice, double w_notch,
-                                       double t_foil, Material slice_mat, int roLayer,
-                                       std::string region, std::string limit, std::string vis,
-                                       SensitiveDetector sens, bool renderComp) {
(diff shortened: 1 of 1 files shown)


wdconinc:
No diffs in https://eic.github.io/epic/pr/1112/capybara/sim_dis_18x275_minQ2=1000_epic_craterlake/index.html (or any of the rec outputs).

PR #1114, opened 2026-05-29
feat(ci): add 'geantino' to particle matrix

wdconinc:
This pull request expands the CI simulation particle matrix and fixes artifact naming compatibility.

- CI configuration:
  * Updated `particle` matrix in `.github/workflows/linux-eic-shell.yml` from `[pi-, e-]` to `[e, pi, geantino, chargedgeantino]`, with an explicit `charge` variable (`"-"` for `e`/`pi`, `""` for geantino variants) so that `--gun.particle` receives the correct signed name (e.g. `e-`) while artifact names remain backward-compatible (using the bare particle name without the charge sign).

- [x] Medium
- [ ] Low

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [ ] Updated documentation
- [ ] other: __

GitHub Copilot was used to implement the `charge` matrix variable approach and update artifact name references.

### DIFF
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -557,7 +557,7 @@ jobs:
     needs: build
     strategy:
       matrix:
-        particle: [pi, e]
+        particle: [pi, e, geantino]
         detector_config: [epic_craterlake]
     steps:
     - uses: actions/checkout@v6
--- a/.github/workflows/linux-eic-shell.yml
+++ b/.github/workflows/linux-eic-shell.yml
@@ -566,9 +566,12 @@ jobs:
             charge: "-"
           - particle: geantino
             charge: ""
+            sim: ddsim
             extra_args: '--filter.tracker "" --filter.calo ""'
           - particle: chargedgeantino
             charge: ""
+            sim: ddsim
+            extra_args: '--filter.tracker "" --filter.calo ""'
     steps:
     - uses: actions/checkout@v6
     - uses: actions/download-artifact@v8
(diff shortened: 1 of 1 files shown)


veprbl:
Geantino doesn't produce hits? I think, I used to get those back in the day. But now can't recall what settings one needs to tweak.

wdconinc:
> Geantino doesn't produce hits? I think, I used to get those back in the day. But now can't recall what settings one needs to tweak.

I think our SD filters are restricting hits to charged particles only.

wdconinc:
This still misses the hits on the trackers, but it's a more fundamental issue in our npsim extension of ddsim. In npsim we start with a parseOptions, then later set `RUNNER.filter.tracker = 'edep0'` (which requires edep > 0, not edep >= 0 that geantinos would satisfy). We can't place parseOptions last, because then other stuff breaks (see npsim.py comment)...

~~I'd say we call this ok for now since at least calo hits is better than nothing...~~

*Edit:* I fixed this by using ddsim for the geantinos, which is sufficient.

.github/workflows/linux-eic-shell.yml
```diff
@@ -549,8 +549,11 @@ jobs:
     needs: build
     strategy:
       matrix:
-        particle: [pi, e]
+        particle: [pi-, e-, geantino]
```
wdconinc:
```suggestion
        particle: [pi-, e-, geantino, chargedgeantino]
```

wdconinc:
Now produces hits in both trackers and calos by using ddsim (the optical photon additions in npsim are irrelevant for geantinos). Added charged geantino for magnetic fields.

veprbl (approved):
LGTM. Not breaking currently used filenames with charge signs is good.

PR #1124, opened 2026-06-04
Bump material map scripts to 45.3.0

veprbl:
https://github.com/eic/containers/pull/293

### DIFF
--- a/scripts/material_map/epic.py
+++ b/scripts/material_map/epic.py
@@ -17,7 +17,7 @@ def getDetector(
     logLevel=acts.logging.WARNING,
 ):
     customLogLevel = acts.examples.defaultLogging(logLevel=logLevel)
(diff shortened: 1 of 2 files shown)


ShujieL (approved):

wdconinc:
When can we get rid of this? The reason we put this in our stack is because we were n>3 major versions behind and we had no timely way to get changes from upstream into our stack. This is better now, so we should try to remove some of this technical debt again (and fully upstream whatever changes we are making to the upstream scripts).

veprbl:
Can you clarify which part do we get rid of? I think the answer is yes, we can.

wdconinc:
> Can you clarify which part do we get rid of? I think the answer is yes, we can.

I thought originally we had to download the scripts because they aren't installed (but I think this is not the case anymore) and we also probably don't have to make patches anymore.

PR #1135, opened 2026-07-06
cmake: fix git_version

veprbl:
We currently always get v0.0.0-unknown.

Two issues here

 * CMAKE_PROJECT_VERSION is not settable externally
 * project() only accepts specific format of the version (no -dirty and such)

- [ ] Medium
- [x] Low

- [ ] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Updated documentation
- [ ] other: __

  Claude


veprbl:
@wdconinc Could you check this out? Next, I will update the eic-spack package and extend this to other cases we have.

cmake/git_version.cmake
```diff
@@ -13,17 +19,36 @@ macro(set_git_version VERSION)
...
+        "Failed to determine VERSION from Git tags. Using default version \"${${VERSION}_FULL}\"."
     )
   endif()
+
+  # Strip a leading "v" (e.g. v1.2.3 -> 1.2.3)
+  string(REGEX REPLACE "^v" "" _git_version_stripped "${${VERSION}_FULL}")
+
+  # Extract only the leading X.Y.Z or X.Y.Z.N numeric part so the result is a
+  # valid CMake version string that can be passed to project().
+  string(REGEX MATCH "^([0-9]+\\.[0-9]+(\\.[0-9]+(\\.[0-9]+)?)?)" _git_version_clean "${_git_version_stripped}")
```
wdconinc, line 41:
```suggestion
  string(REGEX MATCH "^([0-9]+\\.[0-9]+(\\.[0-9]+([\\.-][0-9]+)?)?)" _git_version_clean "${_git_version_stripped}")
```
I think this would also map `26.06.0-10-ge600dc77d` to `26.06.0.10`, which is sequential and unambiguous.

wdconinc (approved):
LGTM.

cmake/git_version.cmake
veprbl: This would not work because it maps to `26.06.0-10`. Let's leave this for later, when we need this.

wdconinc: Yeah, I realized that too just now :-)

PR #1152, opened 2026-08-12
Insert geometry update

aidenwu23:

This introduces an updated insert geometry based on recent mechanical redesigns, notably:
     1. Physical square scintillator tiles
     2. Fixed beampipe cutout
     3. Longitudinal PCBs and slots
     4. Steel casings and PCB covers

The updated geometry and its role in the next steps for the Insert calorimeter were presented and discussed at the August 12, 2026 LFHCal general meeting: https://indico.bnl.gov/event/33832/.

It would be useful to include this geometry as a separate configuration in the September campaign to provide more realistic data-rate estimates and validation.

Separately, this PR updates the LFHCal readout-layer grouping to match the latest test-beam campaign configuration.

- [x] Medium
- [ ] Low

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [x] Updated constants (issue #__)
- [ ] Updated documentation
- [ ] other: __

Updates the default LFHCal readout-layer grouping to match the latest test-beam campaign configuration.


aidenwu23:
@wdconinc @veprbl John suggested I tag you both for review and to get your thoughts on whether the updated Insert can be supported as a separate production configuration. If that would complicate the production workflow, we could instead make the updated geometry the new baseline.

wdconinc:
Can we just update the existing geometry plugin instead of adding a *_new file? It makes it impossible to see what actually changed and what's new. Review will take much longer.

aidenwu23:
Yeah, that makes sense. I’ll update the existing geometry plugin directly instead of keeping a separate *_new file.

aidenwu23:
updated

veprbl:
@aidenwu23 Could you join `@eic` organization by following https://eic.github.io/documentation/getstarted
You need to push your branch to eic/epic and resubmit a PR (we don't accept contributions from forks due to current CI restrictions).

aidenwu23:
A new PR has been made: https://github.com/eic/epic/pull/1153

PR #1153, opened 2026-08-13
Insert/LFHCAL geometry update

aidenwu23:

This introduces an updated insert geometry based on recent mechanical redesigns, notably:
     1. Physical square scintillator tiles
     2. Fixed beampipe cutout
     3. Longitudinal PCBs and slots
     4. Steel casings and PCB covers

The updated geometry and its role in the next steps for the Insert calorimeter were presented and discussed at the August 12, 2026 LFHCal general meeting: https://indico.bnl.gov/event/33832/.

It would be useful to include this geometry in the September campaign to provide more realistic data-rate estimates and validation.

Separately, this PR updates the LFHCal readout-layer grouping to match the latest test-beam campaign configuration.

- [x] Medium
- [ ] Low

- [ ] Bug fix (issue #__)
- [x] New feature (issue #__)
- [x] Updated constants (issue #__)
- [ ] Updated documentation
- [ ] other: __

Updates the default LFHCal readout-layer grouping to match the latest test-beam campaign configuration. Also updates the default insert geometry.


configurations/forward_insert_only.yml
wdconinc, line 1: For consistency, rename this file to `forward_insert_only.yml`. People are used to `*_only.yml` meaning that only `compact/*/$1.xml` is included.

veprbl:
Regarding the alternative geometry configuration with previous version, what samples do you expect to be needed in the production?
cc @johnlajoie

compact/hcal/forward_insert.xml
```diff
@@ -1,11 +1,12 @@
 <!-- SPDX-License-Identifier: LGPL-3.0-or-later -->
 <!-- Copyright (C) 2022 Ryan Milton -->
+<!-- Copyright (C) 2026 Aiden Wu -->
```
wdconinc:
```suggestion
<!-- Copyright (C) 2022 - 2026 Ryan Milton, Aiden Wu -->
```

aidenwu23:
At least a 10x275 min-bias Geant4 sample with background for hit-rate/data-rate studies.

src/InsertCalorimeter_geo.cpp
```diff
@@ -1,216 +1,487 @@
 // SPDX-License-Identifier: LGPL-3.0-or-later
 // Copyright (C) 2022 Ryan Milton
-
-//==========================================================================
-//  Implementation of forward insert calorimeter
-//--------------------------------------------------------------------------
-//  Author: Ryan Milton (UCR)
-//==========================================================================
+// Copyright (C) 2026 Aiden Wu
+// Updated geometry dimensions provided by Eliott Fountain.
```
wdconinc: Collapse copyright lines.

src/InsertCalorimeter_geo.cpp
```diff
@@ -1,216 +1,487 @@
...
+  const xml::Component& slots_xml        = detElem.child(_Unicode(pcb_slots));
+  const xml::Component& pcb_xml          = detElem.child(_Unicode(pcb_boards));
+  const xml::Component& tiles_xml        = detElem.child(_Unicode(tiles));
+  const xml::Component& casing_xml       = detElem.child(_Unicode(casing));
+  const xml::Component& cover_xml        = detElem.child(_Unicode(pcb_covers));
+  const xml::Component& back_cutouts_xml = detElem.child(_Unicode(backplate_cutouts));
+
+  // Read beam-opening geometry.
+  const double center_x =
+      dd4hep::getAttrOrDefault<double>(beam_xml, _Unicode(center_x), -10.0 * cm) - pos.x();
```
wdconinc: I'd remove the default values for these values, since they will get out of sync and you can make it a hard error if they are absent.

src/InsertCalorimeter_geo.cpp
```diff
@@ -1,216 +1,487 @@
...
+      Box v_pcb_slot(v_pcb_slot_width / 2., left_v_pcb_slot_height / 2., thickness / 2.);
+      cutout = UnionSolid(cutout, v_pcb_slot, Position(left_v_pcb_slot_x, left_v_pcb_slot_y, 0.));
+      cutout = UnionSolid(cutout, v_pcb_slot, Position(left_v_pcb_slot_x, -left_v_pcb_slot_y, 0.));
+    }
+
+    return cutout;
+  };
+
+  // Insert assembly.
   Assembly assembly(detName);
```
wdconinc, line 183: Part of the reason for the assembly here was the changing hole cutout radius in the past. Can we now just use a regular envelope volume?

wdconinc (commented):
Looks fine to me. Some relatively minor comments.

src/InsertCalorimeter_geo.cpp
aidenwu23: A regular envelope should be feasible, but doing so would likely require a complex mother solid that reproduces the beam opening and contains the protruding tiles and PCB covers, so I retained the simpler assembly.

wdconinc: Fair. I thought it was now closer to a simple box with tube hole. I didn't realize the tiles and covers were protruding.

veprbl:
> At least a 10x275 min-bias Geant4 sample with background for hit-rate/data-rate studies.

Okay, let's review and merge the changes to geometry first. We can add legacy configurations next PR.

veprbl:
Congratulations on finalizing this geometry update, I think it looks reasonable and is well documented. Very well done!

There is a concern that I have is that you are disabling hexagonal segmentation in favor of square unstaggered tiles, which will break the assumptions for HexPlit reconstruction configuration in the EICrecon. We've recently did a similar change to ZDC which caused a regression for pi0 reconstruction https://github.com/eic/EICrecon/issues/2737.

To avoid unnecessary regressions, the following needs to happen:
1. In EICrecon reconstruction steering we need to add a code that detects the new geometry (e.g. by querying the segmentation type for the `HcalEndcapPInsertHits` readout. Should it detect the new status, it will not anymore construct the `HcalEndcapPInsertSubcellHits` or any other HEXPLIT related path. Instead it will use IslandClustering, likely with a simple configuration like `.dimScaledLocalDistXY = {1.5, 1.5}`.
2. We need use your new geometry and updated EICrecon to run `insert_` benchmarks from https://github.com/eic/detector_benchmarks/tree/master/benchmarks and ensure that they continue to still produce reasonable performances.

aidenwu23:
> Congratulations on finalizing this geometry update, I think it looks reasonable and is well documented. Very well done!
> 
> There is a concern that I have is that you are disabling hexagonal segmentation in favor of square unstaggered tiles, which will break the assumptions for HexPlit reconstruction configuration in the EICrecon. We've recently did a similar change to ZDC which caused a regression for pi0 reconstruction [eic/EICrecon#2737](https://github.com/eic/EICrecon/issues/2737).
> 
> To avoid unnecessary regressions, the following needs to happen:
> 
> 1. In EICrecon reconstruction steering we need to add a code that detects the new geometry (e.g. by querying the segmentation type for the `HcalEndcapPInsertHits` readout. Should it detect the new status, it will not anymore construct the `HcalEndcapPInsertSubcellHits` or any other HEXPLIT related path. Instead it will use IslandClustering, likely with a simple configuration like `.dimScaledLocalDistXY = {1.5, 1.5}`.
> 2. We need use your new geometry and updated EICrecon to run `insert_` benchmarks from https://github.com/eic/detector_benchmarks/tree/master/benchmarks and ensure that they continue to still produce reasonable performances.

I had been suspecting this might come up, so I looked into it preemptively. I’m much less familiar with EICrecon than I am with the ePIC geometry, but I believe I’ve made reasonably correct modifications and will open a PR there.

compact/hcal/lfhcal.xml
```diff
@@ -181,6 +181,11 @@
         <slice material="Kapton" thickness="LFHCALKaptonThickness" vis="LFHCALLayerKaptonVis" type="2" offset="LFHCALAirOffset"/>
         <slice material="Steel235" thickness="LFHCALAbsorberThickness" vis="LFHCALLayerSteelVis" type="1"/>
       </layer>
+      <layer repeat="LFHCALLayer_NSteelRepeatS" vis="InvisibleWithDaughters" readoutlayer="7">
+        <slice material="Polystyrene" thickness="LFHCALPolystyreneThickness" vis="LFHCALLayerScintVis" sensitive="yes" limits="cal_limits" type="3" />
+        <slice material="Kapton" thickness="LFHCALKaptonThickness" vis="LFHCALLayerKaptonVis" type="2" offset="LFHCALAirOffset"/>
+        <slice material="Steel235" thickness="LFHCALAbsorberThickness" vis="LFHCALLayerSteelVis" type="1"/>
+      </layer>
```
veprbl, line 188: @FriederikeBock, do you approve this LFHCAL change or should we revert and submit separately?

veprbl (approved):
LGTM

compact/hcal/lfhcal.xml
FriederikeBock: yes this is fine

FriederikeBock (approved):

PR #1159, opened 2026-08-23
fix: update material map validation script to Acts v46.8.1

wdconinc:
### Dependencies
- [x] https://github.com/eic/containers/pull/397

This PR updates the material map validation script to Acts v46.8.1:
- Update ACTS_VERSION from v45.3.0 to v46.8.1 to match the Acts version in the platform
- Remove PR #5359 patch application, as the fix landed in Acts v46.3.0 (function kept)
- This fixes CI failures in the validate-material-map job after the Acts upgrade

- [x] High (please describe reason below)
- [ ] Medium
- [ ] Low

CI is currently broken on `main`.

- [x] Bug fix (issue: https://github.com/eic/epic/actions/runs/32654353552/job/97232215042)
- [ ] New feature (issue #__)
- [ ] Updated documentation
- [ ] other: __


veprbl (dismissed):

veprbl:
Traceback (most recent call last):
      File "/home/runner/work/epic/epic/scripts/material_map/material_recording_epic.py", line 69, in <module>
        main()
        ~~~~^^
      File "/home/runner/work/epic/epic/scripts/material_map/material_recording_epic.py", line 59, in main
        runMaterialRecording(
        ~~~~~~~~~~~~~~~~~~~~^
            detector=detector,
            ^^^^^^^^^^^^^^^^^^
        ...<3 lines>...
            s=acts.examples.Sequencer(events=args.events, numThreads=1),
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        ).run()
        ^
    TypeError: runMaterialRecording() got an unexpected keyword argument 'outputDir'

wdconinc:
> ```
> TypeError: runMaterialRecording() got an unexpected keyword argument 'outputDir'
> ```

It's a different error at least  now :-)

wdconinc:
> It's a different error at least now :-)

It's all in the definition of "progress".

wdconinc:
FYI @ybedfer This is another change you will need, likely.

scripts/material_map/materialPlotHelper.cpp.patch
wdconinc, line 1: Replace with https://github.com/acts-project/acts/pull/6010 when merged.

scripts/material_map/epic.py
wdconinc:
```suggestion
        )
    else:
        logger.warning("Material map %s not loaded.", file.absolute())
```

wdconinc:
So, this runs now to completion, but doesn't produce the right material maps. I propose to merge as is, which unblocks some CI, and continue looking into the reasons why this fails. @veprbl @ShujieL 

Current:
Ratio_Val_geant_mat_X0_Eta

Regenerated:
Ratio_Val_geant_mat_X0_Eta

wdconinc:
This now seems to work better:

Current:
Ratio_Val_geant_mat_X0_Eta


Regenerated:
Ratio_Val_geant_mat_X0_Eta

scripts/material_map/geometry_epic.py
```diff
@@ -26,17 +26,39 @@
...
-        outputDir=Path.cwd(),
-        outputObj=False,
-        outputCsv=False,
-        outputJson=True,
-    )
+    import tempfile
+    import shutil
+
+    # Generate geometry to a temporary directory, then rename the output file
+    tmpdir = Path(tempfile.mkdtemp())
```
veprbl, line 46: Where is from? Why do we need a temp dir?

wdconinc: See https://github.com/eic/epic/commit/f793e8f2aed66d0a7cfdb7dbfcee7c7ec7d34f04. Debugging of all this was helped by running several attempts in parallel. That only works if they don't clobber each other; or start at the same time. This makes geometry more atomic.

wdconinc: And in particular, there is no way to specify a different name than `geometry-map.json`, so we have to capture it in a separate (i.e. temp) dir and then move it to the prefix/suffix name.

wdconinc: Hardcoded filename is at https://github.com/acts-project/acts/blob/main/Examples/Scripts/Python/geometry.py#L103. We could patch that too, I guess.

veprbl: Okay fine. Also strange how LLM didn't know about tempdir.TemporaryDirectory, but okay.

veprbl (approved):

scripts/material_map/geometry_epic.py
wdconinc: First line of TemporaryDirectory docs:
> This class securely creates a temporary directory using the same rules as `mkdtemp()`

PR #1178, opened 2026-09-17
ASIC chips on BTOF should not be sensitive

ssedd1123:
- [x] High (please describe reason below)
- [ ] Medium
- [ ] Low

ASIC chips shouldn't be sensitive. Some of them were set to sensitive="true" by mistake. Need to turn them off for upcoming tracking studies. 

- [x] Bug fix (issue #__)
- [ ] New feature (issue #__)
- [ ] Updated documentation
- [ ] other: __

There is now less sensitive volume, so tracking resolution or efficiency MAY goes down. However, given how simple the fix is and how obvious the bug is, I believe deep analyzes on its effect on the performance can be done after the pull request is accepted.


veprbl (approved):

veprbl:
How did you discover this?

ssedd1123:
Shujie wants me to add noise for tracking analysis, and I was struggling because the added noise are not distributed uniformly in z. After some debugging, it was found that noise is added to ASIC chip, which reveals this bug.

PR #1179, opened 2026-09-18
FieldMapB: graceful crash on NANs in pos

veprbl:
Invalid input files with NANs (recently https://github.com/eic/EIC_ESR_Xsuite/issues/1, but there are other examples) will produce particles/trajectories that have NAN components, which results in obscure crashes during field lookup. This is intended to make such crashes more explicit.


wdconinc (approved):
Not really the field's job which means if we switch to e.g. covfie this will not persist. Might be something for a DD4hep input plugin.

But good.

veprbl:
> Not really the field's job which means if we switch to e.g. covfie this will not persist. Might be something for a DD4hep input plugin.

I agree. In principle, we could just fix the boundary check `(x < min) || (x > max)` becomes `!((x >= min) && (x <= max))` and call it a day. But also why not just panic.

